PR #7, opened 2015-02-27
Maven configuration to enable publishing binaries to Sonatype repository.

anthmgoogle:
Tooling configuration to allow automated deployment of binary releases to https://oss.sonatype.org. Some additional metadata to satisfy requirements.


ejona86:
@anthmgoogle, LGTM

PR #11, opened 2015-03-27
Breaking Change: Add uri parameter to getRequestMetadata.

anthmgoogle:
Applying a breaking change from the specification that is required to enable the JwtAccess credential implementation.

### DIFF
--- a/credentials/pom.xml
+++ b/credentials/pom.xml
@@ -5,7 +5,7 @@
 
   <parent>
     <groupId>com.google.auth</groupId>
-    <artifactId>google-auth-library</artifactId>
+    <artifactId>google-auth-library-parent</artifactId>
     <version>0.1.0</version>
     <relativePath>../pom.xml</relativePath>
   </parent>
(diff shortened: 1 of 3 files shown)


ejona86:
This is going to be a bit of a problem to implement on the gRPC side, as we don't know the scheme and we normally won't know the host with the current implementation. @louiscryan, do you have ideas for handling the new API in gRPC?

anthmgoogle:
For reference @tbetbetbe and Julien Boeuf have implemented equivalent plumbing for Ruby and C++. We need at least the uri to construct the JwtAccess/IdToken header.

credentials/java/com/google/auth/Credentials.java
```diff
@@ -27,7 +27,7 @@
    *
    * @throws IOException if there was an error getting up-to-date access.
    */
-  public abstract Map<String, List<String>> getRequestMetadata() throws IOException;
+  public abstract Map<String, List<String>> getRequestMetadata(String uri) throws IOException;
```
ejona86: Instead of passing in a String, should this be a URI?

anthmgoogle:
Updated to use URI instead of String.

ejona86:
Do you think URI makes more sense? I am okay either way, but wanted to pose the question.

@anthmgoogle, LGTM

anthmgoogle:
Yes, I think URI works better for Java.
I'll wait until we confirm that we will be able to pass this in before merging.

PR #13, opened 2015-04-01
Add URI parameter to getRequestMeatadata.

anthmgoogle:
A revised version of request #11 that is non-breaking for callers, which should make the staging easier.

### DIFF
--- a/credentials/java/com/google/auth/Credentials.java
+++ b/credentials/java/com/google/auth/Credentials.java
@@ -1,6 +1,7 @@
 package com.google.auth;
 
 import java.io.IOException;
+import java.net.URI;
 import java.util.List;
 import java.util.Map;
 
@@ -27,7 +28,21 @@ public abstract class Credentials {
    *
    * @throws IOException if there was an error getting up-to-date access.
    */
-  public abstract Map<String, List<String>> getRequestMetadata() throws IOException;
+  public Map<String, List<String>> getRequestMetadata() throws IOException {
+    return getRequestMetadata(null);
+  }
+
+  /**
+   * Get the current request metadata, refreshing tokens if required.
+   *
+   * <p>This should be called by the transport layer on each request, and the data should be
+   * populated in headers or other context. The operation can block and fail to complete and may do
+   * things such as refreshing access tokens.
+   *
+   * @param uri URI of the entry point for the request.
+   * @throws IOException if there was an error getting up-to-date access.
+   */
+  public abstract Map<String, List<String>> getRequestMetadata(URI uri) throws IOException;
 
   /**
    * Whether the credentials have metadata entries that should be added to each request.
--- a/oauth2_http/java/com/google/auth/http/HttpCredentialsAdapter.java
+++ b/oauth2_http/java/com/google/auth/http/HttpCredentialsAdapter.java
@@ -7,6 +7,7 @@ import com.google.api.client.http.HttpRequestInitializer;
 import com.google.api.client.util.Preconditions;
 
 import java.io.IOException;
+import java.net.URI;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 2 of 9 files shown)


oauth2_http/javatests/com/google/auth/TestUtils.java
```diff
@@ -64,6 +67,14 @@ public static InputStream jsonToInputStream(GenericJson json) throws IOException
     return map;
   }
 
+  public static URI toURI(String uri) {
+    try {
+      return new URI(uri);
```
ejona86: URI.create?

credentials/java/com/google/auth/Credentials.java
```diff
@@ -21,13 +22,25 @@
...
+   *
    * <p>This should be called by the transport layer on each request, and the data should be
    * populated in headers or other context. The operation can block and fail to complete and may do
    * things such as refreshing access tokens.
    *
+   * @param uri URI of the entry point for the request.
    * @throws IOException if there was an error getting up-to-date access.
    */
-  public abstract Map<String, List<String>> getRequestMetadata() throws IOException;
+  public abstract Map<String, List<String>> getRequestMetadata(URI uri) throws IOException;
```
louiscryan: Can the URI be provided at Credential construction time? From talking to Jose the cardinality of 'audience' values is likely to be quite low (just as for scopes) so it may be more practical to have a credential-instance-per-audience than requiring it to be provided per-call by some framework layer below the application code. e.g. the ClientAuthInterceptor in grpc.

@casiboy

oauth2_http/javatests/com/google/auth/TestUtils.java
anthmgoogle: Ah. That method can replace his whole helper. Will change to use that.

credentials/java/com/google/auth/Credentials.java
anthmgoogle: There was a lot of discussion about this w.r.t. the C++ implementation, where that pattern was proposed and it created pretty significant difficulties for that implementation. Another motivation is that we can improve the Http experience with this, by looking up and caching scopes from discovery if they are missing.

However, if this model creates difficulty for Java GRPC, it would be possible to support both the parameter model and constructor model. Credentials like JwtAccess that need this can also provide a way to specify a default value for the URI in the constructor and a cloning method.

If this sounds good, I can just revise the documentation to clarify that the parameter free overload is not temporary, and add this behavior into the JwtAccess credentials implementation.

anthmgoogle:
Revised to use URI.create in the test code.

credentials/java/com/google/auth/Credentials.java
```diff
@@ -21,13 +22,25 @@
   /**
    * Get the current request metadata, refreshing tokens if required.
    *
+   * <p>For temporary compatibility only. Passes a null URI.
```
louiscryan: Consider updating the comment here?

anthmgoogle: Yes, I'll revise the comment here.

anthmgoogle:
There was some discussion about whether URI should be part of getRequestMetadata. The revised PR has both overloads and removed a comment indicating one would be deprecated. This makes it a non-breaking change for the transport and we can evaluate both for use with gRPC.

oauth2_http/javatests/com/google/auth/TestUtils.java
```diff
@@ -9,9 +9,12 @@
 import com.google.common.base.Splitter;
 import com.google.common.collect.Lists;
 
+import java.lang.RuntimeException;
```
ejona86: These imports look strange. You didn't change this file otherwise and you imported something from java.lang.

ejona86:
@anthmgoogle, other than the strange imports, LGTM

oauth2_http/javatests/com/google/auth/TestUtils.java
anthmgoogle: Thanks for catching this error. It was left over from a previous helper. I'll remove.

anthmgoogle:
Updated to remove unintended changes to imports.

ejona86:
@anthmgoogle, LGTM

louiscryan:
LGTM

PR #15, opened 2015-06-19
Fix build warnings.

anthmgoogle:


anthmgoogle:
@ejona86, @louiscryan, available for a quick code review of warning fixes?

oauth2_http/java/com/google/auth/http/HttpCredentialsAdapter.java
```diff
@@ -32,7 +32,6 @@ public HttpCredentialsAdapter(Credentials credentials) {
      *
      * @param request HTTP request
      */
-  @Override
```
ejona86: Why remove this @Override?

ejona86:
@anthmgoogle, except for the removal of @Override, LGTM

oauth2_http/java/com/google/auth/http/HttpCredentialsAdapter.java
anthmgoogle: I get a build error in Eclipse if this is removed. I think it is because it is implementing an interface method, not overriding a virtual method.

anthmgoogle:
Replied to open comment and added a couple of files missing from first PR: ServiceAccountJwtAccessCredentials and ServiceAccountJwtAccessCredentialsTest

oauth2_http/java/com/google/auth/http/HttpCredentialsAdapter.java
ejona86: I thought that issue had been fixed with Java 6... sigh...

ejona86:
@anthmgoogle LGTM

ejona86:
I don't remember, am I supposed to merge the PR, or will you?

anthmgoogle:
Mergd. I was planning to do the merge once I had the LGTM. I think it is probably best for the submitter to do it if they have permission or explicitly request the merge from a writer so they have control over timing or ordering of their changes going in.

ejona86:
SGTM. That's what we do in grpc-java. The grpc repo doesn't though, since they use the merge commit itself as a "record of review" of sorts.

PR #16, opened 2015-06-25
Google App Engine Credentials.

anthmgoogle:
Initial implementation of AppEngineCredentials in new module because of GAE SDK dependencies.

### DIFF
--- a/appengine/java/com/google/auth/appengine/AppEngineCredentials.java
+++ b/appengine/java/com/google/auth/appengine/AppEngineCredentials.java
@@ -1,0 +1,88 @@
+/*
+ * Copyright 2015, Google Inc. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met:
+ *
+ *    * Redistributions of source code must retain the above copyright
+ * notice, this list of conditions and the following disclaimer.
+ *    * Redistributions in binary form must reproduce the above
+ * copyright notice, this list of conditions and the following disclaimer
+ * in the documentation and/or other materials provided with the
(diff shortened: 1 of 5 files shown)


appengine/java/com/google/auth/appengine/AppEngineCredentials.java
```diff
@@ -0,0 +1,87 @@
...
+  private final Collection<String> scopes;
+  
+  private final boolean scopesRequired;  
+  
+  public AppEngineCredentials(Collection<String> scopes) {
+    this(scopes, null);
+  }
+
+  public AppEngineCredentials(Collection<String> scopes, AppIdentityService appIdentityService) {
+    this.scopes = scopes;
```
ejona86: Do we need to make a defensive copy?

ejona86:
@anthmgoogle LGTM

appengine/java/com/google/auth/appengine/AppEngineCredentials.java
anthmgoogle: Done.

anthmgoogle:
Change to make collection immutable copy as suggested.

ejona86:
Are we interested in doing the same in [ServiceAccountCredentials](https://github.com/google/google-auth-library-java/blob/5868c19e27efa4910c09d17f95719d1d7b4f354b/oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java#L87)?

anthmgoogle:
Yes. We did unmodifiable in the previous gen libraries, but we should probably do immutable. Bug tracking this:
https://github.com/google/google-auth-library-java/issues/17

PR #19, opened 2015-07-01
Fix security exceptions running Default Credentials in GAE sandbox.

anthmgoogle:
When this API is called from GAE, the file system checks will throw a security exception. These should be treated as if the files did not exists to enable the other types of credentials to work.


ejona86:
@anthmgoogle LGTM

PR #20, opened 2015-07-01
Support for App Engine in Application Default Credentials.

anthmgoogle:


oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java
```diff
@@ -141,6 +158,76 @@ private final File getWellKnownCredentialsFile() {
...
+    Exception cause = null;
+    Field environmentField;
+    try {
+      environmentField = systemPropertyClass.getField("environment");
+      Object environmentValue = environmentField.get(null);
+      Class<?> environmentType = environmentField.getType();
+      Method valueMethod = environmentType.getMethod("value");
+      Object environmentValueValue = valueMethod.invoke(environmentValue);
+      return (environmentValueValue != null);
+    } catch (NoSuchFieldException exception) {
```
ejona86: Instead of catching `NoSuchFieldException`, `IllegalAccessException`, `NoSuchMethodException`, and `InvocationTargetException`, would it be better to just catch `ReflectiveOperationException`?

oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java
```diff
@@ -141,6 +158,76 @@ private final File getWellKnownCredentialsFile() {
...
+      return null;
+    }
+    Exception innerException = null;
+    try {
+      Class<?> credentialClass = forName(APP_ENGINE_CREDENTIAL_CLASS);
+      Constructor<?> constructor = credentialClass
+          .getConstructor(Collection.class);
+      Collection<String> emptyScopes = Collections.emptyList();
+      return (GoogleCredentials) constructor.newInstance(emptyScopes);
+    } catch (ClassNotFoundException e) {
```
ejona86: Ditto above, but including `ClassNotFoundException` and `InstantiationException`.

ejona86:
@anthmgoogle LGTM

oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java
anthmgoogle: It was initially implemented that way. Unfortunate that causes compile errors in the version of the JDK consumed by Andriod so this breakout is required.

oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java
anthmgoogle: Same issue as above.

PR #21, opened 2015-07-14
Adds support for retrieving credentials in the Cloud Shell environment.

wilford:


ejona86:
@anthmgoogle, are you going to review this one?

anthmgoogle:
Yes, I'll review this today.

oauth2_http/java/com/google/auth/oauth2/CloudShellCredentials.java
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * OAuth2 credentials representing the built-in service account for Google Cloud Shell.
+ *
+ * <p>Fetches access tokens from the App Identity service.
+ */
+public class CloudShellCredentials extends GoogleCredentials {
+
+  private final static int ACCESS_TOKEN_INDEX = 2;
+  private final static int READ_TIMEOUT = 5000;
```
anthmgoogle: Perhaps add the units for readability here?

oauth2_http/java/com/google/auth/oauth2/CloudShellCredentials.java
```diff
@@ -0,0 +1,86 @@
...
+  protected final static String GET_AUTH_TOKEN_REQUEST = "2\n[]";
```
anthmgoogle: There might be opportunity to improve this name to clarify what the content means.

anthmgoogle:
Sorry for the delay. LGTM. A couple of minor naming suggestions.

anthmgoogle:
One more thought: it may be worth porting this to google-api-client-java as well since that library currently  also has an ADC implementation, but will have broader reach for a while.

aalexand:
> One more thought: it may be worth porting this to google-api-client-java as well since that library currently also has an ADC implementation, but will have broader reach for a while.

It looks like Dataflow SDK (which is the one we want to make sure working in the Cloud Shell now) uses com.google.api.client.googleapis.auth.oauth2.GoogleCredential, see https://github.com/GoogleCloudPlatform/DataflowJavaSDK/blob/master/sdk/src/main/java/com/google/cloud/dataflow/sdk/util/Credentials.java#L24.  That one is from google-api-client-java I believe?

Can one implementation just call into the other so that no code duplication is needed?

anthmgoogle:
Yes, most existing usage is google-api-client-java. Eventually the auto
code in google-auth-library-java will replace it.

Can one implementation just call into the other so that no code duplication
is needed?


This varies depending on the language. For Java, while it would be
theoretically possible to write a clunky wrapper to convert on credential
type into another, given that there is already an ADC implementation, I
would recommend just port it to both places. They both have a similar
structure in the implementation, so it should be relatively low cost to
port.

oauth2_http/java/com/google/auth/oauth2/CloudShellCredentials.java
wilford: Done

oauth2_http/java/com/google/auth/oauth2/CloudShellCredentials.java
wilford: Added a comment that explains the formatting.

wilford:
Added comments and updated name. 

I'll update google-api-client-java separately.

anthmgoogle:
LGTM.

anthmgoogle:
Let me know when you are ready for me to merge in the pull request.

wilford:
We're good to merge.  Thanks!

PR #23, opened 2015-07-15
fixing install for users without gpg keys

garrettjonesgoogle:
Moving maven-gpg-plugin into a profile, which is triggered with the performRelease attribute, so that it isn't triggered during a normal "mvn install" command.


anthmgoogle:
LGTM.

PR #24, opened 2015-07-18
Fixing https://github.com/google/google-auth-library-java/issues/22

garrettjonesgoogle:
Some additional stuff:
* Creating some constants so that hardcoded values aren't repeated
* Fixing a couple spelling errors
* Having MockTokenServerTransport extend HttpTransport directly, since it doesn't use anything from MockHttpTransport

### DIFF
--- a/oauth2_http/java/com/google/auth/http/AuthHttpConstants.java
+++ b/oauth2_http/java/com/google/auth/http/AuthHttpConstants.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2015, Google Inc. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met:
+ *
+ *    * Redistributions of source code must retain the above copyright
+ * notice, this list of conditions and the following disclaimer.
+ *    * Redistributions in binary form must reproduce the above
+ * copyright notice, this list of conditions and the following disclaimer
+ * in the documentation and/or other materials provided with the
+ * distribution.
+ *
+ *    * Neither the name of Google Inc. nor the names of its
+ * contributors may be used to endorse or promote products derived from
+ * this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package com.google.auth.http;
+
+/**
+ * Constants used for auth in http
+ */
+public class AuthHttpConstants {
+  /**
+   * HTTP "Bearer" authentication scheme
+   */
+  public static final String BEARER = "Bearer";
+
+  /**
+   * HTTP "Authentication" header
+   */
+  public static final String AUTHORIZATION = "Authorization";
+}
--- a/oauth2_http/java/com/google/auth/http/HttpCredentialsAdapter.java
+++ b/oauth2_http/java/com/google/auth/http/HttpCredentialsAdapter.java
@@ -4,6 +4,9 @@ import com.google.auth.Credentials;
 import com.google.api.client.http.HttpHeaders;
 import com.google.api.client.http.HttpRequest;
 import com.google.api.client.http.HttpRequestInitializer;
+import com.google.api.client.http.HttpResponse;
+import com.google.api.client.http.HttpStatusCodes;
+import com.google.api.client.http.HttpUnsuccessfulResponseHandler;
 import com.google.api.client.util.Preconditions;
 
 import java.io.IOException;
@@ -11,11 +14,24 @@ import java.net.URI;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import java.util.regex.Pattern;
 
 /**
  * A wrapper for using Credentials with the Google API Client Libraries for Java with Http.
  */
-public class HttpCredentialsAdapter implements HttpRequestInitializer {
+public class HttpCredentialsAdapter
+    implements HttpRequestInitializer, HttpUnsuccessfulResponseHandler {
+
+  private static final Logger LOGGER = Logger.getLogger(HttpCredentialsAdapter.class.getName());
+
+  /**
+   * In case an abnormal HTTP response is received with {@code WWW-Authenticate} header, and its
+   * value contains this error pattern, we will try to refresh the token.
+   */
+  private static final Pattern INVALID_TOKEN_ERROR =
+      Pattern.compile("\\s*error\\s*=\\s*\"?invalid_token\"?");
 
   private final Credentials credentials;
 
@@ -27,12 +43,16 @@ public class HttpCredentialsAdapter implements HttpRequestInitializer {
     this.credentials = credentials;
   }
 
-    /**
-     * Initialize the HTTP request prior to execution.
-     *
-     * @param request HTTP request
-     */
+  /**
+   * {@inheritDoc}
+   *
+   * Initialize the HTTP request prior to execution.
+   *
+   * @param request HTTP request
+   */
   public void initialize(HttpRequest request) throws IOException {
+    request.setUnsuccessfulResponseHandler(this);
+
     if (!credentials.hasRequestMetadata()) {
       return;
     }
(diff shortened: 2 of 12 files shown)


oauth2_http/java/com/google/auth/http/AuthHttpConstants.java
```diff
@@ -0,0 +1,18 @@
+package com.google.auth.http;
+
+import java.util.regex.Pattern;
+
+/**
+ * Constants used for auth in http
+ */
+public class AuthHttpConstants {
```
anthmgoogle: These variables are kind of implementation related and should probably package-private. I'd recommend putting these in OAuth2Utils instead.

anthmgoogle: In general, we should keep public surface to a minimum, since it needs to be documented and impacts the overall digestibility and discover-ability.

oauth2_http/java/com/google/auth/http/HttpCredentialsAdapter.java
```diff
@@ -11,11 +14,17 @@
...
 import java.util.List;
 import java.util.Map;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 /**
  * A wrapper for using Credentials with the Google API Client Libraries for Java with Http.
  */
-public class HttpCredentialsAdapter implements HttpRequestInitializer {
+public class HttpCredentialsAdapter
```
anthmgoogle: The prevailing style convention is to to wrap on as few lines as possible, so consider wrapping in 2 lines instead of 3.

oauth2_http/java/com/google/auth/http/HttpCredentialsAdapter.java
```diff
@@ -44,16 +55,58 @@ public void initialize(HttpRequest request) throws IOException {
     }
     for (Map.Entry<String, List<String>> entry : credentialHeaders.entrySet()) {
       String headerName = entry.getKey();
-      List<String> requestValues = getHeadersValue(requestHeaders, headerName);
-      if (requestValues == null) {
-        requestValues = new ArrayList<String>();
-        requestHeaders.put(headerName, requestValues);
+      List<String> requestValues = new ArrayList<String>();
```
anthmgoogle: Is this related to the other changes. May be worth pulling this change into a separate request.

oauth2_http/java/com/google/auth/http/HttpCredentialsAdapter.java
```diff
@@ -44,16 +55,58 @@ public void initialize(HttpRequest request) throws IOException {
...
+      requestValues.addAll(entry.getValue());
+      requestHeaders.put(headerName, requestValues);
+    }
+  }
+
+  /**
+   * {@inheritDoc}
```
anthmgoogle: Do we need this inheritDoc on initialize as well?

oauth2_http/java/com/google/auth/http/HttpCredentialsAdapter.java
```diff
@@ -44,16 +55,58 @@ public void initialize(HttpRequest request) throws IOException {
...
+   * <p>
+   * Checks if {@code WWW-Authenticate} exists and contains a "Bearer" value
+   * (see <a href="http://tools.ietf.org/html/rfc6750#section-3.1">rfc6750 section 3.1</a> for more
+   * details). If so, it refreshes the token in case the error code contains
+   * {@code invalid_token}. If there is no "Bearer" in {@code WWW-Authenticate} and the status code
+   * is {@link HttpStatusCodes#STATUS_CODE_UNAUTHORIZED} it refreshes the token. If
+   * the token refresh throws an I/O exception, this implementation will log the
+   * exception and return {@code false}.
+   * </p>
+   * NOTE: mostly copied from Credential.java in google-oauth-java-client
```
anthmgoogle: I'd be tempted to leave this out. It is semi-true for a lot of code here. May be better to mention in a central readme.

oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
```diff
@@ -97,4 +98,15 @@ private final Long getExpiresInMilliseconds() {
       return (expirationTime.getTime() - clock.currentTimeMillis());
     }
   }
+
+  /**
+   * Override the access token. This is intended primarily to assist with
+   * testing, so that an initial token can be injected.
+   */
+  public void overrideAccessToken(AccessToken accessToken) {
```
anthmgoogle: I think we should instead do this by allowing the access token to be passed into a constructor. We should sync on this, as I have an interim change that does this.

oauth2_http/javatests/com/google/auth/http/HttpCredentialsAdapterTest.java
```diff
@@ -45,4 +57,77 @@ public void initialize_populatesOAuth2Credentials() throws IOException {
     String authorizationHeader = requestHeaders.getAuthorization();
     assertEquals(authorizationHeader, expectedAuthorization);
   }
+
+  public MockLowLevelHttpResponse makeErrorResponse() {
```
anthmgoogle: I would structure this slightly differently. Public helpers break up the readability of tests a bit. I'd suggest adding these to MockTokenServerTransport.

oauth2_http/javatests/com/google/auth/http/HttpCredentialsAdapterTest.java
```diff
@@ -45,4 +57,77 @@ public void initialize_populatesOAuth2Credentials() throws IOException {
...
+    tokenServerTransport.addClient(CLIENT_ID, CLIENT_SECRET);
+    tokenServerTransport.addRefreshToken(REFRESH_TOKEN, accessToken);
+
+    OAuth2Credentials credentials = new UserCredentials(
+        CLIENT_ID, CLIENT_SECRET, REFRESH_TOKEN, tokenServerTransport);
+    credentials.refresh();
+    HttpCredentialsAdapter adapter = new HttpCredentialsAdapter(credentials);
+
+    HttpTransport primaryHttpTransport = new MockHttpTransport() {
+        @Override
```
anthmgoogle: Related to the above, I'd suggest adding this as feature you can turn on in MockTokenServerTransport. Either way, there is benefit to getting a big dynamic class out of the test body to to make it easier to write more variations that use this custom handling code.

oauth2_http/javatests/com/google/auth/oauth2/MockTokenServerTransport.java
```diff
@@ -16,10 +17,12 @@
 import java.util.HashMap;
 import java.util.Map;
 
+import java.lang.UnsupportedOperationException;
+
 /**
  * Mock transport to simulate providing Google OAuth2 access tokens
  */
-public class MockTokenServerTransport extends MockHttpTransport {
+public class MockTokenServerTransport extends HttpTransport {
```
anthmgoogle: This seems a little fragile in that an error could accidentially cause test code to hit real functionality. Is there a way of doing this that retains it as a pure mock?

anthmgoogle:
Looking good. I've completed a review pass. One the question of putting an explicit access token in OAuth2Credentials, I have an-progress change we should coordinate on.

oauth2_http/java/com/google/auth/http/HttpCredentialsAdapter.java
garrettjonesgoogle: It is related because on retry, initialize is called again to reset the Authorization value.

oauth2_http/java/com/google/auth/http/HttpCredentialsAdapter.java
garrettjonesgoogle: Hmm yes, and actually they should both get the @Override annotation too.

oauth2_http/javatests/com/google/auth/http/HttpCredentialsAdapterTest.java
garrettjonesgoogle: Good idea, I'll move it into MockTokenServerTransport.

garrettjonesgoogle: As discussed offline, this is separate from serving tokens - I'll move it into a separate mock class distinct from MockTokenServerTransport.

oauth2_http/javatests/com/google/auth/oauth2/MockTokenServerTransport.java
garrettjonesgoogle: As discussed offline, I'll undo this change since it's not related to the rest of the changes. 

As for your concern, MockHttpTransport already extends HttpTransport, so this is only cutting out one layer of the hierarchy, which happens to not be used at all by MockTokenServerTransport.

oauth2_http/java/com/google/auth/http/HttpCredentialsAdapter.java
anthmgoogle: Actually, Eclipse can have compile errors if you use @Override on an interface method as opposed to a base class.

oauth2_http/java/com/google/auth/http/AuthHttpConstants.java
garrettjonesgoogle: Plan of action here:
* leave "Authorization" here
* make a constant "Bearer" here
* Make a package-private constant for "Bearer" + " " in each scope it is needed
* Move INVALID_TOKEN_ERROR to where it is used.

oauth2_http/java/com/google/auth/http/HttpCredentialsAdapter.java
```diff
@@ -11,11 +14,24 @@
...
+public class HttpCredentialsAdapter
+    implements HttpRequestInitializer, HttpUnsuccessfulResponseHandler {
+
+  private static final Logger LOGGER = Logger.getLogger(HttpCredentialsAdapter.class.getName());
+
+  /**
+   * In case an abnormal HTTP response is received with {@code WWW-Authenticate} header, and its
+   * value contains this error pattern, we will try to refresh the token.
+   */
+  public static final Pattern INVALID_TOKEN_ERROR =
```
anthmgoogle: Seems like this should be package-private or private.

anthmgoogle:
LGTM. Let's try to get a reviewer from library team on both first.

oauth2_http/java/com/google/auth/http/HttpCredentialsAdapter.java
garrettjonesgoogle: Good catch, that was an oversight on my part.

oauth2_http/java/com/google/auth/http/AuthHttpConstants.java
```diff
@@ -0,0 +1,16 @@
+package com.google.auth.http;
```
ejona86: Add license header.

oauth2_http/java/com/google/auth/http/HttpCredentialsAdapter.java
ejona86: I thought that @Override was not permitted for overriding interfaces in Java 5. In Java 6 it began to be permitted. Maybe Eclipse was configured with the wrong language version?

ejona86:
@anthmgoogle, currently the grpc client doesn't support retries, so we can't really duplicate the 401 logic now.

oauth2_http/java/com/google/auth/http/AuthHttpConstants.java
garrettjonesgoogle: Hmm, I opened up other files to copy their license header, and it looks like it's missing from most of the files. Under oauth2_http in particular, only ServiceAccountJwtAccessCredentials.java has it. Question: Do you want me to add the license header to just this file for this change? Or, not add it, but do a follow-up change adding the license header to all of the files?

ejona86: I could be okay with either. I'd probably assume having it correct in the initial commit for the new files in the PR. There seems to be little benefit in making the follow-up change larger. Your choice.

ejona86:
And @garrettjonesgoogle, I've not reviewed this closely enough to give a LGTM, but I did review it closely enough to have no concerns about it (you have my Approval :) ). Are y'all wanting a full review from me?

garrettjonesgoogle:
@ejona86, we don't need a full review at this time. Thanks!

oauth2_http/java/com/google/auth/http/HttpCredentialsAdapter.java
garrettjonesgoogle: I have confirmed - In Eclipse, @Override causes errors with java 1.5 compliance but works with 1.6.

PR #25, opened 2015-07-20
Add AccessToken to UserCredentials constructors.

anthmgoogle:
3-legged-auth has access token with optional refresh token. Breaking change to UserCredentials constructors.

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
@@ -106,6 +106,22 @@ public abstract class GoogleCredentials extends OAuth2Credentials {
   }
 
   /**
+   * Default constructor.
+   **/
+  public GoogleCredentials() {
+    this(null);
+  }
+
+  /**
+   * Constructor with explicit access token.
+   *
+   * @param accessToken Initial or temporary access token.
+   **/
+  public GoogleCredentials(AccessToken accessToken) {
+    super(accessToken);
+  }
+
+  /**
    * Indicates whether the credentials require scopes to be specified via a call to
    * {link GoogleCredentials#createScoped} before use.
    */
--- a/oauth2_http/java/com/google/auth/oauth2/UserCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/UserCredentials.java
@@ -36,39 +36,44 @@ public class UserCredentials extends GoogleCredentials {
    * @param refreshToken A refresh token resulting from a OAuth2 consent flow.
    */
   public UserCredentials(String clientId, String clientSecret, String refreshToken) {
-    this(clientId, clientSecret, refreshToken, null, null);
+    this(clientId, clientSecret, refreshToken, null, null, null);
   }
 
   /**
-   * Constructor allow override of transport for getting access tokens.
+   * Constructor to allow both refresh token and initial access token for 3LO scenarios.
    *
    * @param clientId Client ID of the credential from the console.
    * @param clientSecret Client ID of the credential from the console.
    * @param refreshToken A refresh token resulting from a OAuth2 consent flow.
-   * @param transport HTTP object used to get acces tokens.
+   * @param accessToken Initial or temporary access token.
    */
   public UserCredentials(
-      String clientId, String clientSecret, String refreshToken, HttpTransport transport) {
-    this(clientId, clientSecret, refreshToken, transport, null);
+      String clientId, String clientSecret, String refreshToken, AccessToken accessToken) {
+    this(clientId, clientSecret, refreshToken, accessToken, null, null);
   }
 
+
   /**
-   * Constructor allowing override of transport and URL for getting access tokens.
+   * Constructor with all parameters allowing custom transport and server URL.
    *
    * @param clientId Client ID of the credential from the console.
    * @param clientSecret Client ID of the credential from the console.
    * @param refreshToken A refresh token resulting from a OAuth2 consent flow.
-   * @param transport HTTP object used to get acces tokens.
-   * @param tokenServerUrl URL of the endpoint that provides tokens.
+   * @param accessToken Initial or temporary access token.
+   * @param transport HTTP object used to get access tokens.
+   * @param tokenServerUrl URL of the end point that provides tokens.
    */
   public UserCredentials(String clientId, String clientSecret, String refreshToken,
-      HttpTransport transport, GenericUrl tokenServerUrl) {
+      AccessToken accessToken, HttpTransport transport, GenericUrl tokenServerUrl) {
+    super(accessToken);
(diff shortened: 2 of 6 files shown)


anthmgoogle:
@garrettjonesgoogle and @ejona86 can you review?

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -106,6 +106,22 @@ public static GoogleCredentials fromStream(InputStream credentialsStream, HttpTr
   }
 
   /**
+   * Default constructor.
+   **/
+  public GoogleCredentials() {
+    this(null);
+  }
```
garrettjonesgoogle: Generally I've seen constructors come at the top of a class.

oauth2_http/java/com/google/auth/oauth2/UserCredentials.java
```diff
@@ -98,6 +103,10 @@ static UserCredentials fromJson(Map<String, Object> json, HttpTransport transpor
    */
   @Override
   public AccessToken refreshAccessToken() throws IOException {
+    if (refreshToken ==  null) {
+      throw new IOException("UserCredentials instance cannot refresh because no refresh token was"
+          +" specified in constructor.");
```
garrettjonesgoogle: add a space after '+'. Also, you might want to avoid referring to coincidental details of the implementation, such as the fact that currently, the only path to providing the refresh token is the constructor; maybe we might add other paths to setting it later.

oauth2_http/java/com/google/auth/oauth2/UserCredentials.java
```diff
@@ -36,39 +36,44 @@
...
-   * @param transport HTTP object used to get acces tokens.
-   * @param tokenServerUrl URL of the endpoint that provides tokens.
+   * @param accessToken Initial or temporary access token.
+   * @param transport HTTP object used to get access tokens.
+   * @param tokenServerUrl URL of the end point that provides tokens.
    */
   public UserCredentials(String clientId, String clientSecret, String refreshToken,
-      HttpTransport transport, GenericUrl tokenServerUrl) {
+      AccessToken accessToken, HttpTransport transport, GenericUrl tokenServerUrl) {
+    super(accessToken);
```
garrettjonesgoogle: Adding new parameters in the middle of a parameter list is potentially hazardous, especially when nulls are being thrown around.

oauth2_http/javatests/com/google/auth/oauth2/UserCredentialsTest.java
```diff
@@ -32,6 +33,15 @@
   private static final URI CALL_URI = URI.create("http://googleapis.com/testapi/v1/foo");
 
   @Test
+  public void constructor_accessAndRefreshTokenNull_throws() {
+    try {
+      new UserCredentials(CLIENT_ID, CLIENT_SECRET, null, null);
+      fail("Should not be able to refresh without refresh token.");
+    } catch (Exception expected) {
+    }
+  }
```
garrettjonesgoogle: Do we not have access to the version of junit that supports @Test(expected=Exception) ?

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
anthmgoogle: The prevailing convention from the Java V1 libraries was to have them after factories.

oauth2_http/java/com/google/auth/oauth2/UserCredentials.java
anthmgoogle: Agreed. Changed.

oauth2_http/java/com/google/auth/oauth2/UserCredentials.java
anthmgoogle: Agreed there is some risk here. Before locking down the AcessToken definitely belongs before the HttpTransport and GenericUrl, as the latter are only for testing or relatively rare extensible scenarios, while AccessToken is for the most common real scenario. We could have a discussion about whether the breaking change should be staged in some way. I'm inclined to go with this since the project is in alpha and it will break cleanly since the types are all different.

garrettjonesgoogle: Ok it seems like it should be fine since the project is in alpha.

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
garrettjonesgoogle: Ahh ok, this seems fine then.

oauth2_http/javatests/com/google/auth/oauth2/UserCredentialsTest.java
anthmgoogle: We do have that one. Changed this case. I left the case below but left it with the try/catch since there are multiple statements in the test that could cause it.

anthmgoogle:
Addressed code review feedback. One behavior change was to change the exception from refreshing the credentials with no refresh token into IllegalStateException which seems like a better fit. I can squash commits we are ready to submit.

garrettjonesgoogle:
LGTM.

oauth2_http/java/com/google/auth/oauth2/UserCredentials.java
```diff
@@ -36,39 +36,44 @@
...
     this.clientId = Preconditions.checkNotNull(clientId);
     this.clientSecret = Preconditions.checkNotNull(clientSecret);
-    this.refreshToken = Preconditions.checkNotNull(refreshToken);
+    this.refreshToken = refreshToken;
```
ejona86: Hmm... aren't the clientId and clientSecret unnecessary if you only have an access token (no refresh token)? Should there be a 1-arg constructor that just accepts an AccessToken?

ejona86:
@anthmgoogle, FYI, the `UserCredentials` constructor change is no problem for grpc, as we aren't using the class directly.

oauth2_http/java/com/google/auth/oauth2/UserCredentials.java
anthmgoogle: Yes, you can theoretically use a user access token without a client id or secret. I think to enable this, we want a different class, since if you only have an access token, then it is a general purpose temporary credential. To model this I think we should either have a TemporaryCredentials class, or we should change OAuth2Credenthals to be a non-abstract type where you can pass in an AccesToken directly. I was planning to do this in a subsequent change. But for 3LO scenarios, even without a refresh token, you would always have the Client ID and secret on hand because you need it get the access token.

ejona86: Okay, that works for me.

ejona86:
@anthmgoogle LGTM

PR #28, opened 2015-07-29
Updating version to 0.2.0

garrettjonesgoogle:


pom.xml
```diff
@@ -11,7 +11,7 @@
 
   <groupId>com.google.auth</groupId>
   <artifactId>google-auth-library-parent</artifactId>
-  <version>0.2.0-SNAPSHOT</version>
```
anthmgoogle: This is the right version change. It also needs to be updated at the top of the individual sub-component pom.xml files also. We have explored this before, and there is not a known way to share this value in a single field accross the poms.

anthmgoogle:
Need to update the other POMs also.

garrettjonesgoogle:
I have updated the subcomponent pom.xml files too now.

anthmgoogle:
LGTM. I don't think we needed anything else in this minor version, so merge when you are ready.

Once the version is updated, the next update before any other changes go in should be to change it to 0.3.0-SNAPSHOT.

PR #33, opened 2015-08-13
Standardize URI type for custom token server

anthmgoogle:
For UserCredentials and ServiceAccountCredentials.

Breaking change to long form constructors of UserCredentials and ServiceAccountCredentials.

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/OAuth2Utils.java
+++ b/oauth2_http/java/com/google/auth/oauth2/OAuth2Utils.java
@@ -10,6 +10,8 @@ import com.google.auth.http.AuthHttpConstants;
 
 import java.io.IOException;
 import java.math.BigDecimal;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.nio.charset.Charset;
 import java.util.Collection;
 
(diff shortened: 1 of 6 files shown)


oauth2_http/java/com/google/auth/oauth2/OAuth2Utils.java
```diff
@@ -17,7 +19,7 @@
  * Internal utilities for the com.google.auth.oauth2 namespace.
  */
 class OAuth2Utils {
-  static final String TOKEN_SERVER_URL = "https://accounts.google.com/o/oauth2/token";
+  static final URI TOKEN_SERVER_URI = parseStaticUri("https://accounts.google.com/o/oauth2/token");
```
ejona86: `URI.create()` throws a `RuntimeException` instead of a checked exception; it's what you want.

anthmgoogle: Yes, this is intentional as you can't initialize a static field with something that can throw a checked exception type. From checking Stack Overflow, this seems to be how most people work around the problem.

ejona86: That's not what I was getting at.

s/parseStaticUri/URI.create/

URI.create doesn't throw a checked exception, so is safe from a static field.

anthmgoogle: Got it. Changed to use URI.create and remove the helper method.

ejona86:
@anthmgoogle LGTM, assuming you want my review :)

PR #34, opened 2015-08-13
Add change listening support to OAuth2Credentials

anthmgoogle:


oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
```diff
@@ -94,6 +95,11 @@ public void refresh() throws IOException {
       requestMetadata = null;
       temporaryAccess = null;
       temporaryAccess = refreshAccessToken();
+      if (changeListeners != null) {
+        for (CredentialsChangedListener listener : changeListeners) {
```
ejona86: This is not thread-safe.

ejona86:
What was the primary use-case for the listener?

oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
anthmgoogle: This is happening under synchronized(lock) earlier in the method. This is similar to the V1 analog:
https://github.com/google/google-oauth-java-client/blob/dev/google-oauth-client/src/main/java/com/google/api/client/auth/oauth2/Credential.java#L485

anthmgoogle:
The primary use case is needed for the 3LO scenario. It stores an access token and refresh token in permanent storage, and needs to be able to write out an updated access token when it gets refreshed. The V1 Credential type also did this.

We have also talked about caching the access token for service accounts, which no code currently does, but we are potentially getting more access tokens than strictly needed by not caching them.

oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
ejona86: Doh. I missed that synchronized. This could still have problems with concurrent modification, but that seems unlikely given there isn't a `removeChangeListener`

ejona86:
Caching, that makes sense.

@anthmgoogle LGTM

PR #38, opened 2015-09-03
Fix broken JWTAccess payload

anthmgoogle:


ejona86:
@anthmgoogle LGTM. Thank you!

PR #39, opened 2015-09-03
3-Legged-OAuth2 base layer

anthmgoogle:


anthmgoogle:
@ejona8, @sqrrrl please review.

oauth2_http/java/com/google/auth/oauth2/ClientId.java
```diff
@@ -0,0 +1,143 @@
...
+  /**
+   * Constructs a Client ID from JSON file stored as a resource.
+   *
+   * @param relativeClass A class in the same namespace as the resource.
+   * @param resourceName The name of the resource
+   * @return The constructed ClientID instance based on the JSON in the resource.
+   * @throws IOException The JSON could not be loaded or parsed.
+   */
+  @SuppressWarnings("rawtypes")
+  public static ClientId fromResource(Class relativeClass, String resourceName) throws IOException {
```
ejona86: Just use Class<?> and you don't need the suppresswarnings.

anthmgoogle: Yes, fixed.

anthmgoogle:
Did a pass on all open comments and updated change to fix fromResource. @ejona8, @sqrrrl please let me know if anything further is required.

anthmgoogle:
Added one more precondition and test case.

anthmgoogle:
Fixed a problem that test resources only worked if you have built in eclipse.

ejona86:
@anthmgoogle, it LGTM, but I'm also not sure how much my review is worth for this.

sqrrrl:
Ditto. Base layer looks fine, was able to adapt to ruby with only a few minor differences in the storage layer (storing a little more info -- the client ID & scope.) Things diverge a little more at the web layer due to framework differences, but I think that's expected.

PR #46, opened 2015-11-17
change version to 0.3.1 to release bug fixes

pongad:


anthmgoogle:
LGTM. 

You should have contributor permissions so you should be able the merge the pull request whenever you are ready after you have an LGTM. We have found that it can be more flexible to allow contributors to control the timing. Let me know if there is any difficulty.

PR #51, opened 2015-11-19
Add release instructions

### LINKED ISSUE #49: Add Publishing Instructions

pongad:
Fixes #49

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,87 @@
+Release Instructions
+====================
+
+Set up Sonatype Account
+-----------------------
+* Sign up for a Sonatype JIRA account [here](https://issues.sonatype.org)
+* Click *Sign Up* in the login box, follow instructions
+
+Get access to repository
+------------------------
+* Go to [community support](https://issues.sonatype.org/browse/OSSRH)
+* Ask for publish rights by creating an issue similar to [this one](https://issues.sonatype.org/browse/OSSRH-16798)
+  * You must be logged in to create a new issue
+  * Use the *Create* button at the top tab
+
+Set up PGP keys
+---------------
+* Install GNU Privacy Guard (GPG)
+  * GPG is installed by default on Ubuntu systems
+  * For other systems, see [GnuPG download page](https://www.gnupg.org/download/)
+
+* Generate the key ```gpg --gen-key```
+
+  * Keep the defaults, but specify a passphrase
+  * The passphrase can be random; you just need to remember it long enough to finish the next step
+  * One way to make a random passphrase: ```base64 /dev/urandom | head -c20; echo;```
+
+* Find the ID of your public key ```gpg --list-secret-keys```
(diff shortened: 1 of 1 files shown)


RELEASE.md
```diff
@@ -0,0 +1,81 @@
...
+====================
+
+Set up Sonatype Account
+-----------------------
+* Sign up for a Sonatype JIRA account [here](https://issues.sonatype.org)
+* Click *Sign Up* in the login box, follow instructions
+
+Get access to repository
+------------------------
+* Go to [community support](https://issues.sonatype.org/browse/OSSRH)
```
anthmgoogle: When you navigate here, it is not clear how to create a new issue. Perhaps we could list more specific steps for how to create the request.

RELEASE.md
```diff
@@ -0,0 +1,81 @@
...
+* Ask for publish rights by creating an issue similar to [this one](https://issues.sonatype.org/browse/OSSRH-16798)
+
+Set up PGP keys
+---------------
+* Generate the key ```gpg --gen-key```
```
anthmgoogle: I remember a step for me was working out how to intall gpg. I don't think that is a requirement for running the default build any more, so we should include instructions or a link at how to install gpg.

RELEASE.md
```diff
@@ -0,0 +1,81 @@
...
+* Update all ```pom.xml``` files in the package to the release version you want. Sumit a pull request, get it reviewed, and submit
+* ```mvn clean install deploy -DperformRelease=true```
+* Verify the publishment [here](https://oss.sonatype.org/#nexus-search;quick~com.google.auth)
+  * If there is a problem, undo by ```mvn nexus-staging:drop```
+* ```mvn nexus-staging:release -DperformRelease=true```
+* Update all ```pom.xml``` files to the new snapshot version
+
+Publish the release
+-------------------
+* Go to [Sonatype](https://oss.sonatype.org/) and log in
```
anthmgoogle: If I recall, the above Deploy to Sonatype should also publish it, so this could incorrectly imply you need to also the manual step. I think the simplest thing might be to not have this set of separate manual steps under "Publish the Release". However, we should also check with @garrettjonesgoogle since he added this section to the source document.

anthmgoogle:
Finished first pass.

RELEASE.md
pongad: According to http://central.sonatype.org/pages/releasing-the-deployment.html , deployment puts the code in "a separate, temporary repository, that is private to your projects members". We need to publish it to make the change permanent.

RELEASE.md
pongad: done

RELEASE.md
pongad: done

pongad:
Updated per comments from @anthmgoogle .

anthmgoogle:
LGTM.

PR #59, opened 2016-03-04
Standardize the naming convention of base64-encoded binary data.

zhangkun83:
I am working on populating our internal (prod) authentication object onto wire in gRPC.  I am going to use `Credentials`, and because the authentication object is serialized as binary headers, I will base64-encode them.

The `Credentials` object will be passed to gRPC `ClientAuthInterceptor` and populated to gRPC `Metadata`, which accepts binary value (`byte[]`) when the key ends with `"-bin"`, and string value otherwise.  Therefore, `ClientAuthInterceptor` needs to decode the base64-encoded binary headers returned by `Crendentials`, and pass through the others.

`ClientAuthInterceptor` could use `"-bin"` as the indicator. However, because it consumes any `Crendentials` implementation, it's not safe for it to rely on this indicator unless it's defined on the `Credentials` interface.

This is a documentation-only no-op change.  I have talked with @ejona86, and he is OK with this change.

### DIFF
--- a/credentials/java/com/google/auth/Credentials.java
+++ b/credentials/java/com/google/auth/Credentials.java
@@ -26,6 +26,9 @@ public abstract class Credentials {
    * populated in headers or other context. The operation can block and fail to complete and may do
    * things such as refreshing access tokens.
    *
+   * <p>The convention for handling binary data is for the key in the returned map to end with
+   * {@code "-bin"} and for the corresponding values to be base64 encoded.
+   *
    * @throws IOException if there was an error getting up-to-date access.
    */
   public Map<String, List<String>> getRequestMetadata() throws IOException {
@@ -38,6 +41,9 @@ public abstract class Credentials {
    * <p>This should be called by the transport layer on each request, and the data should be
    * populated in headers or other context. The operation can block and fail to complete and may do
    * things such as refreshing access tokens.
+   *
+   * <p>The convention for handling binary data is for the key in the returned map to end with
+   * {@code "-bin"} and for the corresponding values to be base64 encoded.
    *
    * @param uri URI of the entry point for the request.
    * @throws IOException if there was an error getting up-to-date access.


ejona86:
@anthmgoogle, how do you feel about this?

credentials/java/com/google/auth/Credentials.java
```diff
@@ -26,6 +26,8 @@
    * populated in headers or other context. The operation can block and fail to complete and may do
    * things such as refreshing access tokens.
    *
+   * <p>If a key ends with {@code "-bin"}, its values are base64-encoded binary data.
+   *
    * @throws IOException if there was an error getting up-to-date access.
```
anthmgoogle: It is not clear how this comment relates to the inputs and outputs of this abstract method. Can you clarify which specific  values have this convention and what that means. It may be that this convention should be described somewhere other than the getRequestMetadata method.

anthmgoogle:
I think at minimum the comment needs more clarify, and potentially belongs in a different location.

credentials/java/com/google/auth/Credentials.java
zhangkun83: The goal is to plumb binary data from `Crendentials` to its consumer.

The idea is that if a key ends with `-bin`, e.g., `euc-proto-bin`, its values are __guaranteed__ (by `Credentials` implementations) to be __valid base64-encoded binary data__, so that the consumer of the values (in my case [ClientAuthInterceptor](https://github.com/grpc/grpc-java/blob/master/auth/src/main/java/io/grpc/auth/ClientAuthInterceptor.java)) can use base64 to decode it and get the binary data. Without such guarantee from the `Crendentials` interface, `ClientAuthInterceptor` wouldn't have a way to plumb binary data from `Crendentials`. It's a contract between the implementations and the users of this method. I can't think of a better place for this to be.

anthmgoogle: Perhaps. "The convention for handling binary data is for the key in returned map to end with "-bin" and for the corresponding values to be base64 encoded."

zhangkun83:
I have updated the PR per comment. PTAL.

credentials/java/com/google/auth/Credentials.java
```diff
@@ -26,6 +26,9 @@
    * populated in headers or other context. The operation can block and fail to complete and may do
    * things such as refreshing access tokens.
    *
+   * <p>The convention for handling binary data is for the key in returned map to end with {@code
```
anthmgoogle: Gammar nit: "key in the returned map"?

anthmgoogle:
Nit with the grammar, but other than that LGTM.

PR #64, opened 2016-06-14
Add asynchronous getRequestMetadata.

### LINKED ISSUE #3: Cannot perform asynchronous refresh of access tokens prior to expiration.
The interface of OAuth2Credentials does not allow an external class to know the expiration time of the currently cached access token. This prevents implementation of asynchronous refresh patterns.

Currently callers must use

credentials.getRequestMetadata()

which will block if the token has expired. This is not ideal for high-performance clients.

zhangkun83:
Resolves #3

@louiscryan @ejona86

### DIFF
--- a/credentials/java/com/google/auth/Credentials.java
+++ b/credentials/java/com/google/auth/Credentials.java
@@ -58,7 +58,7 @@ public abstract class Credentials {
           Map<String, List<String>> result;
           try {
             result = getRequestMetadata(uri);
-          } catch (IOException e) {
+          } catch (Throwable e) {
             callback.onFailure(e);
             return;
           }
--- a/credentials/java/com/google/auth/Credentials.java
+++ b/credentials/java/com/google/auth/Credentials.java
@@ -40,9 +40,9 @@ public abstract class Credentials {
    * Get the current request metadata without blocking.
    *
    * <p>This should be called by the transport layer on each request, and the data should be
-   * populated in headers or other context. The implementation can either call the callback inline,
-   * or from the given executor asynchronously. Either way it should <strong>never block</strong> in
-   * this method.
+   * populated in headers or other context. The implementation can either call the callback inline
+   * or asynchronously. Either way it should <strong>never block</strong> in this method. The
+   * executor is provided for tasks that may block.
    *
    * <p>The default implementation will just call {@link #getRequestMetadata(URI)} then the callback
    * from the given executor.
--- a/credentials/java/com/google/auth/Credentials.java
+++ b/credentials/java/com/google/auth/Credentials.java
@@ -50,16 +50,19 @@ public abstract class Credentials {
    * <p>The convention for handling binary data is for the key in the returned map to end with
    * {@code "-bin"} and for the corresponding values to be base64 encoded.
    */
-  public void getRequestMetadata(final URI uri, Executor executor, final Callback callback) {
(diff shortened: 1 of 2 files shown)


credentials/java/com/google/auth/Credentials.java
```diff
@@ -77,4 +106,20 @@
    * @throws IOException if there was an error getting up-to-date access.
    */
   public abstract void refresh() throws IOException;
+
+  /**
+   * The callback that receives the result of the asynchronous {@link #getRequestMetadata(URI,
+   * Executor, Callback)}. Exactly one method should be called.
+   */
+  public interface Callback {
```
anthmgoogle: Would this make sense as a a top level type with a name like RequestMetadataCallback? Unless they are very obscure it be better to avoid nested types.

anthmgoogle:
LGTM. We should get either @louiscryan or @ejona86 to sign off on this also.

credentials/java/com/google/auth/Credentials.java
zhangkun83: I am fine with top-level callback. @ejona86 WDYT?

ejona86: I'm okay with using RequestMetadataCallback. I would generally prefer the inner class (it isn't nested, since it is static) in this case, but it's also bikeshedding. Given how few things are in the com.google.auth package, I expect it doesn't matter much one way or another here.

credentials/java/com/google/auth/Credentials.java
```diff
@@ -36,6 +37,34 @@
...
+   * <p>The convention for handling binary data is for the key in the returned map to end with
+   * {@code "-bin"} and for the corresponding values to be base64 encoded.
+   */
+  public void getRequestMetadata(final URI uri, Executor executor, final Callback callback) {
+    executor.execute(new Runnable() {
+        @Override
+        public void run() {
+          try {
+            Map<String, List<String>> result = getRequestMetadata(uri);
+            callback.onSuccess(result);
```
ejona86: If the callback throws, should that call onFailure? Hmm... tough call. I think I'd prefer it propagating to the `executor` simply so that each implementation isn't supposed to worry about the issue.

ejona86:
LGTM.

credentials/java/com/google/auth/Credentials.java
zhangkun83: Changed to let it propagate.

credentials/java/com/google/auth/Credentials.java
```diff
@@ -36,6 +37,37 @@
   }
 
   /**
+   * Get the current request metadata without blocking.
+   *
+   * <p>This should be called by the transport layer on each request, and the data should be
+   * populated in headers or other context. The implementation can either call the callback inline,
+   * or from the given executor asynchronously. Either way it should <strong>never block</strong> in
```
ejona86: How about removing "from the given executor"? Just "or asynchronously. There shouldn't be a requirement that the executor is used when doing async.

ejona86:
@zhangkun83 still LGTM

credentials/java/com/google/auth/Credentials.java
zhangkun83: Rephrased.

ejona86:
@zhangkun83 LGTM

zhangkun83:
@ejona86 could you merge it?

PR #65, opened 2016-08-19
Implement the async method in oauth2 creds and JWT creds.

zhangkun83:
The OAuth2 creds will call the callback in network thread if cache is used. Otherwise, will do the blocking IO in the provided executor.

The JWT creds always calls the callback in network thread because it's only local CPU work.

### DIFF
--- a/oauth2_http/javatests/com/google/auth/oauth2/MockExecutor.java
+++ b/oauth2_http/javatests/com/google/auth/oauth2/MockExecutor.java
@@ -1,0 +1,40 @@
+package com.google.auth.oauth2;
+
+import java.util.LinkedList;
+import java.util.concurrent.Executor;
+
+/**
+ * Mock thread-less executor.
+ */
+public final class MockExecutor implements Executor {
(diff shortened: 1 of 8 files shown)


zhangkun83:
Please review the PR as a whole.

zhangkun83:
@ejona86

oauth2_http/javatests/com/google/auth/oauth2/MockExecutor.java
```diff
@@ -0,0 +1,40 @@
+package com.google.auth.oauth2;
```
ejona86: This and other new files need copyright header.

ejona86:
@zhangkun83 LGTM

oauth2_http/javatests/com/google/auth/oauth2/MockExecutor.java
zhangkun83: Done.

ejona86:
@zhangkun83 LGTM

Issue #1, opened 2015-02-13
Add License Headers to Java Files


anthmgoogle:
From feedback from initial port:
You also need license headers in the java files.

mziccard:
This was fixed in #66

PR #66, opened 2016-09-14
Add missing license headers and automatic checkstyle checks

mziccard:
This adds missing license headers.

It also adds automatic header checks on `mvn verify` (as well as `mvn test`). `verify` will fail if some java files are missing the license header.

### DIFF
--- a/credentials/java/com/google/auth/Credentials.java
+++ b/credentials/java/com/google/auth/Credentials.java
@@ -1,3 +1,34 @@
+/*
+ * Copyright 2015, Google Inc. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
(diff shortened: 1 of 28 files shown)


lesv:
@anthmgoogle I thought I saw a note that you were changing the way this repo was updated (from internal to external) instead of this way.

anthmgoogle:
LGTM.

Note to @lesv, there is no plan to change this from GitHub being the source of truth. There is one internal change on the the other side as a one-off, but this is still where we would encourage edits.

garrettjonesgoogle:
LGTM

mziccard:
I don't have access rights to merge PRs, can you guys merge it for me? Or grant me write access to the repository?

mziccard:
I believe this will fix #1

lesv:
merged

Issue #68, opened 2016-09-23
Using guava instead of guava-jdk5 dependency


mziccard:
We now depend on `guava-jdk5:13.0`. In both `google-cloud-java` and `gax-java` we instead depend on standard `guava` (version 19 and 18 respectively).

Is there any reason for depending on `guava-jdk5`? If not excluded (as we do in `google-cloud-java`), this just results in a bunch of duplicated classes in all projects that use standard `guava`. /cc @anthmgoogle @garrettjonesgoogle

mziccard:
This was fixed with #69

suvodeep-pyne:
This is still an issue.
`1.23.0` depends on `guava-jdk5:17.0` and it causes dependency conflicts with dropwizard `1.2.0` which uses the standard guava `com.google.guava:guava:23.1-jre`

Unfortunately, the method Throwables.throwIfUnchecked(e) doesn't exist in `guava-jdk5:17.0` causing failures in tests.

Could you guys please take a look. The Google client library is the recommended way for authenticating with firebase and we don't want to switch to the legacy API because of this problem.

For readers, this will fix the issue if you are using Gradle
```
compile (group: 'com.google.api-client', name: 'google-api-client', version: '1.23.0') {
  // excludes all guava packages including com.google.guava:guava-jdk5:17.0 which gets pulled in
  exclude group: 'com.google.guava'
}
```

martinbonnin:
Same problem here with 

`
    implementation (group: 'com.google.apis', name:'google-api-services-androidpublisher', version:'v2-rev47-1.23.0') {
    implementation (group: 'com.google.cloud', name: 'google-cloud-storage', version: '1.12.0')
`

I had to `exclude group: 'com.google.guava'` as @suvodeep-pyne mentioned

arunsankar8:
@suvodeep-pyne Thanks man, It worked

PR #69, opened 2016-09-26
Add HttpTransportFactory class, make credentials serializable, add tests

mziccard:


mziccard:
This also updates the `guava` dependency from `guava-jdk5:13.0` to `guava:19.0` (see issue #68).

/cc @garrettjonesgoogle @anthmgoogle

appengine/java/com/google/auth/appengine/AppEngineCredentials.java
```diff
@@ -31,41 +31,48 @@
 
 package com.google.auth.appengine;
 
-import com.google.auth.oauth2.AccessToken;
-import com.google.auth.oauth2.GoogleCredentials;
-import com.google.common.collect.ImmutableList;
 import com.google.appengine.api.appidentity.AppIdentityService;
 import com.google.appengine.api.appidentity.AppIdentityService.GetAccessTokenResult;
 import com.google.appengine.api.appidentity.AppIdentityServiceFactory;
+import com.google.appengine.repackaged.com.google.common.collect.ImmutableSet;
```
garrettjonesgoogle: Don't use repackaged classes

oauth2_http/java/com/google/auth/oauth2/ServiceAccountJwtAccessCredentials.java
```diff
@@ -64,8 +68,8 @@
   private final String privateKeyId;
   private final URI defaultAudience;
 
-  // Allow clock to be overriden by test code
-  Clock clock = Clock.SYSTEM;
+  // Until we expose this to the users it can remain transient and non-serializable
+  transient Clock clock = Clock.SYSTEM;
```
garrettjonesgoogle: Should this have `@VisibleForTesting`?

oauth2_http/java/com/google/auth/oauth2/UserCredentials.java
```diff
@@ -155,4 +166,40 @@ public final String getClientSecret() {
...
+  public int hashCode() {
+    return Objects.hash(super.hashCode(), clientId, clientSecret, refreshToken, tokenServerUri,
+        transportFactoryClassName);
+  }
+
+  @Override
+  public String toString() {
+    return toStringHelper()
+        .add("clientId", clientId)
+        .add("clientSecret", clientSecret)
```
garrettjonesgoogle: is the client secret a private key?

mziccard:
Also related to #68 we should decide the java version we want to support adding something like the following to the `pom.xml`:
```xml
      <plugin>
        <artifactId>maven-compiler-plugin</artifactId>
        <version>3.5.1</version>
        <configuration>
          <source>1.7</source>
          <target>1.7</target>
          <encoding>UTF-8</encoding>
          <compilerArgument>-Xlint:unchecked</compilerArgument>
        </configuration>
      </plugin>
```

Right now we are not specifying any java version, which defaults us to 1.5. This is a very old version and prevents from using features like try-with-resources and diamond operator. Any reason we are targeting 1.5? Can we explicitly target 1.7 instead?

oauth2_http/java/com/google/auth/oauth2/UserCredentials.java
mziccard: You are right we should not output this. Fixed

oauth2_http/java/com/google/auth/oauth2/ServiceAccountJwtAccessCredentials.java
mziccard: Fixed

appengine/java/com/google/auth/appengine/AppEngineCredentials.java
mziccard: Fixed

garrettjonesgoogle:
The change LGTM. It might be good to have one more reviewer. 

Regarding Java 1.5 vs 1.7 - @anthmgoogle should have more background on that than me.

oauth2_http/javatests/com/google/auth/oauth2/ServiceAccountCredentialsTest.java
```diff
@@ -156,6 +162,134 @@ public void getScopes_nullReturnsEmpty() throws IOException {
...
+        tokenServer);
+    OAuth2Credentials otherCredentials = ServiceAccountCredentials.fromPkcs8(SA_CLIENT_ID,
+        SA_CLIENT_EMAIL, SA_PRIVATE_KEY_PKCS8, SA_PRIVATE_KEY_ID, SCOPES, transportFactory,
+        tokenServer);
+    assertTrue(credentials.equals(otherCredentials));
+    assertTrue(otherCredentials.equals(credentials));
+  }
+
+  @Test
+  public void equals_false() throws IOException {
```
mziccard: I am fine with smaller tests. I am happy to split them, will take care of it before merging.

oauth2_http/javatests/com/google/auth/oauth2/BaseSerializationTest.java
```diff
@@ -0,0 +1,64 @@
...
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
+
+/**
+ * Base class for serialization tests.
+ */
+public class BaseSerializationTest {
+
+  @SuppressWarnings("unchecked")
+  public <T> T serializeAndDeserialize(T obj) throws IOException, ClassNotFoundException {
```
mziccard: You mean reusing this code in `AppEngineCredentialsTest`? To do that we need to introduce a dependency among test artifacts. This is completely fine (we do that in `google-cloud-java`), just wanted everybody to be aware of it.

mziccard:
@anthmgoogle Thanks for the comments. Are you fine with using `guava 19.0` instead of `guava-jdk`? Also, can we be explicit in targeting java 7 or are there requirements to support older versions? We could use try-with-resources and diamond operators to make our code look cleaner.

anthmgoogle:
@mziccard regarding Java versions and dependencies: The original plan with dependencies for this library was to be consistent with the google-oauth-java-client and related libraries, which were constrained to older versions of Java because of the need to work on Android. For example, 2 years ago you could not using Java compiler 1.7 for that reason and had to stick to 1.6. I don't know if there is any existing use on Android, or if these constraints still apply. To play it safe I would keep the compiler version at 1.6 unless we confirm that 1.7 works on Android now. 

No concerns I'm aware of on the Guava version.

lesv:
https://developer.android.com/guide/platform/j8-jack.html
I'm not recommending a change, just providing data.

mziccard:
From @lesv's link it seems that java7 should be safe to use. Also doing some digging in the library I found out it already uses classes introduced only since 1.7, see [StandardCharsets usage](https://github.com/google/google-auth-library-java/blob/master/oauth2_http/java/com/google/auth/oauth2/ClientId.java#L113), for instance.

anthmgoogle:
Cool. No concerns with these dependencies then.

It sounds like we need to tread more carefully with Java 8.

mziccard:
I added 4 commits that fix all the comments. PTAL.

mziccard:
Can you guys merge this PR? I still don't have write access here. Write access would also allow me to setup Travis. Do you know who has the rights to add users to this repo?

anthmgoogle:
I am one of the admins. I have given you admin access.

mziccard:
Thank you @anthmgoogle

PR #73, opened 2016-10-21
Add signing capabilities to AppEngine and service account classes

mziccard:
This PR does the following:

- Add `ServiceAccountSigner` interface and related tests
- Add signing capabilities to `AppEngineCredentials` class in `appengine` package
- Add signing capabilities to service account and AE credentials in `oauth` package

### DIFF
--- a/appengine/java/com/google/auth/appengine/AppEngineCredentials.java
+++ b/appengine/java/com/google/auth/appengine/AppEngineCredentials.java
@@ -34,6 +34,7 @@ package com.google.auth.appengine;
 import com.google.appengine.api.appidentity.AppIdentityService;
 import com.google.appengine.api.appidentity.AppIdentityService.GetAccessTokenResult;
 import com.google.appengine.api.appidentity.AppIdentityServiceFactory;
+import com.google.auth.ServiceAccountSigner;
 import com.google.auth.oauth2.AccessToken;
 import com.google.auth.oauth2.GoogleCredentials;
 import com.google.common.base.MoreObjects;
@@ -51,7 +52,7 @@ import java.util.Objects;
  *
  * <p>Fetches access tokens from the App Identity service.
  */
-public class AppEngineCredentials extends GoogleCredentials {
+public class AppEngineCredentials extends GoogleCredentials implements ServiceAccountSigner {
 
   private static final long serialVersionUID = -2627708355455064660L;
 
@@ -95,6 +96,16 @@ public class AppEngineCredentials extends GoogleCredentials {
   @Override
   public GoogleCredentials createScoped(Collection<String> scopes) {
     return new AppEngineCredentials(scopes, appIdentityService);
+  }
+
+  @Override
+  public String getAccount() {
+    return appIdentityService.getServiceAccountName();
+  }
+
+  @Override
+  public byte[] sign(byte[] toSign) {
+    return appIdentityService.signForApp(toSign).getSignature();
   }
 
   @Override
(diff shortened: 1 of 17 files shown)


credentials/java/com/google/auth/ServiceAccountSigner.java
```diff
@@ -0,0 +1,82 @@
...
+
+import java.util.Objects;
+
+/**
+ * Interface for a service account signer. A signer for a service account is capable of signing
+ * bytes using the private key associated with its service account.
+ */
+public interface ServiceAccountSigner {
+
+  class SigningException extends RuntimeException {
```
garrettjonesgoogle: It doesn't quite make sense to me for an exception to be package private if it can be thrown in public methods - callers would have no way to catch it specifically if they wanted.

oauth2_http/javatests/com/google/auth/oauth2/ServiceAccountJwtAccessCredentialsTest.java
```diff
@@ -68,7 +73,7 @@
       "36680232662-vrd7ji19qe3nelgchd0ah2csanun6bnr.apps.googleusercontent.com";
   private final static String SA_PRIVATE_KEY_ID =
       "d84a4fefcf50791d4a90f2d7af17469d6282df9d";
-  static final String SA_PRIVATE_KEY_PKCS8 = "-----BEGIN PRIVATE KEY-----\n"
+  private static final String SA_PRIVATE_KEY_PKCS8 = "-----BEGIN PRIVATE KEY-----\n"
```
garrettjonesgoogle: I'd like to use consistent ordering of `final` and `static` here - either change yours or change the rest.

credentials/java/com/google/auth/ServiceAccountSigner.java
mziccard: The class is defined inside an interface, it's public by default. Adding `public` is just redundant here.

garrettjonesgoogle: Oh! I haven't personally encountered a class defined in an interface before, but of course that makes sense.

appengine/javatests/com/google/auth/appengine/MockAppIdentityService.java
```diff
@@ -72,7 +74,11 @@ public void setExpiration(Date expiration) {
 
   @Override
   public SigningResult signForApp(byte[] signBlob) {
-    return null;
+    return signingResult;
+  }
+
+  public void setSignature(byte[] signature) {
+    this.signingResult = new SigningResult("keyName", signature);
```
garrettjonesgoogle: It looks like throughout this library, strings are defined as constants instead of being hardcoded - could you make a constant for this case?

appengine/java/com/google/auth/appengine/AppEngineCredentials.java
```diff
@@ -51,7 +52,7 @@
  *
  * <p>Fetches access tokens from the App Identity service.
  */
-public class AppEngineCredentials extends GoogleCredentials {
+public class AppEngineCredentials extends GoogleCredentials implements ServiceAccountSigner {
```
anthmgoogle: While not all credential types can sign, is it worth considering adding the signing methods and a boolean to check whether it is supported to the high level Credentials base type? If a credential can do it, the behavios is strongly tied to the identity.That might simplify the hierarchy, although the interface is certainly more clean and targeted. Interested in your thoughts on the pros/cons here.

mziccard: While I see why you are suggesting to move the `sign` method to the `GoogleCredentials` class I believe having a dedicated interface has some added value. Mainly because with a dedicated interface we can expose methods that take a `ServiceAccountSigner` and are granted that the provided parameter will support signing. Otherwise we would end up throwing runtime exceptions.

For instance in `google-cloud-java` we have the `Storage.signUrl` method that can take an option `SignUrlOption.signWith(ServiceAccountSigner)` to specify which account/credentials must be used to sign the URL.

PR #76, opened 2016-11-01
Add javadoc for version 0.5.0

mziccard:
This adds javadoc to gh-pages. This should not be merged until we release 0.5.0.


lesv:
LGTM - in principle - I'm wondering if we shouldn't publish this somewhere else, such as googlecloudplatform.github.io so it can actually be used / accessed? 

Or should I enlist a TW to get it published at d.g.c somewhere such as /identity ??

lesv:
HOLD BEFORE MERGING - I'd like to research something first.

lesv:
Is there a reason you prefer pushing this to the gh-pages branch, rather than to a [/docs](https://help.github.com/articles/configuring-a-publishing-source-for-github-pages/#publishing-your-github-pages-site-from-a-docs-folder-on-your-master-branch) folder?

Note - I'm ok, with the branch, but /docs seems like it might be better for long term maintenance.

mziccard:
@lesv I prefer to keep javadocs, sites, etc.. out of the codebase as they tend to grow in size when you preserve them for older versions (and we should preserve them for older versions).

As a contributor of the library I wouldn't want to check out several mb of javadoc (each version is ~1mb) along with the code I want to contribute to.

But this is just a personal preference and I am happy to change. Why do you think it would be better for long term maintenance?

anthmgoogle:
I'm in favor of having this github hosted. Since this particular library is much broader than Cloud, I would recommend hosting the doc with this repo rather than in the google-cloud docs. Looks like we need to wait for a release first.

lesv:
SGTM - I'm ok w/ doing it on the gh-pages.  Good to link from google-cloud-java to here.

PR #79, opened 2016-11-03
Update version to 0.6.0-SNAPSHOT

mziccard:


garrettjonesgoogle:
If we intend on doing a purely additive addition (the `fromStream` methods), do we want 0.5.1?

mziccard:
@garrettjonesgoogle I have no preference here. According to semver the PATCH number should be incremented for "backwards-compatible bug fixes" while the MINOR is for when you add functionality in a backwards-compatible manner. To be honest I believe that in other projects we haven't been very strict following semver. I jumped to `0.6.0-SNAPSHOT` following instructions at [RELEASING.md](https://github.com/google/google-auth-library-java/blob/master/RELEASE.md#deploy-to-sonatype):

> unless it's a bugfix release, we update from 0.4.0 to 0.5.0-SNAPSHOT

garrettjonesgoogle:
Well in general, I think < 1.0 is kind of fuzzy under semver anyway. I guess since this is noted in RELEASING.md, let's go ahead with 0.6.0. LGTM

PR #92, opened 2017-03-14
Minor cleanup removing unnecessary imports and suppress warnings

kinow:
As the title goes, there are a few unnecessary imports and suppress warnings annotations.


lesv:
Thank you for your submission.

PR #94, opened 2017-03-15
Look for CREDENTIAL_ENV_VAR in system properties when no envvar available

kinow:
Hello,

At the moment, the code looks for the CREDENTIAL_ENV_VAR (i.e. GOOGLE_APPLICATION_CREDENTIALS) in the environment variables.

When an application using, for example, Google Natural Language, is deployed to Amazon AWS, developers are able to specify only [system properties](http://docs.aws.amazon.com/elasticbeanstalk/latest/dg/java-configuration.html).

The easiest [way I found](http://kinoshita.eti.br/2017/03/15/using-google-natural-language-api-in-aws-elastic-beanstalk-application/) to deploy this application was creating a LanguageServiceSettings and creating scoped GoogleCredentials from a file manually.

In this pull request, the change is to fallback to system properties when there are no environment variables.

Hope that helps someone else with a similar problem.

Cheers
Bruno

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java
+++ b/oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java
@@ -122,6 +122,9 @@ class DefaultCredentialsProvider {
     // First try the environment variable
     GoogleCredentials credentials = null;
     String credentialsPath = getEnv(CREDENTIAL_ENV_VAR);
+    if (credentialsPath == null) {
+        credentialsPath = getProperty(CREDENTIAL_ENV_VAR, null);
+    }
     if (credentialsPath != null && credentialsPath.length() > 0) {
       InputStream credentialsStream = null;
       try {


garrettjonesgoogle:
@anthmgoogle , how does this interact with expectations of the ADC algorithm?

lesv:
@anthmgoogle This seems like a reasonable enhancement.  But since it changes the ADC algorithm, it would be nice to get your opinion.

lesv:
@garrettjonesgoogle Do you still have concerns?  Or is this Ok to merge?  (I'm going to do a release on monday, so I thought I'd pickup all the open PR's)

garrettjonesgoogle:
I want do get feedback from someone responsible for auth - @vchudnov-g do you have an opinion on this?

jonparrott:
I was okay with making a language-specific concession here and treat system properties equivalent to env vars, but after chatting with @lesv I do not feel like this is the right thing to do.

kinow:
@jonparrott 

>I was okay with making a language-specific concession here and treat system properties equivalent to env vars, but after chatting with @lesv I do not feel like this is the right thing to do.

Fair enough. Could you explain your point of view? That might be helpful to others that were not involved in this discussion, help users with similar requirement to understand it, plus educate the OP :-)

jonparrott:
I'm mostly following the guidance from @lesv to understand the Java perspective, so I'll let him comment on his reasoning here.

kinow:
Thanks @jonparrott ! My use case for that was Elastic Beanstalk with AWS, but right now I am using another solution. There's a workaround for this issue as well, which isn't really nice to have, but works.

So no hard feelings if this issue gets closed and marked as Won't Fix 👍 as long as we have the reason documented here for prosperity.

jonparrott:
Sounds good. We're definitely happy to be transparent about this. It was just easier to chat with @lesv instead of going back and forth in github comments. :)

lesv:
Hi @kinow,

Note - I'm totally open to discussion.  (And initially I was in favor of this, before Jon asked me slightly out of context about system properties)  I think I'll write several posts, so that I can think a bit between them.

In looking at [EBS](https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/environment-configuration-methods-after.html#configuration-options-after-ebcli), I see: `eb SETENV blah="hi"`

That really doesn't help you, as it doesn't actually send a file just the environment variable, which can still be hidden in a resource of a war or somewhere in your jar, but it's not all that clean.

In thinking about what you are asking, the questions should be:

1. What is the best practice to inject secrets into my app from another cloud provider?
2. Specifically can I do this with EBS?

The [ADC](https://developers.google.com/identity/protocols/application-default-credentials) algorithm is defined to:
a. Look at environment variable GOOGLE_APPLICATION_CREDENTIALS
b. Look for `gcloud auth application-default login`
c. Get a service account from the GCP Metadata server (GAE & GCP are a bit different, but it's the same idea)

You can always specify the service account JSON file when initializing any of the API's. (or at least you used to be able to - I didn't just check).

I don't see how using a service account gives you any real benefit over using a file and either envvar or adding it to your init's for the api.  That said, there was another proposal #144 to encode the service account json file into an EnvVar.  @jonparrott didn't like that, I didn't like it because it was overriding G_A_C and using it in an unsafe way.

We should, however, look to see if there is a way to safely and securely inject a secret like a service account JSON into a running on another cloud platform.  (item c above).  Unlike Jon, I'd be happy defining a way to encode the json file into an envvar as a way to inject our secret, just not overriding an existing one.  There may be another way. 

Thanks for bringing this up, as it's actually a very important item for us.  We are trying get to the point where our customers docker containers can run anywhere, on GCP, on a competing cloud provider, in their own datacenter, or locally in their laptop with just a change in the configuration.

kinow:
Hi @lesv 

Thanks a lot for the detailed explanation. I had a look at the Python, PHP, and NodeJS implementations, and they all seem to follow the same algorithm (though I could probably overwrite the PHP environment variable with `setenv` as it's a bit easier than with Java).

As it is defined in the algorithm, and since it could be hard to replicate this same behaviour in all languages, I can see the trouble in making an exception for the Java language to read system properties, or even for changing the algorithm.

I'm happy with the issue being closed as Won't Fix as I mentioned before. And I believe users now will have a good answer when/if someone asks for this feature again.

I am not using GCP right now (on a VMWare + AWS & OpenStack project), but if I had to use it again, it wouldn't be a blocker, as I have one workaround, and am sure there are other workarounds for this. I haven't checked, but maybe AWS changed or will change and allow developers to modify env vars for Elastic Beanstalk instances.

Thank heaps
Bruno

PR #100, opened 2017-04-26
Retry HTTP failures in ServiceAccountCredentials.refreshAccessToken

tcoffee-google:
Fix https://github.com/google/google-auth-library-java/issues/91.

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/ExponentialRetryAlgorithm.java
+++ b/oauth2_http/java/com/google/auth/oauth2/ExponentialRetryAlgorithm.java
@@ -1,0 +1,136 @@
+/*
+ * Copyright 2017, Google Inc. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met:
+ *
+ *    * Redistributions of source code must retain the above copyright
+ * notice, this list of conditions and the following disclaimer.
+ *    * Redistributions in binary form must reproduce the above
+ * copyright notice, this list of conditions and the following disclaimer
+ * in the documentation and/or other materials provided with the
+ * distribution.
+ *
+ *    * Neither the name of Google Inc. nor the names of its
+ * contributors may be used to endorse or promote products derived from
+ * this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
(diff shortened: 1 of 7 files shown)


tcoffee-google:
The retry implementation here borrows as closely from google-cloud-java as possible, to facilitate merging the duplication if dependencies are later restructured.

Efforts to establish test coverage for this functionality became rather involved. I will attempt to pursue this further if others feel it's needed.

lesv:
@garrettjonesgoogle @tcoffee-google I'm going to defer to Garrett on this.  Adding AutoValue may be problematic w/o at least shading it.  The same for Joda.

That said, I'll defer to Garrett.

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -312,11 +316,39 @@ public AccessToken refreshAccessToken() throws IOException {
...
+    TimedAttemptSettings attempt = retry.createFirstAttempt();
+
+    HttpResponse response = null;
+    int code = UNKNOWN_CODE;
+    while (retry.accept(attempt)) {
+      try {
+        try {
+          response = request.execute();
+          break;
+        } catch (SocketException | SocketTimeoutException exception) {
```
michaelbausor: Nit: document that we are retrying here, not just swallowing the exceptions.

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -312,11 +316,39 @@ public AccessToken refreshAccessToken() throws IOException {
...
+        } catch (HttpResponseException exception) {
+          code = exception.getStatusCode();
+          // Token endpoint does not return 429; timeouts return 500 instead of 408.
+          if (!(500 <= code && code < 600
```
michaelbausor: This conditional is hard to reason about - maybe split it up into multiple cases, or use parentheses to make the cases clearer.

oauth2_http/pom.xml
```diff
@@ -65,6 +65,16 @@
       <artifactId>junit</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>joda-time</groupId>
```
michaelbausor: GAX has converted to use ThreeTenBP instead of Joda - see this PR: https://github.com/googleapis/gax-java/pull/294

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -312,11 +316,39 @@ public AccessToken refreshAccessToken() throws IOException {
...
+                .setInitialRpcTimeout(Duration.millis(20000L))
+                .setRpcTimeoutMultiplier(1.0)
+                .setMaxRpcTimeout(Duration.millis(20000L))
+                .setTotalTimeout(Duration.millis(600000L))
+                .build());
+    TimedAttemptSettings attempt = retry.createFirstAttempt();
+
+    HttpResponse response = null;
+    int code = UNKNOWN_CODE;
+    while (retry.accept(attempt)) {
```
michaelbausor: If this returns false and `response` is still null, won't we have a null pointer exception on line 354? We need to handle the case when we should stop retrying.

oauth2_http/java/com/google/auth/oauth2/ExponentialRetryAlgorithm.java
```diff
@@ -0,0 +1,136 @@
...
+
+  private final RetrySettings globalSettings;
+
+  /**
+   * Creates a new exponential retry algorithm instance.
+   *
+   * @param globalSettings global retry settings (attempt independent)
+   * @throws NullPointerException if {@code globalSettings} is null
+   */
+  public ExponentialRetryAlgorithm(RetrySettings globalSettings) {
```
michaelbausor: Why are we removing the ApiClock parameter? That would make testing significantly easier I think.

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -312,11 +316,39 @@ public AccessToken refreshAccessToken() throws IOException {
     HttpRequest request = requestFactory.buildPostRequest(new GenericUrl(tokenServerUri), content);
     request.setParser(new JsonObjectParser(jsonFactory));
 
-    HttpResponse response;
-    try {
-      response = request.execute();
-    } catch (IOException e) {
-      throw new IOException("Error getting access token for service account: ", e);
+    ExponentialRetryAlgorithm retry =
```
michaelbausor: ExponentialRetryAlgorithm is thread-safe - we should not construct a new one every time.

Issue #90, opened 2017-03-07
Java is missing ability to do OAuth2 domain wide delegation


lesv:
This is needed for `google-auth-library-java` and google/google-api-java-client#1037 (creating an issue for both projects):

User's current way:
```java
GoogleCredential googleCredential = new GoogleCredential.Builder()
     .setTransport(TRANSPORT).setJsonFactory(JSON_FACTORY)
     .setServiceAccountId(emailAddress)
     .setServiceAccountPrivateKeyFromP12File(p12File)
     .setServiceAccountScopes(scopes)
     .setServiceAccountUser(user).build();
```

Would like: to do:
```java
GoogleCredential googleCredential = 
    GoogleCredential.fromStream(jsonInputStream, TRANSPORT, JSON_FACTORY).createScoped(Collections.singleton(Oauth2Scopes.USERINFO_EMAIL))
```

Python you just do:
```python
credentials = credentials.create_delegated(user)
```

tcoffee-google:
@vchudnov-g should this be closed?

PR #102, opened 2017-05-11
Add GoogleCredentials.createDelegated() method

vchudnov-g:
This change allows domain-wide delegation, addressing #90 

This PR also include additional tests. See commit descriptions.

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
@@ -181,14 +181,14 @@ public class GoogleCredentials extends OAuth2Credentials {
 
   /**
    * If the credentials support scopes, creates a copy of the the identity with the specified
-   * scopes; otherwise returns the same instance.
+   * scopes; otherwise, returns the same instance.
    */
   public GoogleCredentials createScoped(Collection<String> scopes) {
     return this;
   }
 
   /**
-   * If the credentials support domain-wide delgation, creates a copy
+   * If the credentials support domain-wide delegation, creates a copy
    * of the identity so that it impersonates the specified user;
    * otherwise, returns the same instance.
    */
--- a/oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
@@ -180,10 +180,19 @@ public class GoogleCredentials extends OAuth2Credentials {
   }
 
   /**
-   * If the credentials support scopes, create a copy of the the idenitity with the specified
-   * scopes, otherwise returns the same instance.
+   * If the credentials support scopes, creates a copy of the the identity with the specified
+   * scopes; otherwise returns the same instance.
    */
   public GoogleCredentials createScoped(Collection<String> scopes) {
+    return this;
+  }
+
+  /**
+   * If the credentials support domain-wide delgation, creates a copy
+   * of the identity so that it impersonates the specified user;
+   * otherwise, returns the same instance.
+   */
+  public GoogleCredentials createDelegated(String user) {
     return this;
   }
 }
--- a/oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
@@ -146,7 +146,7 @@ public class ServiceAccountCredentials extends GoogleCredentials implements Serv
    * @param serviceAccountUser Email of the user account to impersonate, if delegating domain-wide
    *        authority to the service account.
    */
-  public ServiceAccountCredentials(
+  ServiceAccountCredentials(
       String clientId, String clientEmail, PrivateKey privateKey, String privateKeyId,
       Collection<String> scopes, HttpTransportFactory transportFactory, URI tokenServerUri,
       String serviceAccountUser) {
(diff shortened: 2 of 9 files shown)


lesv:
LGTM to me, but I'd like @garrettjonesgoogle to look at it.

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -180,10 +180,19 @@ public boolean createScopedRequired() {
...
-   * scopes, otherwise returns the same instance.
+   * If the credentials support scopes, creates a copy of the the identity with the specified
+   * scopes; otherwise returns the same instance.
    */
   public GoogleCredentials createScoped(Collection<String> scopes) {
     return this;
   }
+
+  /**
+   * If the credentials support domain-wide delgation, creates a copy
```
garrettjonesgoogle: spelling (delgation)

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -180,10 +180,19 @@ public boolean createScopedRequired() {
   }
 
   /**
-   * If the credentials support scopes, create a copy of the the idenitity with the specified
-   * scopes, otherwise returns the same instance.
+   * If the credentials support scopes, creates a copy of the the identity with the specified
+   * scopes; otherwise returns the same instance.
```
garrettjonesgoogle: comma after "otherwise"

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -198,9 +222,33 @@ public static ServiceAccountCredentials fromPkcs8(
       String clientId, String clientEmail, String privateKeyPkcs8, String privateKeyId,
       Collection<String> scopes, HttpTransportFactory transportFactory, URI tokenServerUri)
       throws IOException {
+    return fromPkcs8(clientId, clientEmail, privateKeyPkcs8, privateKeyId, scopes, transportFactory, tokenServerUri, null);
+  }
+
+    /**
+   * Factory with miniumum identifying information and custom transport using PKCS#8 for the
```
garrettjonesgoogle: The `/**` line needs to be two spaces to the left

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -417,8 +456,28 @@ public boolean equals(Object obj) {
         && Objects.equals(this.scopes, other.scopes);
   }
 
-  private void readObject(ObjectInputStream input) throws IOException, ClassNotFoundException {
-    input.defaultReadObject();
-    transportFactory = newInstance(transportFactoryClassName);
+  protected String createAssertion(JsonFactory jsonFactory, long currentTime) throws IOException {
```
garrettjonesgoogle: `protected` is appropriate for when you want to allow a subclass to override the method, but you want to prevent code outside the subclass & outside the package from calling it; I don't think that's what you're intending though - it looks like you want to make it visible to the test. If that's the case, use package-private.

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -127,6 +128,28 @@ public ServiceAccountCredentials(
...
+   * @param transportFactory HTTP transport factory, creates the transport used to get access
+   *        tokens.
+   * @param tokenServerUri URI of the end point that provides tokens.
+   * @param serviceAccountUser Email of the user account to impersonate, if delegating domain-wide
+   *        authority to the service account.
+   */
+  public ServiceAccountCredentials(
+      String clientId, String clientEmail, PrivateKey privateKey, String privateKeyId,
+      Collection<String> scopes, HttpTransportFactory transportFactory, URI tokenServerUri,
+      String serviceAccountUser) {
```
garrettjonesgoogle: I think I want to avoid an explosion of constructors. We are stuck-ish with the existing ones, but maybe make your new one private because you really just want to use it from `fromPkcs8`. Although, we seem to also have an explosion of methods with that name, too. Maybe we could even hide that one - someone could use the existing `fromPkcs8` function and call `createDelegated` on the result to get what they want.

oauth2_http/javatests/com/google/auth/oauth2/ServiceAccountCredentialsTest.java
```diff
@@ -112,8 +117,52 @@ public void createdScoped_clones() throws IOException {
...
+
+  @Test
+  public void createAssertion_correct() throws IOException {
+    PrivateKey privateKey = ServiceAccountCredentials.privateKeyFromPkcs8(SA_PRIVATE_KEY_PKCS8);
+    List<String> scopes = Arrays.asList("scope1", "scope2");
+    ServiceAccountCredentials credentials = new ServiceAccountCredentials(
+        SA_CLIENT_ID, SA_CLIENT_EMAIL, privateKey, SA_PRIVATE_KEY_ID, scopes, null, null, SERVICE_ACCOUNT_USER);
+
+    JsonFactory jsonFactory = OAuth2Utils.JSON_FACTORY;
+    long currentTime = Clock.SYSTEM.currentTimeMillis();
```
garrettjonesgoogle: Call this `currentTimeMillis`

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
vchudnov-g: I can make it package-private to allow for testing. My concern, though, is that if the other constructors are public, then presumably there's a case for calling them directly and not using `fromPkcs8`. In that case, we omit the use case where there caller intends to set a `serviceAccountUser`. That said, I imagine the case for calling the constructor directly is probably infrequent.

How about I make this constructor non-public for now, and then if users express a need for accessing it directly, we make it public again?

garrettjonesgoogle: That's exactly what I am aiming for. If there are 10 candidates for package-private vs public, and only 1 of them is requested by anyone, we have avoided making 9 more methods public, which is a big win in my mind.

garrettjonesgoogle:
LGTM

PR #103, opened 2017-05-12
adding app engine credential check bypass using environment variable

jabubake:


jabubake:
@lesv , @garrettjonesgoogle PTAL

lesv:
Why?

lesv:
b/37359655 internal

Issue #91, opened 2017-03-10
Retry http failures in ServiceAccountCredentials.refreshAccessToken


garrettjonesgoogle:
As reported in https://github.com/GoogleCloudPlatform/google-cloud-java/issues/1545 , users have seen failures in `ServiceAccountCredentials.refreshAccessToken`, which has no retries on the http call. These failures should be retried. 

```
java.io.IOException: Error getting access token for service account:
         at com.google.auth.oauth2.ServiceAccountCredentials.refreshAccessToken(ServiceAccountCredentials.java:319)
         at com.google.auth.oauth2.OAuth2Credentials.refresh(OAuth2Credentials.java:149)
         at com.google.auth.oauth2.OAuth2Credentials.getRequestMetadata(OAuth2Credentials.java:135)
         at com.google.auth.http.HttpCredentialsAdapter.initialize(HttpCredentialsAdapter.java:96)
         at com.google.cloud.HttpServiceOptions$1.initialize(HttpServiceOptions.java:224)
         at com.google.api.client.http.HttpRequestFactory.buildRequest(HttpRequestFactory.java:93)
         at com.google.api.client.googleapis.services.AbstractGoogleClientRequest.executeUnparsed(AbstractGoogleClientRequest.java:423)
         at com.google.api.client.googleapis.services.AbstractGoogleClientRequest.executeUnparsed(AbstractGoogleClientRequest.java:352)
         at com.google.api.client.googleapis.services.AbstractGoogleClientRequest.execute(AbstractGoogleClientRequest.java:469)
         at com.google.cloud.storage.spi.DefaultStorageRpc.create(DefaultStorageRpc.java:245)
         ... 10 more
 Caused by: java.net.SocketTimeoutException: connect timed out
  ...
```

tcoffee-google:
Following testing investigations: Due to the limitations of various mechanisms available for dealing with timeouts in this context, it looks as though it will be exceedingly complicated to configure retrying in a way that's reliably compatible with the retrying specification used in GAX. Thus, for this use case, I'm planning to abandon this approach and instead use the distinct retrying functionality native to google-http-java-client.

PR #104, opened 2017-05-25
Retry HTTP failures in ServiceAccountCredentials.refreshAccessToken

tcoffee-google:
Addresses #91.

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
@@ -34,6 +34,9 @@ package com.google.auth.oauth2;
 import static com.google.common.base.MoreObjects.firstNonNull;
 
 import com.google.api.client.http.GenericUrl;
+import com.google.api.client.http.HttpBackOffIOExceptionHandler;
+import com.google.api.client.http.HttpBackOffUnsuccessfulResponseHandler;
+import com.google.api.client.http.HttpBackOffUnsuccessfulResponseHandler.BackOffRequired;
 import com.google.api.client.http.HttpRequest;
 import com.google.api.client.http.HttpRequestFactory;
 import com.google.api.client.http.HttpResponse;
@@ -43,6 +46,7 @@ import com.google.api.client.json.JsonFactory;
 import com.google.api.client.json.JsonObjectParser;
 import com.google.api.client.json.webtoken.JsonWebSignature;
 import com.google.api.client.json.webtoken.JsonWebToken;
+import com.google.api.client.util.ExponentialBackOff;
 import com.google.api.client.util.GenericData;
 import com.google.api.client.util.Joiner;
 import com.google.api.client.util.PemReader;
@@ -56,7 +60,6 @@ import com.google.common.collect.ImmutableSet;
 
 import java.io.IOException;
 import java.io.InputStream;
-import java.io.ObjectInputStream;
 import java.io.Reader;
 import java.io.StringReader;
 import java.net.URI;
@@ -339,6 +342,23 @@ public class ServiceAccountCredentials extends GoogleCredentials implements Serv
     HttpRequestFactory requestFactory = transportFactory.create().createRequestFactory();
     HttpRequest request = requestFactory.buildPostRequest(new GenericUrl(tokenServerUri), content);
     request.setParser(new JsonObjectParser(jsonFactory));
+
+    request.setIOExceptionHandler(new HttpBackOffIOExceptionHandler(new ExponentialBackOff()));
+    request.setUnsuccessfulResponseHandler(
+        new HttpBackOffUnsuccessfulResponseHandler(new ExponentialBackOff()).setBackOffRequired(
+            new BackOffRequired() {
+              public boolean isRequired(HttpResponse response) {
+                int code = response.getStatusCode();
+                return (
+                    // Server error --- includes timeout errors, which use 500 instead of 408
+                    code / 100 == 5
+                    // Forbidden error --- for historical reasons, used for rate_limit_exceeded
+                    // errors instead of 429, but there currently seems no robust automatic way to
+                    // distinguish these cases: see
+                    // https://github.com/google/google-api-java-client/issues/662
+                    || code == 403);
+              }
+            }));
 
     HttpResponse response;
     try {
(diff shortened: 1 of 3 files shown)


garrettjonesgoogle:
@vam-google could you take a first look?

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -340,6 +343,23 @@ public AccessToken refreshAccessToken() throws IOException {
     HttpRequest request = requestFactory.buildPostRequest(new GenericUrl(tokenServerUri), content);
     request.setParser(new JsonObjectParser(jsonFactory));
 
+    request.setIOExceptionHandler(new HttpBackOffIOExceptionHandler(new ExponentialBackOff()));
+    request.setUnsuccessfulResponseHandler(
+        new HttpBackOffUnsuccessfulResponseHandler(new ExponentialBackOff()).setBackOffRequired(
```
vam-google: It looks like for `ExponentialBackOff` the default elapsed time will be 15 minutes and default max interval 60 seconds. Both of them (especially the first one) seem very high. Please confirm that it is ok. If it is not, please override the default values.

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -340,6 +343,23 @@ public AccessToken refreshAccessToken() throws IOException {
     HttpRequest request = requestFactory.buildPostRequest(new GenericUrl(tokenServerUri), content);
     request.setParser(new JsonObjectParser(jsonFactory));
 
+    request.setIOExceptionHandler(new HttpBackOffIOExceptionHandler(new ExponentialBackOff()));
+    request.setUnsuccessfulResponseHandler(
```
vam-google: The original [issue](https://github.com/google/google-auth-library-java/issues/91) is about IOException. I'm not sure how IOException and http errors are connected in this case (if connected at all), but please confirm that both (not just one) IO exception and http error response retries are necessary here and that they do not conflict with each other.

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -340,6 +343,23 @@ public AccessToken refreshAccessToken() throws IOException {
...
+        new HttpBackOffUnsuccessfulResponseHandler(new ExponentialBackOff()).setBackOffRequired(
+            new BackOffRequired() {
+              public boolean isRequired(HttpResponse response) {
+                int code = response.getStatusCode();
+                return (
+                    // Server error --- includes timeout errors, which use 500 instead of 408
+                    code / 100 == 5
```
vam-google: This seems like an unnecessary trick. Taking into account that there is nothing above server errors in http standard (no 600+ error codes), please consider changing it to `code >= 500`

oauth2_http/javatests/com/google/auth/oauth2/ServiceAccountCredentialsTest.java
```diff
@@ -241,6 +242,55 @@ public void getRequestMetadata_customTokenServer_hasAccessToken() throws IOExcep
   }
 
   @Test
+  public void refreshAccessToken_retries() throws IOException {
```
vam-google: This looks like multiple tests in one. Ideally each test should test a single piece of functionality. Please consider splitting it into multiple test cases. For example, for this particular big test:
- no retry needed case
- IOException case
- Http error code case
- non-retriable error case

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
tcoffee-google: This is just a result of how the built-in error handlers for google-http-java-client are organized. HTTP errors will ultimately result in an `HttpResponseException`, which is a subclass of `IOException`, but the library has a separate handler for retrying HTTP errors, which is convenient since we do not retry all HTTP error codes. It is true that the original bug report concerned a `SocketTimeoutException`, which would be handled by the `IOExceptionHandler` alone, but it seems that the same sort of catastrophic failure could occur in the case of HTTP errors as well.

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
tcoffee-google: Here's a comparison of parameters with those used many places for gRPC calls in google-cloud-java. These seem to me to be in the same ballpark. I might generally expect longer settings for this use case since, in a typical scenario, it would be critical to all following operations.

Value | google-cloud-java standard | `ExponentialBackOff` default
------------ | ------------- | -------------
initial delay | 0.1 sec | 0.5 sec
base delay multiplier | 1.3 | 1.5
random delay multiplier | 0-1 | 0.5-1.5
max delay | 1 min | 1 min
initial timeout | 20 sec | none
timeout multiplier | 1 | none
max timeout | 20 sec | none
max total | 10 min | 15 min

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
tcoffee-google: Here I was just trying to be consistent with https://developers.google.com/api-client-library/java/google-http-java-client/reference/1.20.0/com/google/api/client/http/HttpBackOffUnsuccessfulResponseHandler.BackOffRequired#ON_SERVER_ERROR .

I would argue that, given the specification, the robust alternative would need to be `code >= 500 && code < 600`, which seems less readable as well as slower.

oauth2_http/javatests/com/google/auth/oauth2/ServiceAccountCredentialsTest.java
tcoffee-google: Adds a good deal of overhead, but okay.

tcoffee-google:
@vam-google PTAL when you get a chance.

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
vam-google: Wow, nice table =). Ok, agree that 10 min and 15 min are in the same ballpark.

PR #115, opened 2017-07-15
Post Release -> 0.7.2-SNAPSHOT

tcoffee-google:


garrettjonesgoogle:
For a snapshot version, I think the convention is to use the next patch version, so that would be `0.7.2-SNAPSHOT`.

tcoffee-google:
This has not been the convention for this repo, but I agree it's a more sensible rule. Done.

PR #116, opened 2017-07-17
Update release process for Javadoc update

tcoffee-google:


update_javadoc.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
```
lesv: Copyright & License

tcoffee-google: Done. Needs approval to submit.

Issue #117, opened 2017-08-07
FR: Expose Project ID from ServiceAccountCredentials


hiranya911:
Currently the `com.google.auth.oauth2.ServiceAccountCredentials` class does not handle the `project_id` field available in Google service account JSON files. It would be useful if the credential can read this field, and expose it via a new `getProjectId()` method.

Usecase: We want to use this library within some Firebase SDKs, and we (in some situations) need to know the project ID along with the credential.

chingor13:
Fixed in #118 and available since 0.8.0

PR #118, opened 2017-08-09
Extracting the project_id field from service account JSON files

hiranya911:
Related to #117


oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -181,18 +183,19 @@ static ServiceAccountCredentials fromJson(
     String clientEmail = (String) json.get("client_email");
     String privateKeyPkcs8 = (String) json.get("private_key");
     String privateKeyId = (String) json.get("private_key_id");
+    String projectId = (String) json.get("project_id");
```
lesv: Some service account files don't have project_id in them -- will this work for them?

If so, I'll give an LGTM and merge.

hiranya911: `json.get()` returns `null` if the key is not present, and the rest of the code accepts null project ID values. Added a test case to verify this scenario.

Issue #120, opened 2017-08-18
Possible Integer Overflow When Parsing expires_in Values


hiranya911:
While running some tests using the `MockTokenServerTransport` of the Google API client, I noticed that sometimes `GoogleCredentials.getExpirationTime()` incorrectly returns timestamps that are several days old. This is what's going on:

1. `MockTokenServerTransport` has a bug where it reports the `expires_in` field in milliseconds (3600000), where it should be seconds. I reported this in their Github repo (https://github.com/google/google-api-java-client/issues/1061).
2. The code in `ServiceAccountCredentials` class will encounter an integer overflow, when the backend server reports a large `expires_in` value:

https://github.com/google/google-auth-library-java/blob/master/oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java#L375

```
    long expiresAtMilliseconds = clock.currentTimeMillis() + expiresInSeconds * 1000;
```

If `expiresInSeconds` is large (say 3600000), this code will overflow, and the resulting timestamp will be in the past.

chingor13:
Fixed in #121 and available since 0.8.0

PR #121, opened 2017-08-21
Fixing an Integer Overflow Issue

hiranya911:
If the token server responds with a large `expires_in` value to the token refresh request, the `ServiceAccountCredential` class runs into an integer overflow. I've added more information at #120.

This patch fixes the issue, and provides a test case (test case fails without the fix). This is not a critical issue, as we can expect most token servers to be well behaved. But it would be nice if the credentials were able to handle the case where the server responds with a large `expires_in` value.

### DIFF
--- a/oauth2_http/javatests/com/google/auth/oauth2/ServiceAccountCredentialsTest.java
+++ b/oauth2_http/javatests/com/google/auth/oauth2/ServiceAccountCredentialsTest.java
@@ -44,6 +44,7 @@ import com.google.api.client.json.GenericJson;
 import com.google.api.client.json.JsonFactory;
 import com.google.api.client.json.webtoken.JsonWebSignature;
 import com.google.api.client.json.webtoken.JsonWebToken;
+import com.google.api.client.testing.http.FixedClock;
 import com.google.api.client.testing.http.MockLowLevelHttpResponse;
 import com.google.api.client.util.Clock;
 import com.google.api.client.util.Joiner;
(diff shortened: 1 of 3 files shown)


oauth2_http/javatests/com/google/auth/oauth2/ServiceAccountCredentialsTest.java
```diff
@@ -292,6 +293,34 @@ public void refreshAccessToken_refreshesToken() throws IOException {
...
+            null);
+    credentials.clock = new FixedClock(0L);
+
+    transport.addServiceAccount(SA_CLIENT_EMAIL, tokenString);
+    AccessToken accessToken = credentials.refreshAccessToken();
+    assertEquals(tokenString, accessToken.getTokenValue());
+    assertEquals(3600 * 1000L, accessToken.getExpirationTimeMillis().longValue());
+
+    // Test for large expires_in values (should not overflow).
+    transport.setExpiresInSeconds(3600 * 1000);
```
lesv: Shouldn't there be an L here as well?

hiranya911: `setExpiresInSeconds()` takes an int. In `MockTokenServerTransport`, this was set to the constant int value 3600. I kept it as an int when adding the `setExpiresInSeconds()` method.

lesv:
@garrettjonesgoogle I've merged, but can revert if you wish.

PR #122, opened 2017-08-30
use metadata server to get credentials for GAE 8 standard environment

neozwu:
GAE java 8 standard environment supports Java native threads.  However `getDefaultCredentials()` call will fail on GAE J8 if it is called in a native thread (i.e. threads not managed by appengine runtime) due to the inaccessibility of native threads to appengine APIs. For GAE java8, same as compute engine, metadata serve should be used to construct `Credentials`.


neozwu:
kindly ping @garrettjonesgoogle @lesv

Issue #127, opened 2017-09-11
OAuth2 over proxy doesn't work.


bullet-tooth:
I'm using [spanner java client](https://github.com/GoogleCloudPlatform/google-cloud-java/tree/master/google-cloud-spanner) with [gcloud SDK](https://cloud.google.com/sdk/gcloud/) and trying to connect to Google Spanner from machine that uses proxy for internet access.   
I have configured gcloud to use my proxy server and it is ok.
I have specified environment variable GRPC_PROXY_EXP and Spanner requests to **spanner.googleapis.com** go through proxy and it's ok as well.  

 But spanner sql requests fail with `com.google.cloud.spanner.SpannerException: UNAUTHENTICATED` exception because of another error related to OAuth2:
```
Caused by: java.net.UnknownHostException: accounts.google.com
	at java.net.AbstractPlainSocketImpl.connect(AbstractPlainSocketImpl.java:184)
	at java.net.SocksSocketImpl.connect(SocksSocketImpl.java:392)
	at java.net.Socket.connect(Socket.java:589)
	at sun.security.ssl.SSLSocketImpl.connect(SSLSocketImpl.java:668)
	at sun.net.NetworkClient.doConnect(NetworkClient.java:175)
	at sun.net.www.http.HttpClient.openServer(HttpClient.java:432)
	at sun.net.www.http.HttpClient.openServer(HttpClient.java:527)
	at sun.net.www.protocol.https.HttpsClient.<init>(HttpsClient.java:264)
	at sun.net.www.protocol.https.HttpsClient.New(HttpsClient.java:367)
	at sun.net.www.protocol.https.AbstractDelegateHttpsURLConnection.getNewHttpClient(AbstractDelegateHttpsURLConnection.java:191)
	at sun.net.www.protocol.http.HttpURLConnection.plainConnect0(HttpURLConnection.java:1105)
	at sun.net.www.protocol.http.HttpURLConnection.plainConnect(HttpURLConnection.java:999)
	at sun.net.www.protocol.https.AbstractDelegateHttpsURLConnection.connect(AbstractDelegateHttpsURLConnection.java:177)
	at sun.net.www.protocol.http.HttpURLConnection.getOutputStream0(HttpURLConnection.java:1283)
	at sun.net.www.protocol.http.HttpURLConnection.getOutputStream(HttpURLConnection.java:1258)
[...]

PR #128, opened 2017-09-12
 Added proxy support for OAuth2 requests.

bullet-tooth:
Implemented proxy support for HTTP and SOCKS.
Fix for issue #127


lesv:
@garrettjonesgoogle This is probably needed for some Enterprise environments, but we should get an expert to review as this might provide a vector for attack.  (I suspect it's Ok, as one would need access to the machine this code was running on, but I may not be thinking of a case.)

lesv:
@bullet-tooth Is this still important to you?

bullet-tooth:
Hi @lesv . We're using dev version with implemented proxy. So if you think that our case very rare and could cause security issues you can discard this PR.

lesv:
@jonparrott FYI
@bullet-tooth I'm going to close this for now.  I actually think this is a much bigger issue that we should address for all our client libraries.  This is an issue for many developers who need to use a proxy inside their companies.

PR #129, opened 2017-09-14
Implemented in-memory TokenStore and added opportunity to save user credentials into file

YuryBandarchuk16:
Firstly, I implemented in-memory TokenStore by just using HashMap.

After that I added some methods that allows to save UserCredentials to file.

Everything, that might be affected by these changes have been tests in appropriate test files.


oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -37,8 +37,7 @@
 import com.google.api.client.util.Preconditions;
 import com.google.auth.http.HttpTransportFactory;
 
-import java.io.IOException;
-import java.io.InputStream;
+import java.io.*;
```
lesv: We don't use *'s
https://google.github.io/styleguide/javaguide.html

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -166,6 +165,24 @@ public static GoogleCredentials fromStream(InputStream credentialsStream,
   }
 
   /**
+   * Saves the end user credentials in specified file
```
lesv: Why do you want this?  Didn't you get them from a file or some other mechanism?

What kind of credentials are you trying to save, many are short lived.

oauth2_http/java/com/google/auth/oauth2/UserCredentials.java
```diff
@@ -58,6 +58,8 @@
  */
 public class UserCredentials extends GoogleCredentials {
 
+  public static final String USER_CREDENTIALS_FILE_NAME = "GOOGLE_AUTH_USER_CREDENTIALS_FILE_NAME";
```
lesv: GOOGLE_APPLICATION_CREDENTIALS is the typical way we do this.

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
YuryBandarchuk16: fixed

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
YuryBandarchuk16: The main purpose is to store refresh tokens as described in this issue: https://github.com/google/google-auth-library-java/issues/54

YuryBandarchuk16:
Would you like me to use the memory token store in UserAuthorizer.java as default token store?

lesv:
Cool - Good to mention issues in the initial commit as then I can come up to speed on the PR.

I'm going to wait for @garrettjonesgoogle to comment before approving this.  But I'm happy with the changes.

I did as a Q, in GoogleCredentials.java, what kind of credentials are you writing?

YuryBandarchuk16:
@lesv In GoogleCredentials.java I am saving the following credentials (updated comment above the method): 
 

- RefreshToken
- ClientId
- ClientSecret
- ServerTokenUri

As in the issue, the main purpose is to store the refresh token.

YuryBandarchuk16:
@lesv In the latest test run there was a failure, because the test checked the following: in case TokenStore is not specified and you tried to get data from it, it will throw an exception. Right now, there is always a default implementation exists for TokenStore. It means that we don't need this test anymore. I removed it in  this commit: https://github.com/google/google-auth-library-java/pull/129/commits/5fcddeb6154fedacaa45545b4929cc4ccf7c8a34

lesv:
@YuryBandarchuk16 - @garrettjonesgoogle is out until at least next week.  Do you need this merged and released soon, or can hold it for a bit?

YuryBandarchuk16:
@lesv Sure, can hold, there is no rush.

YuryBandarchuk16:
@lesv are there any updates?

lesv:
@garrettjonesgoogle Do you have an opinion?  It seems like a reasonable change to me.

garrettjonesgoogle:
I will take a look.

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -166,6 +170,24 @@ public static GoogleCredentials fromStream(InputStream credentialsStream,
...
+   *
+   * @param credentials InputStream containing user credentials in JSON format
+   * @param credentialsFileName Name of file where to store the credentials
+   * @throws IOException An error saving the credentials.
+   */
+  public void saveCredentialsToFile(InputStream credentials, String credentialsFileName) throws IOException {
```
garrettjonesgoogle: It's odd to have a member function (which contains credentials) which also takes distinct credentials as an argument, and doesn't use any member data. The implementation in fact seems generic enough that you could just use other libraries for this, e.g. `FileUtils.copyInputStreamToFile` from Commons IO.

oauth2_http/javatests/com/google/auth/oauth2/UserCredentialsTest.java
```diff
@@ -51,9 +51,7 @@
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
 
-import java.io.ByteArrayInputStream;
-import java.io.IOException;
-import java.io.InputStream;
+import java.io.*;
```
garrettjonesgoogle: Don't use star imports

lesv:
@YuryBandarchuk16 Could you address Garrett's asks - we are trying to release this asap.

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
YuryBandarchuk16: I created the method in ```UserCredentials.java``` called ```saveUserCredentials``` which calls the generic method of it's super class called ```saveCredentialsToFile``` to save them into file.
I fixed the implementation and used ```FileUtils.copyInputStreamToFile``` in order to do this.

oauth2_http/javatests/com/google/auth/oauth2/UserCredentialsTest.java
YuryBandarchuk16: Fixed.

YuryBandarchuk16:
@lesv Sorry for the delay, thanks for letting me know. I think I have already fixed everything @garrettjonesgoogle asked me to fix.

oauth2_http/pom.xml
```diff
@@ -65,6 +65,15 @@
       <artifactId>junit</artifactId>
       <scope>test</scope>
     </dependency>
+      <dependency>
+          <groupId>com.google.guava</groupId>
+          <artifactId>guava</artifactId>
+      </dependency>
+    <dependency>
```
garrettjonesgoogle: Guava is already present in this pom, so remove this

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -36,9 +36,14 @@
 import com.google.api.client.json.JsonObjectParser;
 import com.google.api.client.util.Preconditions;
 import com.google.auth.http.HttpTransportFactory;
+import org.apache.commons.io.FileUtils;
```
garrettjonesgoogle: Could you figure out how to use Guava for this functionality instead of adding a second utility dependency?

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
garrettjonesgoogle: Sorry for guiding you in the wrong direction by having you add a dependency on Commons IO, then asking you to remove it...

In any case, what I meant was that you don't need to have this utility function at all, and you could just put the implementation at the calling site (for you, `UserCredentials.saveUserCredentials`)

oauth2_http/java/com/google/auth/oauth2/storage/MemoryTokensStorage.java
```diff
@@ -0,0 +1,56 @@
...
+
+package com.google.auth.oauth2.storage;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+public class MemoryTokensStorage implements TokenStore {
+
+    private final Map<String, String> tokensStorage = new HashMap<>();
```
garrettjonesgoogle: Indentation should be 2 spaces

oauth2_http/java/com/google/auth/oauth2/storage/TokenStore.java
```diff
@@ -29,7 +29,7 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-package com.google.auth.oauth2;
+package com.google.auth.oauth2.storage;
```
garrettjonesgoogle: Could we leave this in the original package? I'd prefer not to make a breaking change for this.

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -166,6 +171,18 @@ public static GoogleCredentials fromStream(InputStream credentialsStream,
...
 
   /**
+   * Saves the end user credentials in specified file
+   *
+   * @param credentials InputStream containing user credentials in JSON format
+   * @param credentialsFileName Name of file where to store the credentials
+   * @throws IOException An error saving the credentials.
+   */
+  public void saveCredentialsToFile(InputStream credentials, String credentialsFileName) throws IOException {
+    String userDirectory = System.getProperty("user.dir");
```
garrettjonesgoogle: What if the user wants to save the credentials at an absolute path, e.g. under "/tmp/..."?

oauth2_http/java/com/google/auth/oauth2/UserCredentials.java
```diff
@@ -244,6 +249,45 @@ public final String getRefreshToken() {
...
+    return new ByteArrayInputStream(text.getBytes(UTF_8));
+  }
+
+  /**
+   * Saves the end user credentials in file
+   *
+   * @throws IOException An error storing the credentials.
+   */
+  public void saveUserCredentials() throws IOException {
+    super.saveCredentialsToFile(getUserCredentialsStream(), USER_CREDENTIALS_FILE_NAME);
```
garrettjonesgoogle: I don't feel comfortable hardcoding the path where this is saved.

Also, the "UserCredentials" part of this name is redundant since it lives in the UserCredentials class.

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
YuryBandarchuk16: Would you like me to create several implementations or change this one for saving credentials file just at an absolute path?

garrettjonesgoogle: Change this to just not care - if the user gives relative path, then they get relative path; if they give an absolute path, they get an absolute path.

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -179,6 +184,22 @@ public static GoogleCredentials fromStream(InputStream credentialsStream,
   }
 
   /**
+   * Saves the end user credentials in specified file
+   *
+   * @param credentials InputStream containing user credentials in JSON format
+   * @param filePath Path to file where to store the credentials
+   * @throws IOException An error saving the credentials.
+   */
+  public void saveCredentialsToFile(InputStream credentials, String filePath) throws IOException {
```
garrettjonesgoogle: There is still nothing specific to credentials in this function. I would suggest moving it to `OAuth2Utils` so that it doesn't form part of the surface of the library (since `OAuth2Utils` is package-private). I suggest calling it "writeInputStreamToFile".

oauth2_http/java/com/google/auth/oauth2/UserCredentials.java
```diff
@@ -258,6 +261,47 @@ public final String getRefreshToken() {
...
+    if (clientSecret != null) {
+      json.put("client_secret", clientSecret);
+    }
+    json.setFactory(JSON_FACTORY);
+    String text = json.toPrettyString();
+    return new ByteArrayInputStream(text.getBytes(UTF_8));
+  }
+
+  /**
+   * Saves the end user credentials in file
```
garrettjonesgoogle: "in file" -> "into the given file path." (include the period)

oauth2_http/javatests/com/google/auth/oauth2/UserCredentialsTest.java
```diff
@@ -504,6 +506,39 @@ public void fromStream_userNoRefreshToken_throws() throws IOException {
...
   }
 
+  @Test
+  public void saveUserCredentials_saved_throws() throws IOException {
+    UserCredentials userCredentials = UserCredentials.newBuilder()
+            .setClientId(CLIENT_ID)
+            .setClientSecret(CLIENT_SECRET)
+            .setRefreshToken(REFRESH_TOKEN)
+            .build();
+    String filePath = System.getProperty("user.dir") + File.separator + "GOOGLE_APPLICATION_CREDENTIALS";
```
garrettjonesgoogle: As someone running these tests, I don't want this file to show up in my home directory when I run them. Put them under /tmp (use the Java function to auto-generated a file under /tmp)

oauth2_http/java/com/google/auth/oauth2/MemoryTokensStorage.java
```diff
@@ -0,0 +1,56 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package com.google.auth.oauth2;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+public class MemoryTokensStorage implements TokenStore {
```
garrettjonesgoogle: Add a class doc.

oauth2_http/javatests/com/google/auth/oauth2/UserCredentialsTest.java
YuryBandarchuk16: Updated to use temporary files.

oauth2_http/java/com/google/auth/oauth2/UserCredentials.java
YuryBandarchuk16: Fixed

YuryBandarchuk16:
@garrettjonesgoogle I think I made all the changes you asked for.

PR #132, opened 2017-09-18
Fix NPE deserializing ServiceAccountCredentials

igorbernstein2:
Deserialization of ServiceAccoutnCredentials broke in 0.7.0 via this commit:
https://github.com/google/google-auth-library-java/commit/432ee7e507e3dfa730a5001a2e561cd65e0ba721#diff-e186a7a41155398a3502025e16ac1508L479

It removed the java deserialization hook readObject to properly deserialize the transient transportFactory field.

This PR reverts that change and adds a test

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
@@ -60,6 +60,7 @@ import com.google.common.collect.ImmutableSet;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.ObjectInputStream;
 import java.io.Reader;
 import java.io.StringReader;
 import java.net.URI;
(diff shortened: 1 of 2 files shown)


igorbernstein2:
@garrettjonesgoogle @lesv @anthmgoogle: can someone review this? This bug breaks authentication via service accounts for dataflow io connectors. Please see b/65304954 for details

garrettjonesgoogle:
@vchudnov-g could you comment on why the `readObject` method was removed in your commit?

vchudnov-g:
As per my comments, the method seemed to be unused.

If this change bringing `readObject` back fixes things, I'm fine with it.

I'm still curious, though, where/how does `readObject` get invoked?

garrettjonesgoogle:
It's a magic invocation - see https://docs.oracle.com/javase/7/docs/api/java/io/Serializable.html .

vchudnov-g:
OK, thanks for the pointer.

The fix is fine by me. Maybe include a comment saying the method is needed for the Serializable interface?

I get that this is the Way Things Are Done, but these types of magic invocations seem really counterintuitive and lend themselves to issues like these.....

At any rate, my apologies for breaking things.

PR #136, opened 2017-11-02
Surface cleanup

garrettjonesgoogle:
- Removing Guava from surface
- Renaming of() methods to create() (using deprecation)
- Adding @deprecated javadoc to explain changes that callers need to make
- Migrating internal usage of deprecated methods that was missed previously

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
@@ -87,13 +87,19 @@ public class ComputeEngineCredentials extends GoogleCredentials {
    *
    * @param transportFactory The Http transport factory
    * @return the credential instance
+   * @deprecated Use {@link #newBuilder()} instead to construct an instance using the builder. This
+   *             method will be deleted in a later version.
    */
+  @Deprecated
   public static ComputeEngineCredentials of(HttpTransportFactory transportFactory) {
     return ComputeEngineCredentials.newBuilder().setHttpTransportFactory(transportFactory).build();
   }
 
   /**
    * Constructor with minimum information and default behavior.
+   *
+   * @deprecated Use {@link #create()} instead. This constructor will either be deleted or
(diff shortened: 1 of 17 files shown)


oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -114,6 +122,13 @@ public ComputeEngineCredentials(HttpTransportFactory transportFactory) {
   }
 
   /**
+   * Create a new ComputeEngineCredentials instance.
```
vam-google: Please copy here documentation from the deprecated `ComputeEngineCredentials()`

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -87,13 +87,19 @@
    *
    * @param transportFactory The Http transport factory
    * @return the credential instance
+   * @deprecated Use {@link #newBuilder()} instead to construct an instance using the builder. This
```
vam-google: Probably this description (also applies to other deprecated consturctors documentation) is unnecessarily verbose. "to construct an instance using the builder" - it is obvious, and does not need to be documented. "deleted or made private" - made private is same as deleted from point of view of client code.

lesv:
@jabubake FYI

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
garrettjonesgoogle: I can make it less verbose.

You are correct that it is the same from the client code point of view, but documentation is also for maintainers too.

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
garrettjonesgoogle: I will make them consistent.

PR #144, opened 2018-01-08
Modification to reading GOOGLE_APPLICATION_CREDENTIALS 

Tang8330:
Currently, an issue I have using Google Auth is that it requires `GOOGLE_APPLICATION_CREDENTIALS` as an environment variable, and point to the file with credentials.

While using this in Docker, it means that we would have to either:
1. Bake the credentials in the image
2. Mount the file on the image during run-time.

However, if we make a slight modification so that: `GOOGLE_APPLICATION_CREDENTIALS` can also be the JSON string of the file contents...we no longer need to do that.

Proposed changes:
1. Try reading GOOGLE_APPLICATION_CREDENTIALS as a file...
2. If (1) fails: before throwing an error...it will try reading GOOGLE_APPLICATION_CREDENTIALS as a JSON string.

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java
+++ b/oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java
@@ -40,8 +40,6 @@ import java.lang.reflect.Method;
 import java.security.AccessControlException;
 import java.util.Collections;
 import java.util.Locale;
-import java.util.logging.Level;
-import java.util.logging.Logger;
 
 /**
  * Provides the Application Default Credential from the environment.
@@ -124,8 +122,6 @@ class DefaultCredentialsProvider {
 
   private final GoogleCredentials getDefaultCredentialsUnsynchronized(
       HttpTransportFactory transportFactory) throws IOException {
-
-    final Logger LOGGER = Logger.getLogger(GoogleCredentials.class.getName());
 
     // First try the environment variable
     GoogleCredentials credentials = null;
--- a/oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java
+++ b/oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java
@@ -33,17 +33,15 @@ package com.google.auth.oauth2;
 
 import com.google.auth.http.HttpTransportFactory;
 
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.FileNotFoundException;
-import java.io.IOException;
-import java.io.InputStream;
+import java.io.*;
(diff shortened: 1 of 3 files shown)


Tang8330:
I have signed it! Please verify.

Tang8330:
@lesv Could you please take a look at this and see if there's a place for this improvement? I would love to have this change in.

oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java
```diff
@@ -141,12 +137,17 @@ private final GoogleCredentials getDefaultCredentialsUnsynchronized(
...
       } catch (IOException e) {
-        // Although it is also the cause, the message of the caught exception can have very
-        // important information for diagnosing errors, so include its message in the
-        // outer exception message also.
-        throw new IOException(String.format(
-            "Error reading credential file from environment variable %s, value '%s': %s",
-            CREDENTIAL_ENV_VAR, credentialsPath, e.getMessage()), e);
+        try {
+          // Before throwing an error, CREDENTIAL_ENV_VAR is set, let's try to see if it's the credentials
+          credentials = GoogleCredentials.fromStream(new ByteArrayInputStream(credentialsPath.getBytes("UTF-8")), transportFactory);
```
lesv: You are trying to encode the credentials themselves in the EnvVar and not the file.  This is caught as the result of the Throw("File does not exist.").  It might be better to actually test if it's likely first and do an alternate path, a bit more code, but likely to be better.

Look for an opening "{" and a closing "}" within a few characters at the end.  (my file has 0a 7d 0a)

lesv: It will be clearer what's going on that  way.

lesv:
@Tang8330 So, I'd like to discuss this change with a few people before accepting it, and I'll be asking for code changes, so before we go there, let me give you a few alternatives.

On GCP, GAE/Flex and GAE/Standard, you can ask the Metadata server for the default service account.  The Libraries already work with this, so it's not that big a deal to use.  For local machines, it's possible to point `metadata.google.internal` to `localhost` aka `127.0.0.1` and use a python app or a webserver to act as a metadata server.

lesv:
@jonparrott PTAL
@anthmgoogle Care to comment about this?

jonparrott:
I am unsupportive of this change, environment variables [should not be used to hold secrets](https://diogomonica.com/2017/03/27/why-you-shouldnt-use-env-variables-for-secret-data/).

Tang8330:
@jonparrott Yep, fair enough. We'll mount the file at run-time and in deployments.

PR #151, opened 2018-03-20
Add caching for JWT tokens

igorbernstein2:
Use guava's LoadingCache to cache JWT tokens. This should address performance for both clientside & serverside:
- Avoid the need for the client to recompute the signature
- Allows the serverside to cache tokens

Benchmarks (n1-standard-4):
- Generating a JWT token takes about 7~8ms (p50) on an idle machine
- Additionally, the newly generated tokens will add 9ms to p99 processing on serverside (for Cloud Bigtable)
- For reference, Cloud Bigtable has a 5ms response @ p99 using cached JWTs or OAuth tokens.
- Combined, the clientside & serverside overhead for using uncached JWT tokens triples Cloud Bigtable's p99 latency.

Merging this PR will have a significant impact on the consistency of Cloud Bigtable's performance (and other low latency products as well).

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/ServiceAccountJwtAccessCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/ServiceAccountJwtAccessCredentials.java
@@ -48,6 +48,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.MoreObjects;
 
 import com.google.common.base.Throwables;
+import com.google.common.base.Ticker;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.CacheLoader;
 import com.google.common.cache.LoadingCache;
@@ -80,7 +81,9 @@ public class ServiceAccountJwtAccessCredentials extends Credentials
 
   private static final long serialVersionUID = -7274955171379494197L;
   static final String JWT_ACCESS_PREFIX = OAuth2Utils.BEARER_PREFIX;
-  private static final int LIFE_SPAN_SECS = 3600;
+  @VisibleForTesting
+  static final long LIFE_SPAN_SECS = TimeUnit.HOURS.toSeconds(1);
+
 
   private final String clientId;
   private final String clientEmail;
@@ -243,6 +246,14 @@ public class ServiceAccountJwtAccessCredentials extends Credentials
     return CacheBuilder.newBuilder()
         .maximumSize(100)
         .expireAfterWrite(LIFE_SPAN_SECS - 10, TimeUnit.SECONDS)
+        .ticker(
+            new Ticker() {
+              @Override
+              public long read() {
+                return TimeUnit.MILLISECONDS.toNanos(clock.currentTimeMillis());
+              }
+            }
+        )
         .build(
             new CacheLoader<URI, String>() {
               @Override
--- a/oauth2_http/java/com/google/auth/oauth2/ServiceAccountJwtAccessCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/ServiceAccountJwtAccessCredentials.java
@@ -47,6 +47,11 @@ import com.google.auth.http.AuthHttpConstants;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.MoreObjects;
 
+import com.google.common.base.Throwables;
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
+import com.google.common.util.concurrent.UncheckedExecutionException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.ObjectInputStream;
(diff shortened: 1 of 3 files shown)


igorbernstein2:
@garrettjonesgoogle PTAL when you have moment

oauth2_http/java/com/google/auth/oauth2/ServiceAccountJwtAccessCredentials.java
```diff
@@ -275,10 +300,24 @@ public void getRequestMetadata(final URI uri, Executor executor,
...
+      // Should never happen
+      throw new IllegalStateException("generateJwtAccess threw an unexpected checked exception", e.getCause());
+
+    } catch (UncheckedExecutionException e) {
+      Throwables.propagateIfPossible(e);
+      // Should never happen
+      throw new IllegalStateException("generateJwtAccess threw an unchecked exception that couldn't be rethrown", e);
+    }
+  }
+  private String generateJwtAccess(URI uri) throws IOException {
```
garrettjonesgoogle: Put a blank line between the methods

garrettjonesgoogle:
@lesv could you take a look too?

igorbernstein2:
Updated:
- fixed pom versions to be the next snapshot
- add tests
- fixed whitespace

PTAL

garrettjonesgoogle:
Oh I was wanting to do the snapshot bump in a separate PR that gets committed before this one...

igorbernstein2:
Oops, sorry I misunderstood. I cherrypicked the version bump to a separate PR:
https://github.com/google/google-auth-library-java/pull/152

oauth2_http/java/com/google/auth/oauth2/ServiceAccountJwtAccessCredentials.java
```diff
@@ -228,6 +242,28 @@ public static ServiceAccountJwtAccessCredentials fromStream(InputStream credenti
             + " Expecting '%s'.", fileType, SERVICE_ACCOUNT_FILE_TYPE));
   }
 
+  private LoadingCache<URI, String> createCache() {
+    return CacheBuilder.newBuilder()
+        .maximumSize(100)
+        .expireAfterWrite(LIFE_SPAN_SECS - 10, TimeUnit.SECONDS)
```
lesv: 60 seconds would be a better choice

igorbernstein2: I bumped it to 5 minutes just in case

oauth2_http/java/com/google/auth/oauth2/ServiceAccountJwtAccessCredentials.java
```diff
@@ -119,6 +131,7 @@ public ServiceAccountJwtAccessCredentials(String clientId, String clientEmail,
     this.privateKey = Preconditions.checkNotNull(privateKey);
     this.privateKeyId = privateKeyId;
     this.defaultAudience = defaultAudience;
+    this.tokenCache = createCache();
```
lesv: Shouldn't this be in more than one place?  This is a deprecated method -- something appears wrong.

lesv: Ah - method is called - it's just weird, but appears to be Ok.

lesv: need to validate that fromStream() somehow set's tokenCache

oauth2_http/java/com/google/auth/oauth2/ServiceAccountJwtAccessCredentials.java
```diff
@@ -228,6 +241,28 @@ public static ServiceAccountJwtAccessCredentials fromStream(InputStream credenti
             + " Expecting '%s'.", fileType, SERVICE_ACCOUNT_FILE_TYPE));
   }
```
lesv: fromJson doesn't appear to create / set tokenCache.

oauth2_http/java/com/google/auth/oauth2/ServiceAccountJwtAccessCredentials.java
igorbernstein2: I'm not sure I understand what mean.  All initialization passes through this constructor. For example:
fromStream(InputStream) -> fromStream(InputStream, URI) -> fromJson(Map, URI) -> fromPkcs8(...) -> this constructor

I'm guessing `@Deprecated` annotation is intended for external callers to use the  builder instead. 

The only entry point that doesn't end up calling this constructor is when java deserialization. That ends up calling `readObject()`, which needs to recreate the cache., which is why I factored out createCache()

oauth2_http/java/com/google/auth/oauth2/ServiceAccountJwtAccessCredentials.java
igorbernstein2: it does via:
fromJson(Map, URI) -> fromPkcs8(...) -> new ServiceAccountJwtAccessCredentials

igorbernstein2:
@lesv PTAL

oauth2_http/java/com/google/auth/oauth2/ServiceAccountJwtAccessCredentials.java
garrettjonesgoogle: @igorbernstein2 is correct, this is just marked `@Deprecated` to point people to the builder.

lesv: SGTM

Issue #143, opened 2017-12-16
GoogleCredential.getapplicationDefault().getAccessToken() == NULL on GCP


lesv:
It works if you call `refreshAccessToken()`.  We should either:

1. change the documentation idiom to recommend using `refreshAccessToken()` instead
2. comment that in `getAccessToken()` that it might not work.
3. Make `getAccessToken()` call `refreshAccessToken()` when nothing is cached.

chingor13:
I suggest we do (1) and (2).

We should encourage people who are explicitly trying to access the `AccessToken` to do something like:

```java
GoogleCredentials credentials = GoogleCredentials.getApplicationDefault();
AccessToken token = credentials.refreshAccessToken();
// OR
credentials.refresh();
AccessToken token = credentials.getAccessToken();
```

For (3), there are a few issues:
* `OAuth2Credentials#getAccessToken()` would need to either add `throws IOException` or try to refresh and then silently swallow any exceptions.
* For any swallowed exception, you will miss out on the contents of the exception and not realize that we are potentially making extra auth requests
* Anyone who is currently using `getAccessToken()` does not expect to be making IO calls. This is a behavior change and will break downstream users.

chingor13:
Adding a `refreshIfExpired()` that smartly refreshes a possibly expired token would make the way to do this:

```java
GoogleCredentials credentials = GoogleCredentials.getApplicationDefault();
credentials.refreshIfExpired();
AccessToken token = credentials.getAccessToken();
```

This should be easier to reason about then calling `credentials.getRequestMetadata(uri)` which is what the client libraries call before each request (which will load or refresh the credentials if necessary).

PR #162, opened 2018-06-22
Add documentation note that getAccessToken() returns cached value

chingor13:
See #143

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
@@ -37,7 +37,6 @@ import com.google.auth.RequestMetadataCallback;
 import com.google.auth.http.AuthHttpConstants;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.MoreObjects;
-import com.google.common.base.MoreObjects.ToStringHelper;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.Iterables;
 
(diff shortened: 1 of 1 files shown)


oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
```diff
@@ -131,6 +130,11 @@ public boolean hasRequestMetadataOnly() {
     return true;
   }
 
+  /**
+   * Returns the cached access token.
+   *
+   * <p>If not set, you should call {@link #refresh()} to fetch and cache an access token.</p>
+   */
   public final AccessToken getAccessToken() {
```
JustinBeckwith: Wouldn't be better to automatically try obtaining the access token if it hasn't been refreshed yet?

chingor13: A couple of issues mentioned in #143:

* We can't change the method signature to throw an `IOException` without breaking downstream dependencies (some we control, others we may not).
* We could catch and swallow the exceptions which could cause extra, failing requests that are invisible to the user and cannot be acted upon.

chingor13: Technically, this library is pre-1.0 so we could make `getAccessToken()` throw `IOException`.

chingor13: `getAccessToken()` by design is meant to be an accessor to a cached value. It doesn't make sense to try to make it refresh the token and changing the behavior will be unexpected for most users.

We'll document it more clearly that this is only an accessor and introduce a smart refresh request in #163

PR #166, opened 2018-06-28
Log warning if default credentials yields a user token from gcloud sdk

### LINKED ISSUE #165: Soft-warn about application default credentials using gcloud credentials
Due to user issues with quota and API enablement, the auth libraries should issue a soft warning (that can be silenced) when application default credentials uses user credentials from the Cloud SDK. They should not warn if the credentials for the cloud sdk are service account credentials.

```
Your application has authenticated using end user credentials from Google 
Cloud SDK. We recommend that most server applications use service accounts
 instead. If your application continues to use end user credentials from Cloud 
SDK, you might receive a "quota exceeded" or "API not enabled" error. For 
more information about service accounts, see 
https://cloud.google.com/docs/authentication/.
```

Additional context:
https://groups.google.com/a/google.com/forum/#!topic/client-auth-team/DKqrFw6lL1Q
https://buganizer.corp.google.com/issues/64388723
https://github.com/GoogleCloudPlatform/google-auth-library-python/pull/266

chingor13:
Fixes #165

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java
+++ b/oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java
@@ -211,8 +211,7 @@ class DefaultCredentialsProvider {
   }
 
   private void warnAboutProblematicCredentials(GoogleCredentials credentials) {
-    if (credentials != null &&
-        (credentials instanceof UserCredentials) &&
+    if (credentials instanceof UserCredentials &&
         ((UserCredentials)credentials).getClientId().equals(CLOUDSDK_CLIENT_ID)) {
       LOGGER.log(Level.WARNING, CLOUDSDK_CREDENTIALS_WARNING);
     }
--- a/oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java
+++ b/oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java
@@ -44,6 +44,8 @@ import java.lang.reflect.Method;
 import java.security.AccessControlException;
 import java.util.Collections;
 import java.util.Locale;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 /**
  * Provides the Application Default Credential from the environment.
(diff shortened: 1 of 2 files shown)


oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java
```diff
@@ -197,6 +210,14 @@ private final GoogleCredentials getDefaultCredentialsUnsynchronized(
     return credentials;
   }
 
+  private void warnAboutProblematicCredentials(GoogleCredentials credentials) {
+    if (credentials != null &&
```
jadekler: credentials != null can be removed since the next line covers it

jadekler: I believe credentials != null can be removed since the next line covers it.

oauth2_http/javatests/com/google/auth/oauth2/DefaultCredentialsProviderTest.java
```diff
@@ -437,6 +445,46 @@ public void getDefaultCredentials_envAndWellKnownFile_envPrecedence() throws IOE
...
+        new File(cloudConfigDir, DefaultCredentialsProvider.WELL_KNOWN_CREDENTIALS_FILE);
+    TestDefaultCredentialsProvider testProvider = new TestDefaultCredentialsProvider();
+    testProvider.setProperty("os.name", "linux");
+    testProvider.setProperty("user.home", homeDir.getAbsolutePath());
+    testProvider.addFile(wellKnownFile.getAbsolutePath(), userStream);
+
+    testUserProvidesToken(
+        testProvider, GCLOUDSDK_CLIENT_ID, USER_CLIENT_SECRET, REFRESH_TOKEN);
+    LogRecord message = handler.getRecord();
+    assertEquals(Level.WARNING, message.getLevel());
```
jadekler: If the log never happens, `message` ends up being null and the test errors with a null pointer exception trying to getLevel().

It might be nicer if the test failed. Could you add `assertNotNull(message);`?

oauth2_http/javatests/com/google/auth/oauth2/DefaultCredentialsProviderTest.java
```diff
@@ -71,6 +75,10 @@
 
   private static final String USER_CLIENT_SECRET = "jakuaL9YyieakhECKL2SwZcu";
   private static final String USER_CLIENT_ID = "ya29.1.AADtN_UtlxN3PuGAxrN2XQnZTVRvDyVWnYq4I6dws";
+  private static final String GCLOUDSDK_CLIENT_ID =
+      "764086051850-6qr4p6gpi6hn506pt8ejuq83di341hur.apps.googleusercontent.com";
+  private static final String GCLOUD_WARNING_CONTENTS =
```
jadekler: Since this is only used once, and is not really generic, I'm in favour of not having a variable for it. Thoughts?

oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java
chingor13: Done. Removed the null check as the instanceof check also handles the null case.

oauth2_http/javatests/com/google/auth/oauth2/DefaultCredentialsProviderTest.java
chingor13: Done.

oauth2_http/javatests/com/google/auth/oauth2/DefaultCredentialsProviderTest.java
chingor13: SGTM, moved to an inline string.

PR #174, opened 2018-08-07
Update new token urls

### LINKED ISSUE #173: Use new urls for authentication
[Parent issue](https://github.com/GoogleCloudPlatform/google-cloud-common/issues/260)

The following specific library changes are required:

### OAuth 2.0 Authorization Grant Flow (aka 3LO)

For clients using this flow, change (and test):
- https://accounts.google.com/o/oauth2/token -> https://oauth2.googleapis.com/token
- https://accounts.google.com/o/oauth2/revoke -> https://oauth2.googleapis.com/revoke
- https://accounts.google.com/o/oauth2/tokeninfo -> https://oauth2.googleapis.com/tokeninfo
- *DO NOT* change https://accounts.google.com/o/oauth2/auth

If the client pulls the data from the client json object, that will be updated on the server side for newly created clients and no changes to the client libraries are required.

### OAuth 2.0 Assertion Flow (aka OAuth 2.0 Service Account Flow, aka “Google Credentials Auth”)

For clients using these flows, change:
- https://accounts.google.com/o/oauth2/token -> https://oauth2.googleapis.com/token
- The “audience” of the token from https://accounts.google.com/o/oauth2/token ->  https://oauth2.googleapis.com/token

### JWT Bearer

No changes should be required for JWT Bearer authentication

### OIDC Identity Tokens

There are two options.  If possible, the preferred option in this case is to:
- Change https://accounts.google.com/o/oauth2/token ->  https://oauth2.googleapis.com/token
- Change audience verification from “accounts.google.com” to [“accounts.google.com”, “https://accounts.google.com”]
- *DO NOT* Change https://www.googleapis.com/oauth2/v1/certs

If this is too risky in a particular client library, then the change should be:
- https://accounts.google.com/o/oauth2/token -> https://www.googleapis.com/oauth2/v3/token
- *DO NOT* Change https://www.googleapis.com/oauth2/v1/certs

chingor13:
Fixes #173

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/OAuth2Utils.java
+++ b/oauth2_http/java/com/google/auth/oauth2/OAuth2Utils.java
@@ -61,8 +61,8 @@ import java.util.Map;
 class OAuth2Utils {
   static final String SIGNATURE_ALGORITHM = "SHA256withRSA";
 
-  static final URI TOKEN_SERVER_URI = URI.create("https://accounts.google.com/o/oauth2/token");
-  static final URI TOKEN_REVOKE_URI = URI.create("https://accounts.google.com/o/oauth2/revoke");
+  static final URI TOKEN_SERVER_URI = URI.create("https://oauth2.googleapis.com/token");
+  static final URI TOKEN_REVOKE_URI = URI.create("https://oauth2.googleapis.com/revoke");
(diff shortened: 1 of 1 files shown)


JustinBeckwith:
Thank you :) As soon as this lands (here and in PHP), can we release new packages?

PR #188, opened 2018-09-06
add App Engine API to pom.xml with test scope

### LINKED ISSUE #187: Document how to run tests in CONTRIBUTING.MD
In particular it seems the App Engine API SDK needs to be installed somewhere in particular, and is not referenced by the pom.xml. I haven't yet figured out exactly where.

elharo:
I'm not totally clear on why this dependency isn't already in the pom.xml (maybe with scope provided?) but in any case this will make the tests pass and moots #187 

fix #187

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -95,6 +95,12 @@
         <scope>test</scope>
       </dependency>
       <dependency>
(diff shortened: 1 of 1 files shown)


elharo:
seems to be working now

PR #190, opened 2018-09-06
Remove deprecated methods

elharo:
Immediately following the last release is the best time to rip out all the deprecated methods.

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
@@ -59,19 +59,6 @@ public class GoogleCredentials extends OAuth2Credentials {
    *
    * @param accessToken the access token
    * @return the credentials instance
-   * @deprecated Use {@link #create(AccessToken)} instead. This method will be deleted in a later
-   *             version.
-   */
-  @Deprecated
-  public static GoogleCredentials of(AccessToken accessToken) {
-    return create(accessToken);
-  }
-
-  /**
-   * Returns the credentials instance from the given access token.
-   *
-   * @param accessToken the access token
-   * @return the credentials instance
    */
   public static GoogleCredentials create(AccessToken accessToken) {
     return GoogleCredentials.newBuilder().setAccessToken(accessToken).build();
@@ -193,7 +180,6 @@ public class GoogleCredentials extends OAuth2Credentials {
    * @deprecated Use {@link #create(AccessToken)} instead. This constructor will either be deleted
    *             or made protected/private in a later version.
    **/
-  @Deprecated
   public GoogleCredentials(AccessToken accessToken) {
     super(accessToken);
   }
--- a/oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
@@ -176,10 +176,8 @@ public class GoogleCredentials extends OAuth2Credentials {
   /**
    * Constructor with explicit access token.
    *
-   * @param accessToken Initial or temporary access token.
-   * @deprecated Use {@link #create(AccessToken)} instead. This constructor will either be deleted
-   *             or made protected/private in a later version.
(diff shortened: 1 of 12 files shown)


garrettjonesgoogle:
We need to be very careful with removals in this library, and make sure the major libraries in the ecosystem have advanced to the version containing the new methods and have removed references to the deprecated ones. (We need to check grpc, google-cloud-java, and Beam). If we don't do this, this particular library could be a huge source of pain for diamond dependency conflicts. (It is very deep in the dependency tree.) I was almost thinking of recommend that these deprecated methods not even get removed when it is pushed to 1.0.

elharo:
This pre-1.0 library has had one release in the last year+, and only really works on App Engine standard. Now is the time to cut the dead wood, so dependents can upgrade at a more leisurely pace. For now, they can stick with 0.11 and have one PR that upgrades to 0.12 while fixing any deprecated methods. In a non-monorepo world we can't require all usages of methods to be changed at the same time we push a new release. The sooner we make the change, the easier it will be for everyone.

elharo:
To upgrade those libraries one can either manually inspect all possible usages, or simply add a new version with the methods removed and see what breaks. 

Removing these post-1.0 would be an API breaking change that requires a new major version so we definitely don't want to wait till then to do this.

And I'd like to get 1.0 out the door sooner rather than later. The libraries you cite are all GA or include GA artifacts, so depending on this is a problem. 

Looks like I misread: only google-auth-library-appengine depends on the App Engine SDK.

garrettjonesgoogle:
You don't have to publish a new version to maven central to see what breaks; you can `mvn install` your version with deprecated code removed, then compile & run tests in the major consumers locally and see if they break.

I completely agree that the dependency of GA libraries on unstable libraries is a problem - we both worked on https://github.com/GoogleCloudPlatform/cloud-opensource-java/blob/master/library-best-practices/JLBP-4.md , after all :-)

chingor13:
We should check usage for these against artifacts that use [`google-auth-libarary-oauth2-http`](https://mvnrepository.com/artifact/com.google.auth/google-auth-library-oauth2-http/usages) and [`google-auth-library-appengine`](https://mvnrepository.com/artifact/com.google.auth/google-auth-library-appengine/usages)

sduskis:
@elharo, what's the status of this PR?

chingor13:
We still need to verify that we aren't using any of these methods downstream (at least across Google controlled libraries)

elharo:
I haven't checked on this lately, so there may be new conflicts with this specific PR, but my opinion is still the same as when I sent it: remove all the deprecated methods now, before 1.0, and without worrying about who might be calling them. No project should be immediately broken by this since dependency upgrades are not automatic. If a project does attempt an upgrade and is broken, this will be a compile time failure and easily detected.

garrettjonesgoogle:
Disagree. This library is "common-law GA".

elharo:
Just to be clear, I don't have an informed technical opinion on the specific methods here. I do believe we should go to 1.0, and we should remove or undeprecate these methods before we do.

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -193,7 +180,6 @@ protected GoogleCredentials() {
    * @deprecated Use {@link #create(AccessToken)} instead. This constructor will either be deleted
    *             or made protected/private in a later version.
    **/
-  @Deprecated
```
sduskis: I see `@Deprecated` was removed here.  If it's actively used, please also remove the `@deprecated` comment`

elharo: done

sduskis:
Who is the owner of the decision of whether or not this ought to be merged?

sduskis:
Given lack of clear ownership, I'll take responsibility.  There will potentially be subtle diamond dependency problems based on this change.  The recommendation is to raise an issue in this repository, but also in the repository that's using old deprecated APIs.

PR #206, opened 2018-11-08
Show error message in case of problems with getting access token

dtretyakov:
Currently in case of problems with token refresh is unclear what caused the problem and this change brings additional details.

Related to https://github.com/JetBrains/teamcity-google-storage/pull/9


dtretyakov:
@chingor13, could you please take a look on it? It'll help to diagnose credentials-related problems.

PR #207, opened 2018-11-09
Option to suppress end user credentials warning.

### LINKED ISSUE #193: How to suppress warning about Google Cloud SDK credentials?
I'm wondering how to suppress the warning:

```
WARNING: Your application has authenticated using end user credentials from Google Cloud SDK. We recommend that most server applications use service accounts instead. If your application continues to use end user credentials from Cloud SDK, you might receive a "quota exceeded" or "API not enabled" error. For more information about service accounts, see https://cloud.google.com/docs/authentication/.
```

Our application is a desktop application that uses the users default credentials to access files on the cloud and this warning is confusing to users.  Is there a way to suppress it?  I thought I'd found one but I accidentally asked in the php version of the auth library and not the java one.  

Would it be possible to port the patch that they implemented to suppress the warning when the environment variable `SUPPRESS_GCLOUD_CREDS_WARNING` is set to `true`?

https://github.com/googleapis/google-auth-library-php/pull/199#issuecomment-411490419

lbergelson:
Defining a new environment variable SUPPRESS_GCLOUD_CREDS_WARNING.
Setting this to true suppresses the end user credentials warning.

Fixes #193

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java
+++ b/oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java
@@ -212,9 +212,9 @@ class DefaultCredentialsProvider {
   }
 
   private void warnAboutProblematicCredentials(GoogleCredentials credentials) {
-    if (!Boolean.parseBoolean(getEnv(SUPPRESS_GCLOUD_CREDS_WARNING_ENV_VAR))
-        && credentials instanceof UserCredentials
-        && ((UserCredentials) credentials).getClientId().equals(CLOUDSDK_CLIENT_ID)) {
+    if (credentials instanceof UserCredentials
+        && ((UserCredentials) credentials).getClientId().equals(CLOUDSDK_CLIENT_ID)
+        && !Boolean.parseBoolean(getEnv(SUPPRESS_GCLOUD_CREDS_WARNING_ENV_VAR))) {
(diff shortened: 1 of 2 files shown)


oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java
```diff
@@ -211,8 +212,9 @@ private final GoogleCredentials getDefaultCredentialsUnsynchronized(
   }
 
   private void warnAboutProblematicCredentials(GoogleCredentials credentials) {
-    if (credentials instanceof UserCredentials &&
-        ((UserCredentials)credentials).getClientId().equals(CLOUDSDK_CLIENT_ID)) {
+    if (!Boolean.parseBoolean(getEnv(SUPPRESS_GCLOUD_CREDS_WARNING_ENV_VAR))
```
chingor13: Should we make this the last check (i.e. only check the environment variable if we would log the warning?).

Otherwise, this check will be performed more times than necessary (whenever we attempt to load credentials from the well known location).

lbergelson: @chingor13 I've reordered the checks. Thanks for the review!

lbergelson:
@chingor13 Thank you.

PR #218, opened 2018-12-19
Overload GoogleCredentials.createScoped with variadic arguments

### LINKED ISSUE #5: Overload GoogleCredentials.createScoped to accept repeated strings
Would improve usability to support

GoogleCredentials.createScope(String... scopes)

chingor13:
Fixes #5

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
@@ -37,6 +37,7 @@ import com.google.api.client.json.JsonObjectParser;
 import com.google.api.client.util.Preconditions;
 import com.google.auth.http.HttpTransportFactory;
 
+import com.google.common.collect.ImmutableList;
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -227,6 +227,17 @@ public GoogleCredentials createScoped(Collection<String> scopes) {
...
 
+  /**
+   * If the credentials support scopes, creates a copy of the the identity with the specified
+   * scopes; otherwise, returns the same instance.
+   *
+   * @param scopes Collection of scopes to request.
+   * @return GoogleCredentials with requested scopes.
+   */
+  public GoogleCredentials createScoped(String... scopes) {
+    return createScoped(scopes);
```
danielnorberg: Looks like this will just call itself, eventually resulting in a stack overflow.

Perhaps this should be:

```java
return createScoped(ImmutableList.copyOf(scopes));
```

PR #222, opened 2019-01-16
Enable autorelease

busunkim96:
* allow Autorelease parameters in the Kokoro job 
* enable the releasetool reporter in the release script

See [Setting up Autorelease](https://docs.google.com/document/d/1JfctabcBMeTdHbPQKhalaWiRHgSVGC6zxOPc1JwJGzc/edit#).

### DIFF
--- a/.kokoro/release/stage.sh
+++ b/.kokoro/release/stage.sh
@@ -15,6 +15,10 @@
 
 set -eo pipefail
 
+# Start the releasetool reporter
+python3 -m pip install gcp-releasetool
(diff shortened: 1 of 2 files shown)


.kokoro/release/stage.sh
```diff
@@ -15,6 +15,10 @@
 
 set -eo pipefail
 
+# Start the releasetool reporter
+python3 -m pip install gcp-releasetool
```
chingor13: We may need to ensure that our Java 8 image contains python 3 and pip.

chingor13: Our image does not contain python3. We'll have to add it.

PR #229, opened 2019-03-05
createScoped: fix endless recursion

danielnorberg:
Call `createScoped` implementation instead of endlessly recursing.


sduskis:
yikes.  Thanks for finding this.

PR #230, opened 2019-03-08
Fix#228 update default metadata url

### LINKED ISSUE #228: DefaultCredentialsProvider is broken for java8 app engine
This was reported by a bigtable customer:

> I've traced through where it fails. In DefaultCredentialsProvider.getDefaultCredentialsUnsynchronized() it will only return the AppEngine credentials if the following evaluates to true:
> 
>     this.isOnGAEStandard7()
> 
> I am on GAE Standard but am using Java 8 not 7, so this check is false and it does not return any valid credentials.

ajaaym:
Fixes #228

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
@@ -71,7 +71,7 @@ public class ComputeEngineCredentials extends GoogleCredentials implements Servi
   private static final Logger LOGGER = Logger.getLogger(ComputeEngineCredentials.class.getName());
 
   // Note: the explicit IP address is used to avoid name server resolution issues.
-  static final String DEFAULT_METADATA_SERVER_URL = "http://169.254.169.254";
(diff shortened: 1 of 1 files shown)


oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -71,7 +71,7 @@
   private static final Logger LOGGER = Logger.getLogger(ComputeEngineCredentials.class.getName());
 
   // Note: the explicit IP address is used to avoid name server resolution issues.
-  static final String DEFAULT_METADATA_SERVER_URL = "http://169.254.169.254";
+  static final String DEFAULT_METADATA_SERVER_URL = "http://metadata.google.internal";
```
sduskis: Please use `metadata.google.internal.` with the `.` at the end.  We need that to solve a nasty problem in docker related to that problem.

ajaaym: We had this issue [#4269](https://github.com/googleapis/google-cloud-java/issues/4269) and its fixed with http://metadata.google.internal for GKE and GCE both.

PR #234, opened 2019-03-25
Update dependencies versions

elharo:
This moves Guava and AppEngine API SDK forward. 
It resolves a minor inconsistency in the version definition for the AppEngine API SDK.
It also moves the google-http-java-client *backwards* to 1.27.0 because 1.28.0 made an incompatible change that breaks google-cloud-java (which depends on 1.27.0) and using 1.28.0 here risks runtime problems down the line.


elharo:
It's possible the "Kokoro - Test: Java 8" job is wedged somehow or not reporting status. Both this PR and the much simpler #233 have been stuck at "Expected — Waiting for status to be reported" for a while now.

PR #289, opened 2019-07-08
Adds CI test for mvn dependency:analyze and fixes missing deps

chingor13:
Ensures that we do not declare any extra dependencies and are declaring all direct dependencies.


chingor13:
cc @elharo

PR #292, opened 2019-07-11
Refactor shared IAM-based signing

chingor13:
Moved shared logic into a package private `IamUtils` class


chingor13:
cc @salrashid123 

Additional shared IAM API behavior would belong in the `IamUtils` class

salrashid123:
ok, i'll see if i can use that ofr the `IDToken` implementation.

at the moment i'm having some issues making the refactor working for some reason...i'll update this when i figure that out...

however, for one thing, the delegates aren't applied in `IAMUtils`:

https://github.com/googleapis/google-auth-library-java/blob/master/oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java#L202

chingor13:
Ah, I see, we also need to add the metadata from the source credentials as well. It seems we're missing some tests here. I'll add a few more and fix the implmentation.

salrashid123:
ok, i see whats wrong here:

the `requestHeaders`  in `ImpersonatedCredentials.java` [here](https://github.com/googleapis/google-auth-library-java/blob/30f61a23513bfed6b273411cbe516da753366a3e/oauth2_http/java/com/google/auth/oauth2/IamUtils.java#L77)  represents the _impersonatedAccount_ already...and that can't invoke the `IAMCredetnials.signBlob()` endpoint
```java
  static byte[] sign(String serviceAccountEmail, Map<String, List<String>> requestHeaders,
                            HttpRequestFactory requestFactory, byte[] toSign) {
    BaseEncoding base64 = BaseEncoding.base64();
    String signature;
    try {
      signature = getSignature(serviceAccountEmail, requestHeaders, requestFactory,
              base64.encode(toSign));
    } catch (IOException ex) {
      throw new ServiceAccountSigner.SigningException("Failed to sign the provided bytes", ex);
    }
    return base64.decode(signature);
  }
```


So , if serviceAccountA is the sourceCredential and can impersonate serviceAccountB, the `requestHeaders` there includes an `access_token` for serviceAccountB...and you can't use  that toinvoke `iamcredentials.signBlob()`.

What'll we'll need to do is still use the `sourceCredentials` as the identity that invokes `.signBlob()` like this

https://github.com/googleapis/google-auth-library-java/blob/master/oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java#L206

---

i verfied this is the case by taking the code in this PR as is and first running it...i got 
`Failed to sign the provided bytes: Error code 403 trying to sign provided bytes: The caller does not have permission`  (and in decoding th eaccess token, i saw its for serviceAccountB).

once i set the permissions on serviceAccoutnB to impersonate "itself" (which isn't what we want to do here), the code here worked.   In summary, i think we'll want to go back to using the credentials as in the original PR submitted for this

salrashid123:
i verified your latest commit using both `ServiceAccountCredentials` and `ComputeCredentials` as source (both worked: signed and worked w/ target google-cloud-* library) 

```java
ServiceAccountCredentials sourceCredentials = ServiceAccountCredentials
          .fromStream(new FileInputStream("/home/srashid/gcp_misc/certs/mineral-minutia-820-83b3ce7dcddb.json"));

sourceCredentials = (ServiceAccountCredentials) sourceCredentials
 .createScoped(Arrays.asList("https://www.googleapis.com/auth/iam"));


 //or
// ComputeEngineCredentials sourceCredentials = ComputeEngineCredentials.create();


ImpersonatedCredentials targetCredentials = ImpersonatedCredentials.create(sourceCredentials,
          "impersonated-account@fabled-ray-104117.iam.gserviceaccount.com", null,
          Arrays.asList("https://www.googleapis.com/auth/devstorage.read_only", "https://www.googleapis.com/auth/cloud-platform"), 300);

Storage storage_service = StorageOptions.newBuilder()
      .setCredentials(targetCredentials).setProjectId("fabled-ray-104117")
      .build()
      .getService();

String BUCKET_NAME1= "fabled-ray-104117";
String BLOB_NAME1 = "signed_url_file.txt";
BlobInfo BLOB_INFO1 = BlobInfo.newBuilder(BUCKET_NAME1, BLOB_NAME1).build();

URL url =
      storage_service.signUrl(
        BLOB_INFO1,
        140,
        TimeUnit.MINUTES,
           Storage.SignUrlOption.httpMethod(HttpMethod.GET),
           Storage.SignUrlOption.withV4Signature());
System.out.println(url);

Page<Bucket> buckets =  storage_service.list(BucketListOption.pageSize(100));
for (Bucket bucket : buckets.iterateAll())
      System.out.println(bucket);
```

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -292,55 +288,8 @@ public String getAccount() {
...
-
-  private String getSignature(String bytes) throws IOException {
-    String signBlobUrl = String.format(SIGN_BLOB_URL_FORMAT, getAccount());
-    GenericUrl genericUrl = new GenericUrl(signBlobUrl);
-
-    GenericData signRequest = new GenericData();
-    signRequest.set("payload", bytes);
-    JsonHttpContent signContent = new JsonHttpContent(OAuth2Utils.JSON_FACTORY, signRequest);
-    HttpRequest request = transportFactory.create().createRequestFactory().buildPostRequest(genericUrl, signContent);
-    Map<String, List<String>> headers = getRequestMetadata();
```
kolea2: Highlighting this change in the `getSignature()` code

chingor13: This was converted to use the `HttpCredentialsAdapter` as an `HttpRequestInitializer` to set the headers rather than doing it manually.

PR #301, opened 2019-07-16
Update copyright format

elharo:
@chingor13

### DIFF
--- a/java.header
+++ b/java.header
@@ -1,5 +1,5 @@
 ^/\*$
-^ \* Copyright \d\d\d\d, Google LLC$
+^ \* Copyright \d\d\d\d, Google (Inc\. All rights reserved\.|LLC)$
 ^ \*$
 ^ \* Redistribution and use in source and binary forms, with or without$
 ^ \* modification, are permitted provided that the following conditions are$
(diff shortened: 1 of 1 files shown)


chingor13:
Note, this will cause the tests to fail until we convert all the headers

elharo:
In that case what do you think about disabling the check since it isn't smart enough to distinguish between new and modified code?

java.header
```diff
@@ -1,5 +1,5 @@
 ^/\*$
-^ \* Copyright \d\d\d\d, Google Inc\. All rights reserved\.$
+^ \* Copyright \d\d\d\d, Google LLC$
```
chingor13: This looks like a regex - we can try:
```suggestion
^ \* Copyright \d\d\d\d, Google (Inc\. All rights reserved\.|LLC)$
```

java.header
```diff
@@ -12,7 +12,7 @@
 ^ \* in the documentation and/or other materials provided with the$
 ^ \* distribution\.$
 ^ \*$
-^ \*    \* Neither the name of Google Inc\. nor the names of its$
+^ \*    \* Neither the name of Google LLC nor the names of its$
```
chingor13:
```suggestion
^ \*    \* Neither the name of Google (Inc\.|LLC) nor the names of its$
```

chingor13:
@elharo The java.header uses regex, so the suggested fixes will work and keep backwards compatibility.

PR #312, opened 2019-07-29
chore: enable lint presubmit and run code formatter

chingor13:
To run the formatter, you can use:
```bash
mvn com.coveo:fmt-maven-plugin:format
```

### DIFF
--- a/.kokoro/build.sh
+++ b/.kokoro/build.sh
@@ -22,6 +22,21 @@ java -version
 echo $JOB_TYPE
 
 mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
-mvn test -B
 
-bash $KOKORO_GFILE_DIR/codecov.sh
+case ${JOB_TYPE} in
+test)
+    mvn test -B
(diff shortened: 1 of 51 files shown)


.kokoro/build.sh
```diff
@@ -22,6 +22,21 @@ java -version
...
-mvn test -B
 
-bash $KOKORO_GFILE_DIR/codecov.sh
+case ${JOB_TYPE} in
+test)
+    mvn test -B
+    bash ${KOKORO_GFILE_DIR}/codecov.sh
+    ;;
+lint)
+    mvn com.coveo:fmt-maven-plugin:check
```
elharo: I've always found Github+Maven formatting checks to be more trouble than they're worth. They block PRs in a way the checks in google3 don't, and they don't integrate well into the code review like they do in critique.

chingor13: I also would rather not have to review the Java styling on PRs to instead focus on the meat of the PR.

I can change the failed message to be instructions about fixing the lint.

Issue #314, opened 2019-07-30
ComputeEngineCredentials incorreclty throws RuntimeException


elharo:
Unless we really think getDefaultServiceAccount can't fail hnere, we should not convert the IOException to a RuntimeException. 

```
  @Override
  public String getAccount() {
    if (serviceAccountEmail == null) {
      try {
        serviceAccountEmail = getDefaultServiceAccount();
      } catch (IOException ex) {
        throw new RuntimeException("Failed to to get service account", ex);
      }
    }
    return serviceAccountEmail;
}
```

elharo:
This method is called by sign, which throws SigningException but does not catch this RuntimeException! SigningException is also a runtime exception and shouldn't be.

elharo:
Since this library is pre-1.0 we're allowed to change the API. We should make SigningException (and possibly others) checked since they came fail unexpectedly due to problems external to the program itself, and therefore client code should write handlers.

elharo:
AppEngineCredential creates its service account when the object is created and does properly throw IOException if this fails. That's probably what we should do here.

elharo:
serviceAccountEmail is transient:

```
  private transient String serviceAccountEmail;
```

so if it's serialized it will be looked up again when needed. Can it have a different value then?

chingor13:
If we make SigningException a checked exception, we will require anyone who is using `credentials.sign()` to add exception handling.

Amongst Google libraries, I only see one use (albeit an important one) https://github.com/googleapis/java-storage/blob/6f172eba6fd6e9c11a1f49569249ea6e714ea91f/google-cloud-storage/src/main/java/com/google/cloud/storage/StorageImpl.java#L781

elharo:
That is a good thing and is why we should make this a checked exception.

Everyone now has to handle an unexpected runtime exception. In most cases they're not so they're shipping buggy, unreliable code that fails unexpectedly. 

The alternative is to figure out to make getDefaultServiceAccount or getAccount not fail. That is, what it can do that is correct behavior instead of throwing a runtime exception? But if there is no such behavior it needs to be a checked exception.

JustinBeckwith:
Downgrading this to a P2 since it's not blocking any critical functionality at this stage.  If that should change, let me know :)

elharo:
This library is pre-1.0. We should be able to make this change.

silvolu:
[...]

PR #316, opened 2019-07-31
throw SigningException as documented

elharo:
Least invasive change that partially addresses #314 but  we need to change the API to correctly fix this. @chingor13

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
@@ -277,7 +277,13 @@ public class ComputeEngineCredentials extends GoogleCredentials implements Servi
     return new Builder();
   }
 
+  /**
+   * Returns the email address associated with the GCE default service account.  
+   * 
+   * @throws RuntimeException if the default service account cannot be read
+   */
   @Override
+  // todo(#314) getAccount should not throw a RuntimeException
   public String getAccount() {
     if (serviceAccountEmail == null) {
       try {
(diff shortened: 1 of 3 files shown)


oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -304,12 +304,15 @@ public String getAccount() {
...
-        transportFactory.create(),
-        toSign,
-        Collections.<String, Object>emptyMap());
+    try {
+      String account = getAccount();
+      return IamUtils.sign(
+          account, this, transportFactory.create(), toSign, Collections.<String, Object>emptyMap());
+    } catch (SigningException ex) {
+      throw ex;
+    } catch (RuntimeException ex) {
```
chingor13: Let's add a TODO to clean this up that mentions #314

elharo: done. Also added some JavaDoc to getAccount.

oauth2_http/javatests/com/google/auth/oauth2/ComputeEngineCredentialsTest.java
```diff
@@ -296,7 +295,27 @@ public void sign_sameAs() throws IOException {
     ComputeEngineCredentials credentials =
         ComputeEngineCredentials.newBuilder().setHttpTransportFactory(transportFactory).build();
 
-    assertArrayEquals(expectedSignature, credentials.sign(expectedSignature));
+    assertEquals(defaultAccountEmail, credentials.getAccount());
```
chingor13: I think we want to still test the signature in this test. Perhaps keep both assertions or split the `getAccount()` test into its own test.

elharo: editing glitch. fixed.

PR #329, opened 2019-08-14
chore: hide `IdToken#getJsonWebSignature()`

chingor13:
I think we should hide `getJsonWebSignature()` for now until we know if we actually want to expose this class or a version from a different, better supported JWT library.

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/IdToken.java
+++ b/oauth2_http/java/com/google/auth/oauth2/IdToken.java
@@ -85,7 +85,7 @@ public class IdToken extends AccessToken implements Serializable {
    *
    * @return returns com.google.api.client.json.webtoken.JsonWebSignature
    */
-  public JsonWebSignature getJsonWebSignature() {
+  JsonWebSignature getJsonWebSignature() {
     return jsonWebSignature;
   }
 


chingor13:
cc @salrashid123

salrashid123:
How about we provide a way to return the rawToken instead?

for example,  if there was a method that said `IdToken.getRawToken()` that returned the same stuff that was passed into the constructor, then i could pass that to a method like the one below:

```java
verifyToken(IdToken.getRawToken(), "https://foo.bar")
```


```java
     public static boolean verifyToken(String id_token, String audience) throws Exception {
          JacksonFactory jacksonFactory = new JacksonFactory();
          GoogleIdTokenVerifier verifier = new GoogleIdTokenVerifier.Builder(new NetHttpTransport(), jacksonFactory)
                    .setAudience(Collections.singletonList(audience)).build();
          GoogleIdToken idToken = verifier.verify(id_token);
          if (idToken != null) {
               Payload payload = idToken.getPayload();
               // System.out.println(payload);
               return true;
          } else {
               return false;
          }
     }
```

currently, i'd have to extract out the embedded stuff from JSONWebToken
```java
System.out.println(tokenCredential.getIdToken().getJsonWebSignature().getPayload().getAudienceAsList());
```
and do some comparision..


---

i'm fine if we hide it but would suggest replacing it with the raw token somehow..

i'm trying to fix our JsonWebToken stuff here but its a bit of a mess
https://github.com/googleapis/google-http-java-client/pull/757

chingor13:
It's already available as `String getTokenValue()` (inherited from `AccessToken`)

salrashid123:
oh, right!.  then sgtm

PR #330, opened 2019-08-14
chore: cleanup javadoc warnings

chingor13:


oauth2_http/java/com/google/auth/oauth2/JwtClaims.java
```diff
@@ -89,7 +89,7 @@ public JwtClaims merge(JwtClaims other) {
    * ServiceAccountJwtAccessCredentials#jwtWithClaims(JwtClaims)} or {@link
    * JwtCredentials#jwtWithClaims(JwtClaims)}.
    *
-   * @return
+   * @return whether or not all required fields have been set
```
elharo: "or not" is not needed; better: true if all required fields have been set; false otherwise

chingor13: done

PR #331, opened 2019-08-14
feat: allow arbitrary additional claims for JwtClaims

chingor13:


oauth2_http/java/com/google/auth/oauth2/JwtClaims.java
```diff
@@ -61,8 +63,15 @@
   @Nullable
   abstract String getSubject();
 
+  /**
+   * Returns additional claims for this object; <b>do not mutate</b> the returned map.
```
elharo: Why not? Should this simply return an ImmutableMap instead?

oauth2_http/javatests/com/google/auth/oauth2/JwtClaimsTest.java
```diff
@@ -0,0 +1,136 @@
...
+import static org.junit.Assert.*;
+
+import java.util.Collections;
+import java.util.Map;
+import org.junit.Test;
+
+public class JwtClaimsTest {
+
+  @Test
+  public void claims_merge_overwritesFields() {
```
elharo: This is an unusual name. Consider testMergeOverwritesFields instead. I'm not sure but claims_merge_overwritesFields might have some checkstyle issues.

oauth2_http/java/com/google/auth/oauth2/JwtClaims.java
chingor13: We generally don't want to expose Guava types in the API surface, so we make it a generic map and document it as immutable

oauth2_http/javatests/com/google/auth/oauth2/JwtClaimsTest.java
chingor13: done

oauth2_http/java/com/google/auth/oauth2/JwtClaims.java
elharo: Yes, this is an ugliness in the Java Collections API. I tend to prefer simply returning a mutable copy to avoid UnsupportedOperationExceptions, but if that's not a good idea maybe change "; <b>do not mutate</b> the returned map" to "as an immutable map" or perhaps write as, "The map is not guaranteed to be mutable." The way it's written now sounds like you could mutate it but bad things would then happen.

chingor13: updated

PR #355, opened 2019-10-04
deps: make auto-value-annotations an optional dependency

chingor13:
Really it's a compile-only dependency which Maven doesn't support.


chingor13:
I don't think this is actually necessary.

PR #383, opened 2019-12-09
feat: support reading in quotaProjectId for billing

codyoss:


oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -166,6 +168,24 @@ public static GoogleCredentials fromStream(
             fileType, USER_FILE_TYPE, SERVICE_ACCOUNT_FILE_TYPE));
   }
 
+  /**
+   * Adds quota project ID to requestMetadata if present.
+   *
+   * @return a new map with quotaProjectId added if needed, or else returns the original
```
elharo: This is weird: sometimes it's a copy and sometimes it isn't? Pick one or the other. If this is only used internally, it's probably OK to modify the map.

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -75,11 +75,7 @@
 import java.security.SignatureException;
 import java.security.spec.InvalidKeySpecException;
 import java.security.spec.PKCS8EncodedKeySpec;
-import java.util.Collection;
-import java.util.Date;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
+import java.util.*;
```
elharo: no wild cards

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -119,6 +116,8 @@
    * @param tokenServerUri URI of the end point that provides tokens.
    * @param serviceAccountUser Email of the user account to impersonate, if delegating domain-wide
    *     authority to the service account.
+   * @param projectId The project used for billing.
```
elharo: nit: The --> the
and no period when it isn't a complete sentence

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -291,6 +295,49 @@ public static ServiceAccountCredentials fromPkcs8(
         null);
   }
 
+  /**
+   * Factory with minimum identifying information and custom transport using PKCS#8 for the private
+   * key.
+   *
+   * @param clientId Client ID of the service account from the console. May be null.
```
elharo: client ID
and similar below

oauth2_http/java/com/google/auth/oauth2/UserCredentials.java
```diff
@@ -52,9 +52,7 @@
 import java.io.InputStream;
 import java.io.ObjectInputStream;
 import java.net.URI;
-import java.util.Date;
-import java.util.Map;
-import java.util.Objects;
+import java.util.*;
```
elharo: no wild cards

oauth2_http/javatests/com/google/auth/oauth2/ServiceAccountCredentialsTest.java
```diff
@@ -338,7 +339,12 @@ public void fromJSON_getProjectId() throws IOException {
     transportFactory.transport.addServiceAccount(SA_CLIENT_EMAIL, ACCESS_TOKEN);
     GenericJson json =
         writeServiceAccountJson(
-            SA_CLIENT_ID, SA_CLIENT_EMAIL, SA_PRIVATE_KEY_PKCS8, SA_PRIVATE_KEY_ID, PROJECT_ID);
+            SA_CLIENT_ID,
```
elharo: SA == SERVICE_ACCOUNT?
avoid abbreviations

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
codyoss: upstream the map unmodifiable. I will just always make a new map in this case then.

oauth2_http/javatests/com/google/auth/oauth2/ServiceAccountCredentialsTest.java
codyoss: I would assume so, I did not add this, but I can expand.

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -291,6 +299,49 @@ public static ServiceAccountCredentials fromPkcs8(
...
+   * @param transportFactory HTTP transport factory, creates the transport used to get access
+   *     tokens.
+   * @param tokenServerUri URI of the end point that provides tokens.
+   * @param serviceAccountUser the email of the user account to impersonate, if delegating
+   *     domain-wide authority to the service account.
+   * @param quotaProjectId the project used for quota and billing purposes. May be null.
+   * @return New ServiceAccountCredentials created from a private key.
+   * @throws IOException if the credential cannot be created from the private key.
+   */
+  public static ServiceAccountCredentials fromPkcs8(
```
chingor13: I'm not a huge fan of adding more and more of these helpers for each combination of options. We should consider our options around this design for 1.0.

oauth2_http/javatests/com/google/auth/oauth2/ServiceAccountCredentialsTest.java
```diff
@@ -1032,6 +1054,9 @@ static GenericJson writeServiceAccountJson(
     if (projectId != null) {
       json.put("project_id", projectId);
     }
+    if (quotaProjectId != null) {
```
chingor13: is it possible for the service account json file to have a null entry and is it worth writing a test case for it?

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -166,6 +172,22 @@ public static GoogleCredentials fromStream(
...
   }
 
+  /**
+   * Adds quota project ID to requestMetadata if present.
+   *
+   * @return a new map with quotaProjectId added if needed
+   */
+  static Map<String, List<String>> addQuotaProjectIdToRequestMetadata(
+      String quotaProjectId, Map<String, List<String>> requestMetadata) {
+    Preconditions.checkNotNull(requestMetadata, "requestMetadata");
```
chingor13: nit: this exception message is not particularly useful. Let's make it better or remove the message.

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
codyoss: agreed. Maybe use a builder eventually.

oauth2_http/javatests/com/google/auth/oauth2/ServiceAccountCredentialsTest.java
codyoss: It is, until this feature is live, this value will always be null. This is being implicitly tested via other tests that don't set this field. I am not sure I could add much more value with another test. Unless I am missing something?

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
chingor13: IMO, if this is not necessary for our use case, let's not add it if we're considering deprecating it soon.

codyoss: fair, will revert these changes.

oauth2_http/java/com/google/auth/oauth2/ServiceAccountJwtAccessCredentials.java
```diff
@@ -203,7 +209,32 @@ public static ServiceAccountJwtAccessCredentials fromPkcs8(
...
+   * @param clientId Client ID of the service account from the console. May be null.
+   * @param clientEmail Client email address of the service account from the console.
+   * @param privateKeyPkcs8 RSA private key object for the service account in PKCS#8 format.
+   * @param privateKeyId Private key identifier for the service account. May be null.
+   * @param defaultAudience Audience to use if not provided by transport. May be null.
+   * @param quotaProjectId The project used for quota and billing purposes. May be null.
+   * @return New ServiceAccountJwtAcceessCredentials created from a private key.
+   * @throws IOException if the credential cannot be created from the private key.
+   */
+  public static ServiceAccountJwtAccessCredentials fromPkcs8(
```
chingor13: This one is also unnecessary

chingor13: Is this one actually needed?

codyoss: I would say yes actually for this one. I moved it to package scope though as to not expose more surface to the api.

PR #419, opened 2020-04-22
feat: add ability to verify id tokens

chingor13:
This adds the ability to verify JWT tokens that are signed using either RS256 or ES256. The current proposed interface is a static method call `boolean TokenVerifier.verify(String, TokenOptions)` which throws a new `VerificationException` on failure.

By default, `TokenVerifier` only verifies the signatures and expiration.

Example usage:

```java
String token = "some JWT formatted string";
TokenVerifier.VerifyOptions verifyOptions = 
    TokenVerifier.VerifyOptions.newBuilder()
        .setAudience("audience to verify")
        .setIssuer("issuer to verify")
        .build();
TokenVerifier.verify(token, verifyOptions);
```

## Verifying Signatures

After parsing the JWT, we look up the algorithm and key id (`kid`) from the JWT header. To find/build the `PublicKey` necessary to verify the signature, we look up the value from well-known locations for Google signing keys: (https://www.googleapis.com/oauth2/v3/certs for RS256 and https://www.gstatic.com/iap/verify/public_key-jwk for ES256). The `PublicKey`s parsed from these URLs are cached for 1 hour via a Guava `LoadingCache` implementation.

You can skip the key lookup by providing a `PublicKey` directly in `VerifyOptions`.

You can specify a custom Json Web Token Key Sey URL or x509 Certificates URL by specifying the `certificatesLocation` in `VerifyOptions`.

## Verifying Expiration

By default we compare the JWT expires at field with the current timestamp. For testing purposes, you can supply your own `Clock` interface to `VerifyOptions`.

## Verifying Audience

To verify the expected audience, set the audience field in `VerifyOptions`.

## Verify Issuer

To verify the expected issuer, set the issuer field in `VerifyOptions`.


chingor13:
Closing in favor of #420

PR #435, opened 2020-05-28
feat: add logging at FINE level for each step of ADC

### LINKED ISSUE #430: Lack of any logging or debugging flags to assist in troubleshooting
This plugin goes through a number of steps to try to find credentials. But there's no logging or any way to debug which of the methods was tried, which were unsuccessful, and whether any credentials were returned.  I'm trying to work out if my env var is even being seen by this plugin, but there's no way to tell.

chingor13:
Fixes #430

Also changing the INFO message that we couldn't detect compute engine credentials as that's a perfectly normal case and can be checked by the return value.

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java
+++ b/oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java
@@ -179,7 +179,11 @@ class DefaultCredentialsProvider {
       InputStream credentialsStream = null;
       try {
         if (isFile(wellKnownFileLocation)) {
-          LOGGER.log(Level.FINE, String.format("Loading credentials from well known file: %s", wellKnownFileLocation.getCanonicalPath()));
+          LOGGER.log(
+              Level.FINE,
+              String.format(
+                  "Loading credentials from well known file: %s",
+                  wellKnownFileLocation.getCanonicalPath()));
           credentialsStream = readStream(wellKnownFileLocation);
           credentials = GoogleCredentials.fromStream(credentialsStream, transportFactory);
         }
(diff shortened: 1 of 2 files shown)


oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java
```diff
@@ -145,6 +145,7 @@ private final GoogleCredentials getDefaultCredentialsUnsynchronized(
     GoogleCredentials credentials = null;
     String credentialsPath = getEnv(CREDENTIAL_ENV_VAR);
     if (credentialsPath != null && credentialsPath.length() > 0) {
+      LOGGER.log(Level.FINE, String.format("Loading credentials from file: %s", credentialsPath));
```
BenWhitehead, line 148: nit: "Attempting to load"

oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java
```diff
@@ -178,6 +179,11 @@ private final GoogleCredentials getDefaultCredentialsUnsynchronized(
       InputStream credentialsStream = null;
       try {
         if (isFile(wellKnownFileLocation)) {
+          LOGGER.log(
+              Level.FINE,
+              String.format(
+                  "Loading credentials from well known file: %s",
```
BenWhitehead, line 185: nit: "Attempting to load"

oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java
chingor13: Fixed.

oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java
chingor13: Fixed.

PR #455, opened 2020-08-04
chore: enable profile for AutoValue

chingor13:
Try the configuration from google-cloud-shared-config

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -66,6 +66,7 @@
     <project.autovalue.version>1.7.4</project.autovalue.version>
     <project.findbugs.version>3.0.2</project.findbugs.version>
     <deploy.autorelease>false</deploy.autorelease>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -428,5 +422,86 @@
...
+        <file>
+          <exists>${basedir}/EnableAutoValue.txt</exists>
+        </file>
+      </activation>
+      <properties>
+        <!--
+        We need to back pin to 1.4 because it is the last version of auto-value
+        that was compiled for java 1.7.
+        -->
+        <auto-value.version>1.4</auto-value.version>
```
lesv, line 438: This doesn't appear to be the right one.   `auto-value-annotation.version` ??

lesv: Ah - not needed.

PR #457, opened 2020-08-06
feat: ComputeEngineCredential supports specifying scopes

ahmedtd:
The existing behavior of ComputeEngineCredentials is to not provide the `?scopes` parameter in its access token request to the metadata server.  This is workable on GCE, which has the concept of a scope lock / default scopes applied to the VM.  However, neither new AppEngine nor GKE Workload Identity have this concept; they just return a hardcoded set of default scopes if no 
 specific scopes are requested.  This makes usage of some Google APIs impossible using the Java client libraries.

This change adds scope support to ComputeEngineCredentials.  It's optional, so the current behavior will stay the default. 
 However, if the user specifies scopes on the builder or using `createScoped`, then those scopes will be passed to the metadata server.

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
@@ -49,6 +49,7 @@ import java.io.InputStream;
 import java.io.ObjectInputStream;
 import java.net.SocketTimeoutException;
 import java.net.UnknownHostException;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.Date;
 import java.util.List;
@@ -94,16 +95,23 @@ public class ComputeEngineCredentials extends GoogleCredentials
 
   private final String transportFactoryClassName;
 
+  private final Collection<String> scopes;
+
   private transient HttpTransportFactory transportFactory;
   private transient String serviceAccountEmail;
 
   /**
-   * Constructor with overridden transport.
+   * Constructor with overridden scopes and transport.
+   *
+   * @param scopes Access scopes to request.  If null, the returned access token will have the
+   *     default scopes for the environment.  On GCE, these are the scopes assigned to the VM.  On
+   *     GKE Workload Identity, this will be cloud-platform and userinfo.email.
    *
    * @param transportFactory HTTP transport factory, creates the transport used to get access
    *     tokens.
    */
-  private ComputeEngineCredentials(HttpTransportFactory transportFactory) {
+  private ComputeEngineCredentials(Collection<String> scopes, HttpTransportFactory transportFactory) {
+    this.scopes = scopes;
     this.transportFactory =
         firstNonNull(
             transportFactory,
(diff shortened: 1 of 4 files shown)


ahmedtd:
@chingor13 This is the patch we discussed.

pom.xml
```diff
@@ -193,6 +193,7 @@
           <version>3.0.0-M5</version>
           <configuration>
             <reportNameSuffix>sponge_log</reportNameSuffix>
+			<trimStackTrace>false</trimStackTrace>
```
elharo, line 196: detab please

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -117,13 +126,27 @@ private ComputeEngineCredentials(HttpTransportFactory transportFactory) {
    * @return new ComputeEngineCredentials
    */
   public static ComputeEngineCredentials create() {
-    return new ComputeEngineCredentials(null);
+    return new ComputeEngineCredentials(ImmutableSet.<String>of(), null);
```
chingor13, line 129: Should this default to null instead?

An empty collection of scopes might make the url something like `https://some-url/?scopes=` which may be correct if you ask for no scopes, but if you pass null, it should exclude the scopes param.

chingor13:
cc @salrashid123

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
ahmedtd: Yes, you're right.  I originally wrote this so that an empty collection would result in not sending the `?scopes` param, but then changed it to null.

ahmedtd: Done

pom.xml
ahmedtd: Done

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -117,13 +125,27 @@ private ComputeEngineCredentials(HttpTransportFactory transportFactory) {
...
+    return new ComputeEngineCredentials(scopes, this.transportFactory);
   }
 
   /** Refresh the access token by getting it from the GCE metadata server */
   @Override
   public AccessToken refreshAccessToken() throws IOException {
-    HttpResponse response = getMetadataResponse(getTokenServerEncodedUrl());
+    GenericUrl url = new GenericUrl(getTokenServerEncodedUrl());
+    if (scopes != null) {
+      url.set("scopes", String.join(",", scopes));
```
chingor13, line 145: Looks like String.join was added in Java 8 and this library still currently support Java 7.

ahmedtd:
@elharo I opened #458 .  I have some follow up work to confirm that this works the same way in the AppEngine flex environment.

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
elharo: You can use Guava's Joiner in this class.

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -411,12 +437,21 @@ public Builder setHttpTransportFactory(HttpTransportFactory transportFactory) {
       return this;
     }
 
+    public Builder setScopes(Collection<String> scopes) {
+      this.scopes = scopes;
```
elharo, line 441: You probably want a defensive copy to avoid exposing mutable internal state.

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -411,12 +437,21 @@ public Builder setHttpTransportFactory(HttpTransportFactory transportFactory) {
...
+      return this;
+    }
+
     public HttpTransportFactory getHttpTransportFactory() {
       return transportFactory;
     }
 
+    public Collection<String> getScopes() {
```
elharo, line 449: You probably want a defensive copy to avoid exposing mutable internal state.

oauth2_http/javatests/com/google/auth/oauth2/MockMetadataServerTransport.java
```diff
@@ -80,14 +86,25 @@ public void setIdToken(String idToken) {
...
+    return this.receivedScopes;
+  }
+
   @Override
   public LowLevelHttpRequest buildRequest(String method, String url) throws IOException {
-    if (url.equals(ComputeEngineCredentials.getTokenServerEncodedUrl())) {
+    GenericUrl parsedURL = new GenericUrl(url);
+
+    if ("/computeMetadata/v1/instance/service-accounts/default/token".equals(parsedURL.getRawPath())) {
+      String scopes = (String)parsedURL.getFirst("scopes");
```
elharo, line 98: space after cast

chingor13:
Addressed in #514

PR #461, opened 2020-08-17
chore: fix license on token verifier classes

rafaelmenta:
Adjust the license from Apache to BSD 3, the one used by the project.

Fixes #260


chingor13:
cc @renaudd

chingor13:
These files were authored by Googlers (me) so we are ok fixing the license headers.

PR #350, opened 2019-09-26
fix: make JwtClaims.newBuilder() public

chingor13:
Fixes #349

PR #238, opened 2019-03-27
Add back in deprecated methods in ServiceAccountJwtAccessCredentials

kolea2:
To be deleted at a later time.

PR #473, opened 2020-08-31
fix!: privatize deprecated constructor

elharo:
@kolea2  one of last two deprecated methods to clean up before releasing 1.0

see https://github.com/googleapis/google-auth-library-java/pull/238

Can we take care of this now? I think https://www.github.com/googleapis/google-auth-library-java/issues/350 means we can do this now.


elharo:
[ERROR] 7009: com.google.auth.oauth2.ServiceAccountJwtAccessCredentials: Accessibility of method 'public ServiceAccountJwtAccessCredentials(java.lang.String, java.lang.String, java.security.PrivateKey, java.lang.String)' has been decreased from public to private

is expected and this library is pre-1.0

clirr-ignored-differences.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- see https://www.mojohaus.org/clirr-maven-plugin/examples/ignored-differences.html -->
+<differences>
+  <difference>
+    <differenceType>7009</differenceType>
+    <className>com/google/auth/oauth2/ServiceAccountJwtAccessCredentials</className>
+    
+    com.google.auth.oauth2.ServiceAccountJwtAccessCredentials <<<<
+
```
chingor13, line 9: This looks like a merge error?

pom.xml
```diff
@@ -305,6 +305,7 @@
         <configuration>
           <ignoredDifferencesFile>clirr-ignored-differences.xml</ignoredDifferencesFile>
           <logResults>true</logResults>
+          <failOnError>false</failOnError>
```
chingor13, line 308: I think we want to keep having it fail on non-ignored breaking changes.

chingor13 (changes requested):

clirr-ignored-differences.xml
elharo: fixed

pom.xml
elharo: This is pre-1.0

chingor13 (approved):

PR #476, opened 2020-09-17
feat: implements AWS signature version 4 for signing requests

lsirac:
https://docs.aws.amazon.com/general/latest/gr/signature-version-4.html


oauth2_http/java/com/google/auth/oauth2/AwsRequestSignature.java
```diff
@@ -0,0 +1,191 @@
...
+  }
+
+  /** Returns the HTTP request method. */
+  String getHttpMethod() {
+    return httpMethod;
+  }
+
+  /** Returns the HTTP request canonical headers. */
+  Map<String, String> getCanonicalHeaders() {
+    return canonicalHeaders;
```
elharo, line 102: This exposes mutable internal state. A copy is needed.

oauth2_http/java/com/google/auth/oauth2/AwsRequestSignature.java
```diff
@@ -0,0 +1,191 @@
...
+  }
+
+  /** Returns the request date. */
+  String getDate() {
+    return date;
+  }
+
+  /** Returns a list of sorted headers (in lexicographical order). */
+  List<String> getSortedHeaderNames() {
+    return sortedHeaderNames;
```
elharo, line 112: This exposes mutable internal state. A copy is needed.

oauth2_http/java/com/google/auth/oauth2/AwsRequestSignature.java
```diff
@@ -0,0 +1,191 @@
...
+      this.url = url;
+      return this;
+    }
+
+    Builder setHttpMethod(String httpMethod) {
+      this.httpMethod = httpMethod;
+      return this;
+    }
+
+    Builder setCanonicalHeaders(Map<String, String> canonicalHeaders) {
```
elharo, line 158: This injects mutable internal state. A copy is needed.

oauth2_http/java/com/google/auth/oauth2/AwsRequestSignature.java
```diff
@@ -0,0 +1,191 @@
...
+      this.canonicalHeaders = canonicalHeaders;
+      return this;
+    }
+
+    Builder setDate(String date) {
+      this.date = date;
+      return this;
+    }
+
+    Builder setSortedHeaderNames(List<String> sortedHeaderNames) {
```
elharo, line 168: This injects mutable internal state. A copy is needed.

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,347 @@
...
+      String url,
+      String region,
+      @Nullable String requestPayload,
+      @Nullable Map<String, String> additionalHeaders) {
+    this.awsSecurityCredentials = checkNotNull(awsSecurityCredentials);
+    this.httpMethod = checkNotNull(httpMethod);
+    this.uri = URI.create(url).normalize();
+    this.region = checkNotNull(region);
+    this.requestPayload = requestPayload == null ? "" : requestPayload;
+    this.additionalHeaders =
```
elharo, line 110: This injects mutable internal state. A copy is needed.

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,347 @@
...
+        .setSortedHeaderNames(sortedHeaders)
+        .setRegion(region)
+        .build();
+  }
+
+  /** Task 1: Create a canonical request for Signature Version 4. */
+  private String createCanonicalRequestHash(
+      Map<String, String> headers, List<String> sortedHeaderNames) {
+    // Append the HTTP request method.
+    StringBuilder sb = new StringBuilder(httpMethod).append(NEW_LINE);
```
elharo, line 164: Google style is not to use abbreviations like sb. Perhaps url?

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,347 @@
...
+  private static final String AWS_DATE_FORMAT = "yyyyMMdd'T'HHmmss'Z'";
+
+  // AWS Signature Version 4 signing algorithm identifier.
+  private static final String HASHING_ALGORITHM = "AWS4-HMAC-SHA256";
+
+  // The termination string for the AWS credential scope value as defined in
+  // https://docs.aws.amazon.com/general/latest/gr/sigv4-create-string-to-sign.html
+  private static final String AWS_REQUEST_TYPE = "aws4_request";
+
+  private static final String NEW_LINE = "\n";
```
elharo, line 77: Inline this. The constant doesn't help.

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,347 @@
...
+    // Get the dates to be used to sign the request.
+    AwsDates dates = getDates();
+
+    // Retrieve the service name. e.g. iam.amazonaws.com host => iam service.
+    String serviceName = Splitter.on(".").split(uri.getHost()).iterator().next();
+
+    Map<String, String> canonicalHeaders = getCanonicalHeaders(dates.getOriginalDate());
+    // Headers must be sorted.
+    List<String> sortedHeaders = new ArrayList<>();
+    for (String k : canonicalHeaders.keySet()) {
```
elharo, line 129: k --> key or better yet headerName

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,347 @@
...
+    if (additionalHeaders.containsKey("x-amz-date")) {
+      xAmzDate = additionalHeaders.get("x-amz-date");
+    }
+
+    if (additionalHeaders.containsKey("date")) {
+      String date = additionalHeaders.get("date");
+      Date inputDate;
+      try {
+        inputDate = new SimpleDateFormat("E, dd MMM yyyy HH:mm:ss z").parse(date);
+      } catch (ParseException e) {
```
elharo, line 252: Make the try block cover more lines so the variable declaration and initialization aren't split

oauth2_http/javatests/com/google/auth/oauth2/AwsRequestSignerTest.java
```diff
@@ -0,0 +1,461 @@
...
+ *
+ * <p>Examples of sigv4 signed requests:
+ * https://docs.aws.amazon.com/general/latest/gr/sigv4-signed-request-examples.html
+ */
+public class AwsRequestSignerTest {
+
+  private static final String US_EAST_1 = "us-east-1";
+  private static final String US_EAST_2 = "us-east-2";
+
+  private static final String GET = "GET";
```
elharo, line 58: inline these

oauth2_http/javatests/com/google/auth/oauth2/AwsRequestSignerTest.java
```diff
@@ -0,0 +1,461 @@
...
+
+/**
+ * Tests for {@link AwsRequestSigner}.
+ *
+ * <p>Examples of sigv4 signed requests:
+ * https://docs.aws.amazon.com/general/latest/gr/sigv4-signed-request-examples.html
+ */
+public class AwsRequestSignerTest {
+
+  private static final String US_EAST_1 = "us-east-1";
```
elharo, line 55: inline these

elharo (commented):
See https://twitter.com/elharo/status/976612920565125124 about constants

I need to take a closer look at the exceptions thrown here.

chingor13 (changes requested):

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,346 @@
...
+   *
+   * @param awsSecurityCredentials AWS security credentials.
+   * @param httpMethod The HTTP request method.
+   * @param url The request URL.
+   * @param region The targeted region.
+   * @param requestPayload The request payload.
+   * @param additionalHeaders A map of additional HTTP headers to be included with the signed
+   *     request.
+   */
+  public AwsRequestSigner(
```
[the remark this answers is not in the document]
lsirac: Should have been private - fixed

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,346 @@
...
+      Mac mac = Mac.getInstance(algorithm);
+      mac.init(new SecretKeySpec(key, algorithm));
+      return mac.doFinal(value);
+    } catch (NoSuchAlgorithmException | InvalidKeyException ex) {
+      throw new SigningException("Failed to calculate the AWS V4 Signature.", ex);
+    }
+  }
+
+  public static Builder newBuilder(
+      AwsSecurityCredentials awsSecurityCredentials, String httpMethod, String url, String region) {
```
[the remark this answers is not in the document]
lsirac: I changed the visibility of some of these classes - I don't want a user to use them directly

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,346 @@
...
+
+    String xAmzDate = null;
+    if (additionalHeaders.containsKey("x-amz-date")) {
+      xAmzDate = additionalHeaders.get("x-amz-date");
+    }
+
+    if (additionalHeaders.containsKey("date")) {
+      String date = additionalHeaders.get("date");
+      try {
+        Date inputDate = new SimpleDateFormat("E, dd MMM yyyy HH:mm:ss z").parse(date);
```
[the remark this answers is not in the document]
lsirac: I moved this logic into AwsDates, but opted not to make the date formatter a constant

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,358 @@
...
+    canonicalRequest.append(canonicalHeaders).append("\n");
+
+    // Append the signed headers.
+    canonicalRequest.append(Joiner.on(';').join(sortedHeaderNames)).append("\n");
+
+    // Append the hashed request payload.
+    canonicalRequest.append(DigestUtils.sha256Hex(requestPayload));
+
+    // Return the hashed canonical request.
+    return DigestUtils.sha256Hex(canonicalRequest.toString());
```
chingor13, line 189: Let's switch to guava's sha256 digester so we don't need to use commons-codec

chingor13: Actually, let's go with [MessageDigest](https://docs.oracle.com/javase/7/docs/api/java/security/MessageDigest.html)

oauth2_http/java/com/google/auth/oauth2/AwsRequestSignature.java
```diff
@@ -0,0 +1,193 @@
...
+      this.canonicalHeaders = new HashMap<>(canonicalHeaders);
+      return this;
+    }
+
+    Builder setDate(String date) {
+      this.date = date;
+      return this;
+    }
+
+    Builder setSortedHeaderNames(List<String> sortedHeaderNames) {
```
elharo, line 170: what if these come in unsorted? Nothing here sorts them or checks that they're sorted.

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,358 @@
...
+import java.util.Collections;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.TimeZone;
+import javax.annotation.Nullable;
+import javax.crypto.Mac;
+import javax.crypto.spec.SecretKeySpec;
+import org.apache.commons.codec.digest.DigestUtils;
```
elharo, line 58: Let's not add a dependency on this apache commons here.

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,358 @@
...
+  /**
+   * Signs the specified AWS API request.
+   *
+   * @return the {@link AwsRequestSignature}.
+   */
+  AwsRequestSignature sign() {
+    // Get the dates to be used to sign the request.
+    AwsDates dates = getDates();
+
+    // Retrieve the service name. e.g. iam.amazonaws.com host => iam service.
```
elharo, line 124: . e.g. --> such as
Google guidelines avoid Latin

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,358 @@
...
+    String serviceName = Splitter.on(".").split(uri.getHost()).iterator().next();
+
+    Map<String, String> canonicalHeaders = getCanonicalHeaders(dates.getOriginalDate());
+    // Headers must be sorted.
+    List<String> sortedHeaders = new ArrayList<>();
+    for (String headerName : canonicalHeaders.keySet()) {
+      sortedHeaders.add(headerName.toLowerCase());
```
elharo, line 131: needs a character set

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,358 @@
...
+        + "\n"
+        + credentialScope
+        + "\n"
+        + canonicalRequestHash;
+  }
+
+  /**
+   * Task 3: Calculate the signature for AWS Signature Version 4.
+   *
+   * @param date The date used in the hashing process in YYYYMMDD format.
```
elharo, line 207: the date
and no period

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,358 @@
...
+
+    static AwsDates fromDateHeader(String date) {
+      DateFormat dateFormat = new SimpleDateFormat(X_AMZ_DATE_FORMAT);
+      dateFormat.setTimeZone(TimeZone.getTimeZone("UTC"));
+      try {
+        Date inputDate = new SimpleDateFormat(CUSTOM_DATE_FORMAT).parse(date);
+        String xAmzDate = dateFormat.format(inputDate);
+        return new AwsDates(date, xAmzDate, xAmzDate.substring(0, 8));
+      } catch (ParseException e) {
+        throw new IllegalArgumentException("Invalid date provided: " + date, e);
```
elharo, line 329: this probably shouldn't be a runtime exception; better to bubble the ParseException

elharo (changes requested):
Don't use runtime exceptions for any probles that comes from outside the program such as a malformed HTTP header

oauth2_http/java/com/google/auth/oauth2/AwsRequestSignature.java
chingor13: Usuability wise, it might be easier for this to not be sorted and to ensure the sorting when we do the signing. Otherwise, it'll be quite a few places where we try to ensure sorting.

lsirac: Good points. I shouldn't have added this - removed.

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,362 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package com.google.auth.oauth2;
+
+import static com.google.api.client.util.Preconditions.checkNotNull;
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+import com.google.api.client.util.Clock;
```
elharo, line 37: Anyone know why we use these classes instead of the better maintained versions in Java?

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,362 @@
...
+        return new AwsDates(date, xAmzDate, xAmzDate.substring(0, 8));
+      } catch (ParseException e) {
+        throw new IllegalArgumentException("Invalid date provided: " + date, e);
+      }
+    }
+
+    static AwsDates generateXAmzDate() {
+      DateFormat dateFormat = new SimpleDateFormat(X_AMZ_DATE_FORMAT);
+      dateFormat.setTimeZone(TimeZone.getTimeZone("UTC"));
+      String xAmzDate = dateFormat.format(new Date(Clock.SYSTEM.currentTimeMillis()));
```
elharo, line 340: no need to use Clock here; just use System.getCurrentTimeMillis

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,362 @@
...
+    }
+
+    if (awsSecurityCredentials.getToken() != null && !awsSecurityCredentials.getToken().isEmpty()) {
+      headers.put("x-amz-security-token", awsSecurityCredentials.getToken());
+    }
+
+    // Add all additional headers.
+    for (String key : additionalHeaders.keySet()) {
+      // Header keys need to be lowercase.
+      headers.put(key.toLowerCase(), additionalHeaders.get(key));
```
elharo, line 229: needs locale

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,362 @@
...
+  }
+
+  private static byte[] sign(byte[] key, byte[] value) {
+    try {
+      String algorithm = "HmacSHA256";
+      Mac mac = Mac.getInstance(algorithm);
+      mac.init(new SecretKeySpec(key, algorithm));
+      return mac.doFinal(value);
+    } catch (NoSuchAlgorithmException | InvalidKeyException ex) {
+      throw new IllegalStateException("Failed to calculate the AWS V4 Signature.", ex);
```
elharo, line 251: Pretty sure this is not the right exception here. 

Also, what is the justification for converting this to a runtime exception?

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,362 @@
...
+    }
+  }
+
+  private static String getHexEncodedSha256Hash(byte[] bytes) {
+    try {
+      MessageDigest digest = MessageDigest.getInstance("SHA-256");
+      return BaseEncoding.base16().lowerCase().encode(digest.digest(bytes));
+    } catch (NoSuchAlgorithmException e) {
+      throw new IllegalStateException("Failed to compute SHA-256 hash.", e);
```
elharo, line 260: This is not an illegal state, and this is the wrong exception type to use for this problem.

How do we know SHA-256 is supported? Does the JDK promise this?

elharo (changes requested):

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,362 @@
...
+import javax.crypto.Mac;
+import javax.crypto.spec.SecretKeySpec;
+
+/**
+ * Internal utility that signs AWS API requests based on the AWS Signature Version 4 signing
+ * process.
+ *
+ * <p>https://docs.aws.amazon.com/general/latest/gr/signature-version-4.html
+ */
+public class AwsRequestSigner {
```
elharo, line 66: This class doesn't seem to have any pubic methods so it doesn't need to be public (or did I miss one?)

elharo (changes requested):

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
lsirac: I missed this - my bad! Fixed.

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
lsirac: Should I bubble this up to refreshAccessToken(), and then wrap & throw it as an IOException?

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
lsirac: What exception type should I use? If it's not an unchecked exception, do you want me to wrap it in a IOException and throw it in refreshAccessToken?

Good question - I'd look into it.

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
lsirac: What changes would you like to see here?

elharo: Depends on whether a runtime exception is called for here or not. What does it mean if HmacSHA256 is not supported? What does it mean if the key is invalid? Where does the key come from?

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
elharo: NoSuchAlgorithmException is correct unless there's something wrong with the VM if SHA-256 isn't supported.

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
elharo: What does this exception mean? It's probably not an I?O error if the data can be read, but if the exception is caused by external input it should be checked.

Choose the exception type that fits the problem. Don't force it into an existing type that means something different.

elharo (changes requested):

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
lsirac: It's not an IO error, but I can't just bubble up the parse exception. This utility is used in refreshAccessToken - having that throw new exceptions is a breaking change (?). 

Why do you think throwing an IllegalArgumentException is not the way to go here?

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
lsirac: If I throw NoSuchAlgorithmException here, I will have to handle it in refreshAccessToken in some way - I can't throw it directly. 

Also:

"Every implementation of the Java platform is required to support the following standard MessageDigest algorithms:
MD5
SHA-1
SHA-256"

https://docs.oracle.com/javase/7/docs/api/java/security/MessageDigest.html

elharo: In that case I think a RuntimeException referencing this info is appropriate, since it means the VM isn't following spec. This is essentially a "can't happen" exception, but it's not illegal state

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
elharo: Where does the date come from? IllegalArgumentException means the calling method has a bug and should not have passed this value to this method. That doesn't seem to be the case here.

lsirac: The date can be specified as an additional header. We use the x-amz-date header, so this code path is not used in our flow. I added this because the AWS implementation supports it, and it allows us to use their tests.

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
lsirac: https://docs.oracle.com/javase/7/docs/api/javax/crypto/Mac.html

"Every implementation of the Java platform is required to support the following standard Mac algorithms:

HmacMD5
HmacSHA1
HmacSHA256"

I'll use a RuntimeException here.

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,381 @@
...
+    String urlPath = uri.getRawPath().isEmpty() ? "/" : uri.getRawPath();
+    canonicalRequest.append(urlPath).append("\n");
+
+    // Append the canonical query string.
+    String actionQueryString = uri.getRawQuery() != null ? uri.getRawQuery() : "";
+    canonicalRequest.append(actionQueryString).append("\n");
+
+    // Append the canonical headers.
+    StringBuilder canonicalHeaders = new StringBuilder();
+    for (String headerName : sortedHeaderNames) {
```
elharo, line 176: You're not doing any URI escaping here. Is that OK in context?

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,381 @@
...
+
+    static AwsDates fromDateHeader(String date) {
+      DateFormat dateFormat = new SimpleDateFormat(X_AMZ_DATE_FORMAT);
+      dateFormat.setTimeZone(TimeZone.getTimeZone("UTC"));
+      try {
+        Date inputDate = new SimpleDateFormat(CUSTOM_DATE_FORMAT).parse(date);
+        String xAmzDate = dateFormat.format(inputDate);
+        return new AwsDates(date, xAmzDate, xAmzDate.substring(0, 8));
+      } catch (ParseException e) {
+        throw new RuntimeException("Invalid date provided: " + date, e);
```
elharo, line 352: I still think bubbling the parse exception is best here. Don't convert to a runtime exception until you absolutely have to.

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,381 @@
...
+    private String amzDate;
+    private String formattedDate;
+
+    private AwsDates(String originalDate, String amzDate, String formattedDate) {
+      this.originalDate = checkNotNull(originalDate);
+      this.amzDate = checkNotNull(amzDate);
+      this.formattedDate = checkNotNull(formattedDate);
+    }
+
+    static AwsDates fromXAmzDate(String xAmzDate) {
```
elharo, line 340: The date isn't validated. Do we know its good here?

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,381 @@
...
+    for (String key : additionalHeaders.keySet()) {
+      // Header keys need to be lowercase.
+      headers.put(key.toLowerCase(Locale.US), additionalHeaders.get(key));
+    }
+    return headers;
+  }
+
+  private AwsDates getDates() {
+    if (additionalHeaders.containsKey("date")) {
+      return AwsDates.fromDateHeader(additionalHeaders.get("date"));
```
elharo, line 255: Assuming the headers are read from a HTTP request to an external server, then a checked exception is expected here since the format is not guaranteed.

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,381 @@
...
+  }
+
+  private static byte[] sign(byte[] key, byte[] value) {
+    try {
+      String algorithm = "HmacSHA256";
+      Mac mac = Mac.getInstance(algorithm);
+      mac.init(new SecretKeySpec(key, algorithm));
+      return mac.doFinal(value);
+    } catch (NoSuchAlgorithmException | InvalidKeyException e) {
+      throw new RuntimeException("Failed to calculate the AWS V4 Signature.", e);
```
elharo, line 270: This only covers NoSuchAlgorithmException. InvalidKeyException is a separate issue that should be handled differently since the key being wrong does not indicate a broken JDK.

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,381 @@
...
+  private String region;
+  private String requestPayload;
+  private URI uri;
+
+  /**
+   * Internal constructor.
+   *
+   * @param awsSecurityCredentials AWS security credentials.
+   * @param httpMethod The HTTP request method.
+   * @param url The request URL.
```
elharo, line 86: the request URL (no period per Oracle guidelines)

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,381 @@
...
+   * @param region The targeted region.
```
elharo, line 87: the targeted region

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,381 @@
...
+   * @param requestPayload The request payload.
```
elharo, line 88: the request payload

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,381 @@
...
+   * @param additionalHeaders A map of additional HTTP headers to be included with the signed
```
elharo, line 89: a

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,381 @@
...
+    this.additionalHeaders =
+        (additionalHeaders != null)
+            ? new HashMap<>(additionalHeaders)
+            : new HashMap<String, String>();
+  }
+
+  /**
+   * Signs the specified AWS API request.
+   *
+   * @return the {@link AwsRequestSignature}.
```
elharo, line 113: no period

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
lsirac: Validated in the Builder now.

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
lsirac: Validation is done in the Builder now.

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
lsirac: Discussed offline - we'll throw a SigningException when InvalidKeyException is thrown.

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,406 @@
...
+  AwsRequestSignature sign() {
+    // Get the dates to be used to sign the request.
+    AwsDates dates = getDates();
+
+    // Retrieve the service name. For example: iam.amazonaws.com host => iam service.
+    String serviceName = Splitter.on(".").split(uri.getHost()).iterator().next();
+
+    Map<String, String> canonicalHeaders = getCanonicalHeaders(dates.getOriginalDate());
+    // Headers must be sorted.
+    List<String> sortedHeaders = new ArrayList<>();
```
chingor13, line 125: nit: let's name this `sortedHeaderNames`

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,406 @@
...
+    return String.format(
+        "%s Credential=%s/%s, SignedHeaders=%s, Signature=%s",
+        HASHING_ALGORITHM,
+        accessKeyId,
+        credentialScope,
+        Joiner.on(';').join(sortedHeaderNames),
+        signature);
+  }
+
+  private Map<String, String> getCanonicalHeaders(String date) {
```
chingor13, line 233: nit: name this `defaultDate` to signal that it's not always set?

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,406 @@
...
+      dateFormat.setTimeZone(TimeZone.getTimeZone("UTC"));
+
+      String xAmzDate = null;
+      try {
+        Date inputDate = new SimpleDateFormat(CUSTOM_DATE_FORMAT).parse(date);
+        xAmzDate = dateFormat.format(inputDate);
+      } catch (ParseException e) {
+        // Date is validated at construction.
+      }
+      return new AwsDates(date, xAmzDate, xAmzDate.substring(0, 8));
```
chingor13, line 378: This looks like it's going to NPE on a parse error: `xAmzDate` won't be set if the `parse(date)` throws a `ParseException`

chingor13 (changes requested):

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,406 @@
...
+    static AwsDates fromDateHeader(String date) {
+      DateFormat dateFormat = new SimpleDateFormat(X_AMZ_DATE_FORMAT);
+      dateFormat.setTimeZone(TimeZone.getTimeZone("UTC"));
+
+      String xAmzDate = null;
+      try {
+        Date inputDate = new SimpleDateFormat(CUSTOM_DATE_FORMAT).parse(date);
+        xAmzDate = dateFormat.format(inputDate);
+      } catch (ParseException e) {
+        // Date is validated at construction.
```
elharo, line 376: you should be able to convert to a date there then, instead of passing strings around. Then there's no exception to catch here.

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,381 @@
...
+    for (String headerName : sortedHeaderNames) {
+      canonicalHeaders.append(headerName).append(":").append(headers.get(headerName)).append("\n");
+    }
+    canonicalRequest.append(canonicalHeaders).append("\n");
+
+    // Append the signed headers.
+    canonicalRequest.append(Joiner.on(';').join(sortedHeaderNames)).append("\n");
+
+    // Append the hashed request payload.
+    canonicalRequest.append(getHexEncodedSha256Hash(requestPayload.getBytes()));
```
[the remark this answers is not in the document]
lsirac: Done in the last commit

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,381 @@
...
+
+    // Return the hashed canonical request.
+    return getHexEncodedSha256Hash(canonicalRequest.toString().getBytes());
```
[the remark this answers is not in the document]
lsirac: Done in the last commit

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
lsirac: Fixed - not sure if this is exactly what you meant but I think it makes sense.

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,397 @@
...
+    private static final String CUSTOM_DATE_FORMAT = "E, dd MMM yyyy HH:mm:ss z";
+
+    private String originalDate;
+    private String amzDate;
+    private String formattedDate;
+
+    private AwsDates(String originalDate, String amzDate, String formattedDate) {
+      this.originalDate = checkNotNull(originalDate);
+      this.amzDate = checkNotNull(amzDate);
+      this.formattedDate = checkNotNull(formattedDate);
```
chingor13, line 353: It looks like this value is always computed from the `amzDate` argument which puts burden on the caller to always do the `substring(0,8)` to take the YYYMMDD part from `amzDate`. Intentional for some use-case?

chingor13 (changes requested):
I think @elharo is suggesting keeping a `Date` instance rather than a collection of pre-formatted string values. Then formatting them on the way out when necessary.

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
lsirac: Hmm, I think having a Date instance would then burden the callers to convert the input (always a String) to a date, only to be converted back anyway.

I cleaned it up a little - you're right, there's no need to hold onto the formattedDate since it is always derived from the amzDate.

chingor13 (approved):
The classes are all package private and we can clean up in the final PR if necessary.

PR #490, opened 2020-10-15
fix: remove 1 hour limit for the impersonated token

shinfan:
Cloud IAM now supports extended 12h lifespan for access token:
https://cloud.google.com/iam/docs/creating-short-lived-service-account-credentials#sa-credentials-oauth

A separate PR should be made to allow developers to specify the desired lifespan of the granted token.


haogang:
Thanks @shinfan. Do we know which SDK release will include this change, and when will it be released to the maven repo?

elharo (dismissed):
tests are failing

chingor13 (approved):

shinfan:
Do we need to cut a release for this fix? @chingor13

PR #493, opened 2020-10-15
#492 id token from service account credential api

### LINKED ISSUE #492: feat: Generate id_token with user credential
Linked to the #468  and the PR #469, especially following [the comment](https://github.com/googleapis/google-auth-library-java/pull/469#issuecomment-700802060) of @silvolu, I would like to propose another approach, this time compliant with IAP.

Instead of using the user credential to create an id_token with generic client_id and client_secret, I propose to use the [Service Account Credentials API generateIdToken method](https://cloud.google.com/iam/docs/reference/credentials/rest/v1/projects.serviceAccounts/generateIdToken). 

This method required to use a service account, and not the user credential. By default, I propose to use the Compute Engine default service account of the current quotas project id.

The user can override this value by setting an environment variable `SERVICE_ACCOUNT_APPLICATION_CREDENTIALS` with the email of the service account to use.

The token is generated by calling the Service Account Credentials API generateIdToken method. I propose to reuse the existing method from IamUtils class

Because this method won't use the user credential (role and autorisation) and because it could be a trap for the user, I propose to print clear a warning message the first time that the id_token is generated. In addition, this message informs the users that they need to have the Service Account Token Creator role granted on the service account (printed in the logs also).

I will be happy to discuss about this in this issue or in the coming PR (I worked on the code to validate the feasibility)

Best

guillaumeblaquiere:
Fixes #492, Could replace the #469 which seems not accepted by @silvolu 

Happy to discuss on this. The Env var name are important because this implementation could be reproduce in all other Google auth libraries (in other language) to have a unique way to add this env var (like today the `GOOGLE_APPLICATION_CREDENTIALS` name)

Best

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/UserCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/UserCredentials.java
@@ -40,11 +40,13 @@ import com.google.api.client.http.HttpRequest;
 import com.google.api.client.http.HttpRequestFactory;
 import com.google.api.client.http.HttpResponse;
 import com.google.api.client.http.UrlEncodedContent;
+import com.google.api.client.http.json.JsonHttpContent;
 import com.google.api.client.json.GenericJson;
 import com.google.api.client.json.JsonFactory;
 import com.google.api.client.json.JsonObjectParser;
 import com.google.api.client.util.GenericData;
 import com.google.api.client.util.Preconditions;
+import com.google.auth.http.HttpCredentialsAdapter;
 import com.google.auth.http.HttpTransportFactory;
 import com.google.common.base.MoreObjects;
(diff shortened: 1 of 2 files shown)


silvolu:
Hi Guillaume,

Apologies for the delay in replying on this, I've had some very busy weeks :/

I've been talking to some of the other teams involved, and we want to support your use case but there is some work that we need to do to design some part of this in the larger context of enabling and simplifying sharing credentials across tools (e.g. CLI and client libraries).

We've already started work on this, I'll keep you posted.

guillaumeblaquiere:
@silvolu thanks you for your feedback. Let me know if I can help. It will be a pleasure.
Best

elharo:
closing per @silvolu's comment

PR #501, opened 2020-10-23
chore: use ImpersonatedCredentials for service account impersonation for 3pi

lsirac:
When the service_account_impersonation_url is provided, we must exchange the external account GCP access token for a service account impersonated token.

This PR removes the duplicated code in ExternalAccountCredentials that makes the IAM generateAccessToken call and delegates it to ImpersonatedCredentials.

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
@@ -34,27 +34,18 @@ package com.google.auth.oauth2;
 import static com.google.api.client.util.Preconditions.checkNotNull;
 import static com.google.common.base.MoreObjects.firstNonNull;
 
-import com.google.api.client.http.GenericUrl;
 import com.google.api.client.http.HttpHeaders;
-import com.google.api.client.http.HttpRequest;
-import com.google.api.client.http.HttpResponse;
-import com.google.api.client.http.UrlEncodedContent;
 import com.google.api.client.json.GenericJson;
 import com.google.api.client.json.JsonObjectParser;
-import com.google.api.client.util.GenericData;
-import com.google.auth.http.AuthHttpConstants;
 import com.google.auth.http.HttpTransportFactory;
 import com.google.auth.oauth2.AwsCredentials.AwsCredentialSource;
 import com.google.auth.oauth2.IdentityPoolCredentials.IdentityPoolCredentialSource;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
-import java.text.DateFormat;
-import java.text.ParseException;
-import java.text.SimpleDateFormat;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.Date;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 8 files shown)


oauth2_http/javatests/com/google/auth/oauth2/AwsCredentialsTest.java
```diff
@@ -293,7 +297,7 @@ public void createdScoped_clonedCredentialWithAddedScopes() {
     AwsCredentials credentials =
         (AwsCredentials)
             AwsCredentials.newBuilder(AWS_CREDENTIAL)
-                .setServiceAccountImpersonationUrl("tokenInfoUrl")
+                .setServiceAccountImpersonationUrl(SERVICE_ACCOUNT_IMPERSONATION_URL)
```
chingor13, line 300: It feels weird to be providing the full impersonation URL vs the `targetPrinciple` (it looks like we're just parsing it out of the URL). Is this full URL being provided by some credential file?

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
```diff
@@ -273,52 +298,16 @@ protected AccessToken exchange3PICredentialForAccessToken(
...
-    try {
-      Date date = format.parse(expireTime);
-      return new AccessToken(token, date);
-    } catch (ParseException e) {
-      throw new IOException("Error parsing expireTime: " + e.getMessage());
+    if (startIndex != -1 && endIndex != -1 && startIndex < endIndex) {
+      return serviceAccountImpersonationUrl.substring(startIndex + 1, endIndex);
+    } else {
+      throw new IllegalArgumentException(
+          "Unable to determine target principal from service account impersonation URL.");
```
chingor13, line 310: This method is being called lazily (once the credentials are being used). This will be a weird runtime exception thrown that we should catch while building/instantiating the credentials.

chingor13 (changes requested):

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
lsirac: This is actually called in the constructor (from initializeImpersonatedCredentials)

oauth2_http/javatests/com/google/auth/oauth2/AwsCredentialsTest.java
lsirac: Yes - the URL is being provided by a credential file that will be generated by gcloud.

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
chingor13: For a confidence check, can we make sure to have a test to ensure this is happening early (at instantiation time).

chingor13 (approved):

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
lsirac: Done!

chingor13 (approved):

PR #509, opened 2020-11-30
fix: quotaProjectId should be applied for cached `getRequestMetadata(URI, Executor, RequestMetadataCallback)`

### LINKED ISSUE #507: Quota project id broken for OAuth2Credentials's async getRequestMetadata()
#### Environment details

- grpc: 1.35.0-SNAPSHOT (which uses the async API)
- google-auth-library-java version(s): 0.22.0

#### Steps to reproduce

1. Use `GoogleCredentials.getApplicationDefault()`
2. Use sync `getRequestMetadata()`. `x-goog-user-project` is present
3. Use async `getRequestMetadata()`, with the `RequestMetadataCallback`. `x-goog-user-project` is missing

#### Stacktrace

I was trying to do some development with a weird setup. But I encountered:
```
Status{code=PERMISSION_DENIED, description=Your application has authenticated using end user credentials from the Google Cloud SDK or Google Cloud Shell which are not supported by the trafficdirector.googleapis.com. We recommend configuring the billing/quota_project setting in gcloud or using a service account through the auth/impersonate_service_account setting. For more information about service accounts and how to use them in your application, see https://cloud.google.com/docs/authentication/., cause=null}
```

#### Code snippet

```java
Credentials creds = GoogleCredentials.getApplicationDefault();
// I had used "gcloud auth application-default login", so these were UserCredentials
System.out.println("creds: " + creds);
System.out.println("metadata: " + creds.getRequestMetadata());
creds.getRequestMetadata(null, null, new RequestMetadataCallback() {
  public void onSuccess(Map<String, List<String>> metadata) {
    System.out.println("async metadata: " + metadata);
  }

  public void onFailure(Throwable exception) {}
});
```

#### What's going on

So it is ["obvious"](https://github.com/googleapis/google-auth-library-java/blob/3b50627ae745ca92547530278e1781da8c46b7f7/oauth2_http/java/com/google/auth/oauth2/UserCredentials.java#L282) that `x-goog-user-project` is being added to the metadata. However, that's the slow path. `OAuth2Credentials` has a cache that it uses when available (`requestMetadata`), [to avoid jumping threads](https://github.com/googleapis/google-auth-library-java/blob/3b50627ae745ca92547530278e1781da8c46b7f7/oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java#L127) when `RequestMetadataCallback` is specified. The cache is only set in `useAccessToken(AccessToken)`, [which is called in `refresh()`](https://github.com/googleapis/google-auth-library-java/blob/3b50627ae745ca92547530278e1781da8c46b7f7/oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java#L157). So `requestMetadata` will never be processed by `addQuotaProjectIdToRequestMetadata()`, and hence the async code lacks the quota project if the cache is already available.

Aside: It's confusing that `quota_project_id` is the name used in the JSON key file. Normally "project id" is an integer while "project name" is the pretty name.

chingor13:
Refactors the injection of the quota project id (and possibly other future headers) into the entrypoint that caches the request metadata (in OAuth2Credentials).

Fixes #507

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
@@ -167,6 +167,11 @@ public class OAuth2Credentials extends Credentials {
     }
   }
 
+  /**
+   * Provide additional headers to return as request metadata.
+   *
+   * @return Map of additional headers.
+   */
   protected Map<String, List<String>> getAdditionalHeaders() {
     return EMPTY_EXTRA_HEADERS;
   }
--- a/oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
@@ -170,7 +170,7 @@ public class OAuth2Credentials extends Credentials {
   /**
    * Provide additional headers to return as request metadata.
    *
-   * @return Map of additional headers.
+   * @return additional headers
    */
   protected Map<String, List<String>> getAdditionalHeaders() {
     return EMPTY_EXTRA_HEADERS;
--- a/oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
@@ -38,6 +38,7 @@ import com.google.auth.http.AuthHttpConstants;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.MoreObjects;
 import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Iterables;
 import java.io.IOException;
 import java.io.ObjectInputStream;
@@ -56,6 +57,7 @@ public class OAuth2Credentials extends Credentials {
 
   private static final long serialVersionUID = 4556936364828217687L;
   private static final long MINIMUM_TOKEN_MILLISECONDS = 60000L * 5L;
+  private static final Map<String, List<String>> EMPTY_EXTRA_HEADERS = Collections.emptyMap();
 
   // byte[] is serializable, so the lock variable can be final
   private final Object lock = new byte[0];
@@ -89,7 +91,7 @@ public class OAuth2Credentials extends Credentials {
    */
   protected OAuth2Credentials(AccessToken accessToken) {
     if (accessToken != null) {
-      useAccessToken(accessToken);
+      useAccessToken(accessToken, EMPTY_EXTRA_HEADERS);
     }
   }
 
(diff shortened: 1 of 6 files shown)


oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
```diff
@@ -154,7 +156,9 @@ public void refresh() throws IOException {
     synchronized (lock) {
       requestMetadata = null;
       temporaryAccess = null;
-      useAccessToken(Preconditions.checkNotNull(refreshAccessToken(), "new access token"));
+      useAccessToken(
+          Preconditions.checkNotNull(refreshAccessToken(), "new access token"),
```
elharo, line 160: This is a very strange and possibly inappropriate use of Preconditions.

oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
```diff
@@ -163,6 +167,15 @@ public void refresh() throws IOException {
     }
   }
 
+  /**
+   * Provide additional headers to return as request metadata.
+   *
+   * @return Map of additional headers.
```
elharo, line 173: nit: "map of additional headers" or just "additional headers" since the map is provided by the type

oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
```diff
@@ -56,6 +57,7 @@
 
   private static final long serialVersionUID = 4556936364828217687L;
   private static final long MINIMUM_TOKEN_MILLISECONDS = 60000L * 5L;
+  private static final Map<String, List<String>> EMPTY_EXTRA_HEADERS = Collections.emptyMap();
```
elharo, line 60: will this be added to by subclasses? I'm not quite following but it looks like it might be; but this is immutable.

oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
```diff
@@ -163,6 +167,15 @@ public void refresh() throws IOException {
     }
   }
 
+  /**
+   * Provide additional headers to return as request metadata.
+   *
+   * @return Map of additional headers.
+   */
+  protected Map<String, List<String>> getAdditionalHeaders() {
```
elharo, line 175: nested types (Map of List) suggest there's a missed opportunity for a more semantic class.

oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
chingor13: `addQuotaProjectIdToRequestMetadata` returns a new `ImmutableMap` so nothing should be added to this (nor do we want anyone to).

oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
chingor13: I agree, it's been there for years.

Filed https://github.com/googleapis/google-auth-library-java/issues/510 to track.

oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
chingor13: HTTP headers behave as maps of `String`s to a list of `String`s. We could possible wrap it, but we're not adding any extra behavior. For a general purpose HTTP client, we might do so, but it seem unnecessary here.

Note: `getRequestMetadata()` does return this same type (but we really can't feasibly change that as everything that depends on google-auth-library expects that return type).

oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
chingor13: Changed to "additional headers"

stephaniewang526 (approved):

PR #514, opened 2020-12-09
feat: allow custom scopes for compute engine creds

### LINKED ISSUE #458: Users can't specify non-default scopes for code running on GKE Workload Identity (and possibly AppEngine flex)
The ComputeEngineCredentials class is used for pulling access and identity tokens when the library is running on GCE, GKE Workload Identity, and AppEngine flex.  It doesn't support setting scopes on the resulting access token, because on GCE scopes are handled by assigning them to the VM.

However, this isn't the case on GKE Workload Identity (and possibly AppEngine flex, I am still confirming this).  In this environment, there is no concept of a scope lock or default scopes for the access token.  Instead the metadata server accepts a `?scopes` query parameter on the `/computeMetadata/v1/instance/service-accounts/default/token` request and creates an access token with the requested scopes.

If no `?scopes` are specified, then the returned access tokens has a hardcoded default set of scopes (`cloud-platform` and `userInfo.email`).

This means that users of ComputeEngineCredentials don't have a way to use non-cloud Google APIs when running on GKE Workload Identity, since non-cloud APIs are not included in the `cloud-platform` access scope.

The GCE Metadata Server accepts the `?scopes` parameter without complaint, but ignores it.

I'd like to extend the ComputeEngineCredentials class with a `createScoped` implementation that will make a new ComputeEngineCredentials instance that passes the `?scopes` query parameter, so that code like this will work when running on GKE Workload Identity:

```
GoogleCredentials.getApplicationDefault().createScoped("https://www.googleapis.com/auth/androidpublisher");
```

arithmetic1728:
Fixes #458


arithmetic1728:
The `format/format-code` check failure is not related to this PR. It seems something is wrong with the `/home/runner/work/_actions/googleapis/code-suggester/v1.8.0/Dockerfile` docker build.

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -40,15 +40,18 @@
 import com.google.api.client.http.HttpStatusCodes;
 import com.google.api.client.json.JsonObjectParser;
 import com.google.api.client.util.GenericData;
+import com.google.api.client.util.Joiner;
```
elharo, line 43: just use Guava, which this proxies to.

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -102,13 +107,22 @@
    *
    * @param transportFactory HTTP transport factory, creates the transport used to get access
    *     tokens.
+   * @param scopes Scope strings for the APIs to be called. May be null or an empty collection.
```
elharo, line 110: Scope --> scope

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -102,13 +107,22 @@
...
    */
-  private ComputeEngineCredentials(HttpTransportFactory transportFactory) {
+  private ComputeEngineCredentials(
+      HttpTransportFactory transportFactory, Collection<String> scopes) {
     this.transportFactory =
         firstNonNull(
             transportFactory,
             getFromServiceLoader(HttpTransportFactory.class, OAuth2Utils.HTTP_TRANSPORT_FACTORY));
     this.transportFactoryClassName = this.transportFactory.getClass().getName();
+    this.scopes = (scopes == null) ? ImmutableSet.<String>of() : ImmutableSet.copyOf(scopes);
```
elharo, line 119: what if the collection is not empty but a string in the collection is null or empty?

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
arithmetic1728: done

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
arithmetic1728: done

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
arithmetic1728: modified the code to remove null and empty string from scopes, and updated the test

silvolu (approved):

silvolu:
Jeff, Elliot, recently updated AIP for this at https://google.aip.dev/auth/4115

chingor13 (approved):

PR #515, opened 2020-12-10
feat: allow custom lifetime for ImpersonatedCredentials

### LINKED ISSUE #474: Support for generating a OAuth 2.0 access tokens with lifespan > 1 hour
By default, OAuth 2.0 access tokens are valid for a maximum of 1 hour (3,600 seconds). However, the maximum lifetime for these tokens can be extended to 12 hours (43,200 seconds) according to the documentation: https://cloud.google.com/iam/docs/creating-short-lived-service-account-credentials#sa-credentials-oauth

This is currently not possible since the maximum 1 hour lifespan is hard-coded into both the [ImpersonatedCredentials](https://github.com/googleapis/google-auth-library-java/blob/master/oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java#L221) and the [ServiceAccountCredentials](https://github.com/googleapis/google-auth-library-java/blob/master/oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java#L660) class.

arithmetic1728:
Addresses #474 for impersonated credentials. 

The default maximum lifetime for token is 1 hour, but users can [configure service account](https://cloud.google.com/iam/docs/creating-short-lived-service-account-credentials#sa-credentials-oauth) to extend the maximum lifetime to 12 hours.

This PR allows users to set custom lifetime (<= 12 hours).


arithmetic1728:
The `format / format-code` failure is not caused by the PR.

silvolu (approved):

oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java
```diff
@@ -218,6 +226,9 @@ private ImpersonatedCredentials(Builder builder) {
     if (this.scopes == null) {
       throw new IllegalStateException(SCOPE_EMPTY_ERROR);
     }
+    if (this.lifetime > TWELVE_HOURS_IN_SECONDS) {
```
chingor13, line 229: While checking to see if this could possibly be null (it can't but could be `0`), it looks like we can possibly make a request for a token with `0s` for the lifetime value. Should it actually be unset and rely on the server default value?

chingor13 (approved):

oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java
chingor13: Filed https://github.com/googleapis/google-auth-library-java/issues/526

PR #518, opened 2020-12-11
feat: allow set timeout for gce

### LINKED ISSUE #517: The ComputeEngineCredentials auto-detection timeout is too tight
The autodetection logic in google-auth-library-java uses a 1.5 second timeout (3 tries at 0.5 seconds per try) before deciding that it isn't running on Compute Engine.

Unfortunately, 1.5 seconds is a little tight for GKE Workload Identity / gke-metadata-server, since it needs to wait for IP address information to bounce around between kube-apiserver and the node.

A timeout more like 10 seconds would go a long way towards making the autodetection more stable. Or, perhaps the client library could look for a "I'm definitely running on Compute Engine" environment variable.

arithmetic1728:
Fixes #517 

Allow the GKE users to explicitly set the metadata server timeout via `GCE_METADATA_TIMEOUT` environment variable (this env variable is used by google-auth-library-python as well).


oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -82,6 +82,10 @@
   // "unlikely" since the expected 4-nines time is about 0.5 seconds.
   // This allows us to limit the total ping maximum timeout to 1.5 seconds
   // for developer desktop scenarios.
+  // However, this workaround doesn't work for GKE because GKE metadata resolution takes
+  // much longer time than 1.5 seconds. We won't have a timeout that works for all scenarios,
+  // so GKE users should use the `GCE_METADATA_TIMEOUT` environment variable to explicitly set the
+  // timeout in seconds.
   static final int MAX_COMPUTE_PING_TRIES = 3;
   static final int COMPUTE_PING_CONNECTION_TIMEOUT_MS = 500;
```
chingor13, line 90: Can we prepend `DEFAULT_` to this constant name to reflect that it's now a default.

oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java
```diff
@@ -76,6 +76,9 @@
   static final String RUNTIME_JETTY_LOGGER = System.getProperty("org.eclipse.jetty.util.log.class");
   static final Logger LOGGER = Logger.getLogger(DefaultCredentialsProvider.class.getName());
 
+  /** Timeout in seconds to wait for the GCE metadata server when detecting the GCE environment. */
+  static final String GCE_METADATA_TIMEOUT_SECONDS_ENV_VAR = "GCE_METADATA_TIMEOUT";
```
chingor13, line 80: Let's confirm that this behavior and environment name will be consistent across all 7 auth libraries

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
arithmetic1728: done

oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java
arithmetic1728: This is the env variable name and behavior as python (https://github.com/googleapis/google-auth-library-python/blob/master/google/auth/compute_engine/_metadata.py#L58). The rest auth libs are still using a default value. The behavior is already not consistent for more than 4 years.

This is a cheap fix to unblock GKE users while we figure out if we can do a better fix, there will be some work across 7 auth libs, either this fix or some better fix.

silvolu (approved):

chingor13 (changes requested):
New environment variables need to be clearly documented across the languages. Can we get this in an AIP? go/std-client-env-vars

chingor13:
Also closing this one like .NET. We will revisit once the env variable is approved across the board.

PR #527, opened 2020-12-16
fix: use default timeout if given 0 for ImpersonatedCredentials

### LINKED ISSUE #526: ImpersonatedCredentials can request 0s lifetime
When going through the builder, it's possible for the `lifetime` value to be unset (defaulting to `0` in Java).

```java
ImpersonatedCredentials credentials = ImpersonatedCredentials.newBuilder()
    .setScopes(ImmutableList.of("some scope"))
    .build();
// This builds, but has a lifetime of `0`
```

When we go to [refresh the token](https://github.com/googleapis/google-auth-library-java/blob/edc8d6e0e7ca2c6749d026ba42854a09c4879fd6/oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java#L247), we don't check for 0. It's unclear what sending `0s` for the lifetime value will do on the backend. In the above case where you don't specify a lifetime, it seems that you would be requesting the default (whether client specified or backend specified).

arithmetic1728:
Fixes #526


oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java
```diff
@@ -229,6 +238,9 @@ private ImpersonatedCredentials(Builder builder) {
     if (this.lifetime > TWELVE_HOURS_IN_SECONDS) {
       throw new IllegalStateException(LIFETIME_EXCEEDED_ERROR);
     }
+    if (this.lifetime == 0) {
+      this.lifetime = DEFAULT_LIFETIME_IN_SECONDS;
```
chingor13, line 242: If we're doing this client side, perhaps it would be better to default this in the builder.

chingor13 (changes requested):
Would we ever want this to make a request to the token endpoint without the lifetime parameter at all?

oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java
arithmetic1728: moved to builder

arithmetic1728:
> Would we ever want this to make a request to the token endpoint without the lifetime parameter at all?

I think `lifetime` is required, it is part of json in the POST request sent to the token endpoint. Here is the [reference](https://cloud.google.com/iam/docs/creating-short-lived-service-account-credentials#sa-credentials-oauth).

chingor13 (approved):

PR #537, opened 2021-01-12
fix: remove deprecated methods

elharo:
This library is pre-1.0. If any dependencies need this, we should find out now so we can undeprecate these methods.


chingor13:
For context, I added these overrides because users were confused calling the static `getApplicationDefault()` on the `AppEngineCredentials` class. `AppEngineCredentials` didn't provide this method (it would fall back to use the superclass `GoogleCredentials`' implementaiton) so we added this deprecation warning to keep folks from using this as it will never return an instance of `AppEngineCredentials`.

chingor13 (approved):
Need to remove the direct dependency declaration on google-http-client for the appengine artifact.

PR #547, opened 2021-01-28
feat: add workload identity federation support

lsirac:
See go/guac-3pi-java.

- [x] Add documentation/snippets


elharo (changes requested):
This is a lot easier to review if you keep it on the same branch. Earlier comments apply.

oauth2_http/java/com/google/auth/oauth2/AwsCredentials.java
```diff
@@ -0,0 +1,348 @@
...
+  private String getAwsRegion() throws IOException {
+    // For AWS Lambda, the region is retrieved through the AWS_REGION environment variable.
+    String region = getEnv("AWS_REGION");
+    if (region != null) {
+      return region;
+    }
+
+    AwsCredentialSource awsCredentialSource = (AwsCredentialSource) this.credentialSource;
+    if (awsCredentialSource.regionUrl == null || awsCredentialSource.regionUrl.isEmpty()) {
+      throw new IOException(
```
elharo, line 248: IOException doesn't quite fit here. I'm not sure there's any standard exception that does, so perhaps a custom checked exception?

oauth2_http/java/com/google/auth/oauth2/AwsCredentials.java
```diff
@@ -0,0 +1,348 @@
...
+    // we want `us-east-1`.
+    return region.substring(0, region.length() - 1);
+  }
+
+  @VisibleForTesting
+  AwsSecurityCredentials getAwsSecurityCredentials() throws IOException {
+    // Check environment variables for credentials first.
+    String accessKeyId = getEnv("AWS_ACCESS_KEY_ID");
+    String secretAccessKey = getEnv("AWS_SECRET_ACCESS_KEY");
+    String token = getEnv("Token");
```
elharo, line 264: token looks like it might be null too

oauth2_http/java/com/google/auth/oauth2/AwsCredentials.java
```diff
@@ -0,0 +1,348 @@
...
+    secretAccessKey = (String) genericJson.get("SecretAccessKey");
+    token = (String) genericJson.get("Token");
+
+    // These credentials last for a few hours - we may consider caching these in the
+    // future.
+    return new AwsSecurityCredentials(accessKeyId, secretAccessKey, token);
+  }
+
+  @VisibleForTesting
+  String getEnv(String name) {
```
elharo, line 298: perhaps static

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,399 @@
...
+ * met:
+ *
+ *    * Redistributions of source code must retain the above copyright
+ * notice, this list of conditions and the following disclaimer.
+ *    * Redistributions in binary form must reproduce the above
+ * copyright notice, this list of conditions and the following disclaimer
+ * in the documentation and/or other materials provided with the
+ * distribution.
+ *
+ *    * Neither the name of Google Inc. nor the names of its
```
elharo, line 15: LLC. See https://g3doc.corp.google.com/company/teams/opensource/releasing/licenses.md?cl=head for current text

oauth2_http/java/com/google/auth/oauth2/AwsRequestSignature.java
```diff
@@ -0,0 +1,191 @@
...
+ * met:
+ *
+ *    * Redistributions of source code must retain the above copyright
+ * notice, this list of conditions and the following disclaimer.
+ *    * Redistributions in binary form must reproduce the above
+ * copyright notice, this list of conditions and the following disclaimer
+ * in the documentation and/or other materials provided with the
+ * distribution.
+ *
+ *    * Neither the name of Google Inc. nor the names of its
```
elharo, line 15: LLC. See https://g3doc.corp.google.com/company/teams/opensource/releasing/licenses.md?cl=head for current text

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,399 @@
...
+          dates);
+    }
+  }
+
+  static final class AwsDates {
+
+    private static final String X_AMZ_DATE_FORMAT = "yyyyMMdd'T'HHmmss'Z'";
+    private static final String CUSTOM_DATE_FORMAT = "E, dd MMM yyyy HH:mm:ss z";
+
+    private String originalDate;
```
elharo, line 346: final

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,399 @@
...
+    private String xAmzDate;
```
elharo, line 347: Is this mutable? If not, mark final though then I don't see why there are two variables

oauth2_http/java/com/google/auth/oauth2/AwsSecurityCredentials.java
```diff
@@ -0,0 +1,65 @@
...
+ * met:
+ *
+ *    * Redistributions of source code must retain the above copyright
+ * notice, this list of conditions and the following disclaimer.
+ *    * Redistributions in binary form must reproduce the above
+ * copyright notice, this list of conditions and the following disclaimer
+ * in the documentation and/or other materials provided with the
+ * distribution.
+ *
+ *    * Neither the name of Google Inc. nor the names of its
```
elharo, line 15: LLC

oauth2_http/java/com/google/auth/oauth2/AwsSecurityCredentials.java
```diff
@@ -0,0 +1,65 @@
...
+
+import javax.annotation.Nullable;
+
+/**
+ * Defines AWS security credentials. These are either retrieved from the AWS security_credentials
+ * endpoint or AWS environment variables.
+ */
+class AwsSecurityCredentials {
+
+  private String accessKeyId;
```
elharo, line 42: final

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -31,6 +31,8 @@
 
 package com.google.auth.oauth2;
 
+import static com.google.auth.oauth2.IdentityPoolCredentials.EXTERNAL_ACCOUNT_FILE_TYPE;
```
elharo, line 34: only use static imports for very well known classes. This isn't.

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
```diff
@@ -0,0 +1,313 @@
...
+    private CredentialFormatType credentialFormatType;
+    private String credentialLocation;
+
+    @Nullable private String subjectTokenFieldName;
+    @Nullable private Map<String, String> headers;
+
+    /**
+     * The source of the 3P credential.
+     *
+     * <p>If the this a file based 3P credential, the credentials file can be retrieved using the
```
elharo, line 93: "the this a"?

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
```diff
@@ -0,0 +1,313 @@
...
+     * `file` key.
+     *
+     * <p>If this is URL-based 3p credential, the metadata server URL can be retrieved using the
+     * `url` key.
+     *
+     * <p>The third party credential can be provided in different formats, such as text or JSON. The
+     * format can be specified using the `format` header, which will return a map with keys `type`
```
elharo, line 100: will return --> returns

(Google style prefers present tense)

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
```diff
@@ -0,0 +1,313 @@
...
+
+      if (credentialSourceMap.containsKey("file")) {
+        credentialLocation = (String) credentialSourceMap.get("file");
+        credentialSourceType = IdentityPoolCredentialSourceType.FILE;
+      } else if (credentialSourceMap.containsKey("url")) {
+        credentialLocation = (String) credentialSourceMap.get("url");
+        credentialSourceType = IdentityPoolCredentialSourceType.URL;
+      } else {
+        throw new IllegalArgumentException(
+            "Missing credential source file location or URL. At least one must be specified.");
```
elharo, line 117: It's OK to have both? and then file takes precedence?

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
```diff
@@ -0,0 +1,313 @@
...
+  }
+
+  private String parseToken(InputStream inputStream) throws IOException {
+    if (identityPoolCredentialSource.credentialFormatType == CredentialFormatType.TEXT) {
+      BufferedReader reader =
+          new BufferedReader(new InputStreamReader(inputStream, StandardCharsets.UTF_8));
+      return CharStreams.toString(reader);
+    }
+
+    JsonFactory jsonFactory = OAuth2Utils.JSON_FACTORY;
```
elharo, line 230: inline; local variable is simply an alias

oauth2_http/java/com/google/auth/oauth2/OAuthException.java
```diff
@@ -0,0 +1,79 @@
...
+ * met:
+ *
+ *    * Redistributions of source code must retain the above copyright
+ * notice, this list of conditions and the following disclaimer.
+ *    * Redistributions in binary form must reproduce the above
+ * copyright notice, this list of conditions and the following disclaimer
+ * in the documentation and/or other materials provided with the
+ * distribution.
+ *
+ *    * Neither the name of Google Inc. nor the names of its
```
elharo, line 15: LLC

oauth2_http/java/com/google/auth/oauth2/OAuthException.java
```diff
@@ -0,0 +1,79 @@
...
+
+package com.google.auth.oauth2;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import java.io.IOException;
+import javax.annotation.Nullable;
+
+class OAuthException extends IOException {
+  private static final String FULL_MESSAGE_FORMAT = "Error code %s: %s - %s";
```
elharo, line 40: These are only used in one place. Might as well inline them so the reader doesn't have to scroll for the definition

oauth2_http/java/com/google/auth/oauth2/OAuthException.java
```diff
@@ -0,0 +1,79 @@
...
+  public OAuthException(
+      String errorCode, @Nullable String errorDescription, @Nullable String errorUri) {
+    this.errorCode = checkNotNull(errorCode);
+    this.errorDescription = errorDescription;
+    this.errorUri = errorUri;
+  }
+
+  @Override
+  public String getMessage() {
+    if (errorDescription != null && errorUri != null) {
```
elharo, line 57: This would be clearer if you just built up the string, appending additional parts if they're non-null

oauth2_http/java/com/google/auth/oauth2/StsRequestHandler.java
```diff
@@ -0,0 +1,229 @@
...
+  }
+
+  public static Builder newBuilder(
+      String tokenExchangeEndpoint,
+      StsTokenExchangeRequest stsTokenExchangeRequest,
+      HttpRequestFactory httpRequestFactory) {
+    return new Builder(tokenExchangeEndpoint, stsTokenExchangeRequest, httpRequestFactory);
+  }
+
+  /** Exchanges the provided token for another type of token based on the rfc8693 spec. */
```
elharo, line 96: RFC 8693

oauth2_http/java/com/google/auth/oauth2/StsTokenExchangeResponse.java
```diff
@@ -0,0 +1,134 @@
...
+  private StsTokenExchangeResponse(
+      String accessToken,
+      String issuedTokenType,
+      String tokenType,
+      Long expiresIn,
+      @Nullable String refreshToken,
+      @Nullable List<String> scopes) {
+    checkNotNull(accessToken);
+    this.expiresIn = checkNotNull(expiresIn);
+    long expiresAtMilliseconds = Clock.SYSTEM.currentTimeMillis() + expiresIn * 1000L;
```
elharo, line 64: just use System.currentTimeMillis. Clock is gratuitous since it's not injected.

elharo, line 64: DANGER! POSSIBLE BUG. If expiresIn is not in seconds this calculation is wrong. If it's correct, it needs a comment explaining itself.

oauth2_http/java/com/google/auth/oauth2/StsTokenExchangeResponse.java
```diff
@@ -0,0 +1,134 @@
...
+  private Long expiresIn;
+
+  @Nullable private String refreshToken;
+  @Nullable private List<String> scopes;
+
+  private StsTokenExchangeResponse(
+      String accessToken,
+      String issuedTokenType,
+      String tokenType,
+      Long expiresIn,
```
elharo, line 59: why is this a Long instead of a long? you don't check for null

oauth2_http/javatests/com/google/auth/oauth2/ExternalAccountCredentialsTest.java
```diff
@@ -0,0 +1,346 @@
+/*
+ * Copyright 2020, Google LLC
```
elharo, line 2: 2021

oauth2_http/javatests/com/google/auth/oauth2/ExternalAccountCredentialsTest.java
```diff
@@ -0,0 +1,346 @@
...
+ * met:
+ *
+ *    * Redistributions of source code must retain the above copyright
+ * notice, this list of conditions and the following disclaimer.
+ *    * Redistributions in binary form must reproduce the above
+ * copyright notice, this list of conditions and the following disclaimer
+ * in the documentation and/or other materials provided with the
+ * distribution.
+ *
+ *    * Neither the name of Google Inc. nor the names of its
```
elharo, line 15: LLC

oauth2_http/javatests/com/google/auth/oauth2/ExternalAccountCredentialsTest.java
```diff
@@ -0,0 +1,346 @@
...
+    assertTrue(credential instanceof AwsCredentials);
+  }
+
+  @Test
+  public void fromStream_nullTransport_throws() {
+    try {
+      ExternalAccountCredentials.fromStream(
+          new ByteArrayInputStream("foo".getBytes()), /* transportFactory= */ null);
+      fail("Exception should be thrown.");
+    } catch (NullPointerException | IOException e) {
```
elharo, line 110: In a test this should be one or the other, not both

oauth2_http/javatests/com/google/auth/oauth2/ExternalAccountCredentialsTest.java
```diff
@@ -0,0 +1,346 @@
...
+    }
+  }
+
+  @Test
+  public void fromStream_nullStream_throws() {
+    try {
+      ExternalAccountCredentials.fromStream(
+          /* credentialsStream= */ null, OAuth2Utils.HTTP_TRANSPORT_FACTORY);
+      fail("Exception should be thrown.");
+    } catch (NullPointerException | IOException e) {
```
elharo, line 121: In a test this should be one or the other, whichever is expected, not both

elharo (commented):
Merge master

oauth2_http/java/com/google/auth/oauth2/AwsCredentials.java
lsirac: Yep - token isn't required

oauth2_http/java/com/google/auth/oauth2/AwsCredentials.java
lsirac: Can't - needs to be overriden in unit tests

oauth2_http/java/com/google/auth/oauth2/StsTokenExchangeResponse.java
lsirac: I am checking for null

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
lsirac: oops

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
lsirac: Marked final. The date is needed in both formats.

oauth2_http/java/com/google/auth/oauth2/AwsCredentials.java
lsirac: It kind of fits.. but if you want a custom checked exception, it's going to have to extend from IOException anyway, since this is called in retrieveSubjectToken. Any ideas?

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
lsirac: Yes

oauth2_http/java/com/google/auth/oauth2/StsTokenExchangeResponse.java
lsirac: STS returns expiresIn in seconds. What would you like me to do here?

oauth2_http/java/com/google/auth/oauth2/AwsCredentials.java
elharo: OK, this will do then.

elharo (commented):
merge in master

oauth2_http/java/com/google/auth/oauth2/AwsCredentials.java
```diff
@@ -0,0 +1,348 @@
...
+    GenericJson token = new GenericJson();
+    token.setFactory(OAuth2Utils.JSON_FACTORY);
+
+    token.put("headers", headerList);
+    token.put("method", signature.getHttpMethod());
+    token.put(
+        "url",
+        ((AwsCredentialSource) credentialSource)
+            .regionalCredentialVerificationUrl.replace("{region}", signature.getRegion()));
+    return URLEncoder.encode(token.toString(), "UTF-8");
```
elharo, line 236: the exception really shouldn't happen here unless the VM is broken. Ine if the few cases whee I'm OK converting to a `new RuntimeException("this can't happen", e)`

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,399 @@
...
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package com.google.auth.oauth2;
+
+import static com.google.api.client.util.Preconditions.checkNotNull;
```
elharo, line 34: use Guava or just do it directly. We want to get rid of these utility methods.

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,399 @@
...
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+import com.google.api.client.util.Joiner;
```
elharo, line 37: use guava

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,399 @@
...
+import java.util.TimeZone;
+import javax.annotation.Nullable;
+import javax.crypto.Mac;
+import javax.crypto.spec.SecretKeySpec;
+
+/**
+ * Internal utility that signs AWS API requests based on the AWS Signature Version 4 signing
+ * process.
+ *
+ * <p>https://docs.aws.amazon.com/general/latest/gr/signature-version-4.html
```
elharo, line 64: make this a link above

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,399 @@
...
+  private static byte[] sign(byte[] key, byte[] value) {
+    try {
+      String algorithm = "HmacSHA256";
+      Mac mac = Mac.getInstance(algorithm);
+      mac.init(new SecretKeySpec(key, algorithm));
+      return mac.doFinal(value);
+    } catch (NoSuchAlgorithmException e) {
+      // Will not occur as HmacSHA256 is supported. We may allow other algorithms in the future.
+      throw new RuntimeException(
+          "Invalid algorithm used when calculating the AWS V4 Signature.", e);
```
elharo, line 263: Is the algorithm invalid? The comment suggests the VM doesn't follow the spec.

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
```diff
@@ -0,0 +1,459 @@
...
+ *
+ * <p>Handles initializing third-party credentials, calls to STS and service account impersonation.
+ */
+public abstract class ExternalAccountCredentials extends GoogleCredentials
+    implements QuotaProjectIdProvider {
+
+  /** Base credential source class. Dictates the retrieval method of the 3PI credential. */
+  abstract static class CredentialSource {
+
+    protected Map<String, Object> credentialSourceMap;
```
elharo, line 64: iffy that this is not private

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
```diff
@@ -0,0 +1,459 @@
...
+      "https://www.googleapis.com/auth/cloud-platform";
+
+  static final String EXTERNAL_ACCOUNT_FILE_TYPE = "external_account";
+
+  protected final String transportFactoryClassName;
+  protected final String audience;
+  protected final String subjectTokenType;
+  protected final String tokenUrl;
+  protected final CredentialSource credentialSource;
+  protected final Collection<String> scopes;
```
elharo, line 81: private since it's mutable?

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
```diff
@@ -0,0 +1,459 @@
...
+  protected transient HttpTransportFactory transportFactory;
+
+  @Nullable protected final ImpersonatedCredentials impersonatedCredentials;
+
+  /**
+   * Constructor with minimum identifying information and custom HTTP transport.
+   *
+   * @param transportFactory HTTP transport factory, creates the transport used to get access
+   *     tokens.
+   * @param audience The STS audience which is usually the fully specified resource name of the
```
elharo, line 98: The --> the 
per Oracle JavaDoc guidelines

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
```diff
@@ -0,0 +1,459 @@
...
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package com.google.auth.oauth2;
+
+import static com.google.api.client.util.Preconditions.checkNotNull;
```
elharo, line 34: guava or just do it manually

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
```diff
@@ -0,0 +1,459 @@
...
+    Map<String, List<String>> requestMetadata = super.getRequestMetadata(uri);
+    return addQuotaProjectIdToRequestMetadata(quotaProjectId, requestMetadata);
+  }
+
+  /**
+   * Returns credentials defined by a JSON file stream.
+   *
+   * <p>This will either return {@link IdentityPoolCredentials} or AwsCredentials.
+   *
+   * @param credentialsStream the stream with the credential definition.
```
elharo, line 185: no period unless it's a complete sentence

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
```diff
@@ -0,0 +1,459 @@
...
+    JsonObjectParser parser = new JsonObjectParser(OAuth2Utils.JSON_FACTORY);
+    GenericJson fileContents =
+        parser.parseAndClose(credentialsStream, StandardCharsets.UTF_8, GenericJson.class);
+    return fromJson(fileContents, transportFactory);
+  }
+
+  /**
+   * Returns external account credentials defined by JSON using the format generated by gCloud.
+   *
+   * @param json a map from the JSON representing the credentials.
```
elharo, line 219: no period

oauth2_http/java/com/google/auth/oauth2/StsRequestHandler.java
```diff
@@ -0,0 +1,229 @@
...
+    if (headers != null) {
+      httpRequest.setHeaders(headers);
+    }
+
+    try {
+      HttpResponse response = httpRequest.execute();
+      GenericData responseData = response.parseAs(GenericData.class);
+      return buildResponse(responseData);
+    } catch (IOException e) {
+      if (!(e instanceof HttpResponseException)) {
```
elharo, line 112: instead of instanceof it's clearer to use two catch blocks here: first HttpResponseException, then IOException

silvolu (approved):
Auth bits look good, please resolve comments from @elharo

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
lsirac: Fair, changed the message.

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
lsirac: what would you prefer?

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
lsirac: what do you mean? these are accessed in subclasses

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
```diff
@@ -0,0 +1,455 @@
...
+  @Override
+  public Map<String, List<String>> getRequestMetadata(URI uri) throws IOException {
+    Map<String, List<String>> requestMetadata = super.getRequestMetadata(uri);
+    return addQuotaProjectIdToRequestMetadata(quotaProjectId, requestMetadata);
+  }
+
+  /**
+   * Returns credentials defined by a JSON file stream.
+   *
+   * <p>This will either return {@link IdentityPoolCredentials} or AwsCredentials.
```
elharo, line 180: will return --> returns
present tense, per Google style

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
```diff
@@ -0,0 +1,455 @@
...
+   */
+  public static ExternalAccountCredentials fromStream(InputStream credentialsStream)
+      throws IOException {
+    return fromStream(credentialsStream, OAuth2Utils.HTTP_TRANSPORT_FACTORY);
+  }
+
+  /**
+   * Returns credentials defined by a JSON file stream.
+   *
+   * <p>This will either return a IdentityPoolCredentials or AwsCredentials.
```
elharo, line 194: returns

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
```diff
@@ -0,0 +1,455 @@
...
+   * @param json a map from the JSON representing the credentials
+   * @param transportFactory HTTP transport factory, creates the transport used to get access tokens
+   * @return the credentials defined by the JSON
+   */
+  public static ExternalAccountCredentials fromJson(
+      Map<String, Object> json, HttpTransportFactory transportFactory) {
+    checkNotNull(json);
+    checkNotNull(transportFactory);
+
+    String audience = (String) json.get("audience");
```
elharo, line 225: There are a lot of potential class cast exceptions here, especially given that this is a public method. Does it have to be public?

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
```diff
@@ -0,0 +1,455 @@
...
+      return this;
+    }
+
+    public Builder setCredentialSource(CredentialSource credentialSource) {
+      this.credentialSource = credentialSource;
+      return this;
+    }
+
+    public Builder setScopes(Collection<String> scopes) {
+      this.scopes = scopes;
```
elharo, line 429: test

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
```diff
@@ -0,0 +1,311 @@
...
+     *
+     * <p>Optional headers can be present, and should be keyed by `headers`.
+     */
+    public IdentityPoolCredentialSource(Map<String, Object> credentialSourceMap) {
+      super(credentialSourceMap);
+
+      if (credentialSourceMap.containsKey("file")) {
+        credentialLocation = (String) credentialSourceMap.get("file");
+        credentialSourceType = IdentityPoolCredentialSourceType.FILE;
+      } else if (credentialSourceMap.containsKey("url")) {
```
elharo, line 111: What if it contains both?

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
```diff
@@ -0,0 +1,311 @@
...
+     * `subject_token_field_name`. If the `type` is json, the `subject_token_field_name` must be
+     * provided. If no format is provided, we expect the token to be in the raw text format.
+     *
+     * <p>Optional headers can be present, and should be keyed by `headers`.
+     */
+    public IdentityPoolCredentialSource(Map<String, Object> credentialSourceMap) {
+      super(credentialSourceMap);
+
+      if (credentialSourceMap.containsKey("file")) {
+        credentialLocation = (String) credentialSourceMap.get("file");
```
elharo, line 109: again, a class cast exception is possible in a public method.

oauth2_http/java/com/google/auth/oauth2/OAuthException.java
```diff
@@ -0,0 +1,82 @@
...
+ * Encapsulates the standard OAuth error response. See
+ * https://tools.ietf.org/html/rfc6749#section-5.2.
+ */
+class OAuthException extends IOException {
+
+  private final String errorCode;
+  @Nullable private final String errorDescription;
+  @Nullable private final String errorUri;
+
+  public OAuthException(
```
elharo, line 49: not public

oauth2_http/java/com/google/auth/oauth2/StsRequestHandler.java
```diff
@@ -0,0 +1,226 @@
...
+import com.google.api.client.http.HttpRequest;
+import com.google.api.client.http.HttpRequestFactory;
+import com.google.api.client.http.HttpResponse;
+import com.google.api.client.http.HttpResponseException;
+import com.google.api.client.http.UrlEncodedContent;
+import com.google.api.client.json.GenericJson;
+import com.google.api.client.json.JsonObjectParser;
+import com.google.api.client.json.JsonParser;
+import com.google.api.client.util.GenericData;
+import com.google.api.client.util.Joiner;
```
elharo, line 45: use Guava Joiner instead as it's better maintained and tested

elharo (commented):
merge in master

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
lsirac: you asked this last time - it will default to file

elharo: yes, but does that make sense? What does the caller want if they supply both? Is it an error that should throw an exception?

lsirac: Added an exception

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
lsirac: shouldn't be public. made package private

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
lsirac: This should be package private

oauth2_http/java/com/google/auth/oauth2/StsTokenExchangeResponse.java
```diff
@@ -0,0 +1,136 @@
...
+      this.expiresIn = expiresIn;
+    }
+
+    public StsTokenExchangeResponse.Builder setRefreshToken(String refreshToken) {
+      this.refreshToken = refreshToken;
+      return this;
+    }
+
+    public StsTokenExchangeResponse.Builder setScopes(List<String> scopes) {
+      this.scopes = scopes;
```
elharo, line 127: need a defensive copy here to avoid exposing mutable internal state

oauth2_http/java/com/google/auth/oauth2/StsTokenExchangeResponse.java
```diff
@@ -0,0 +1,136 @@
...
+  private final Long expiresIn;
+
+  @Nullable private final String refreshToken;
+  @Nullable private final List<String> scopes;
+
+  private StsTokenExchangeResponse(
+      String accessToken,
+      String issuedTokenType,
+      String tokenType,
+      Long expiresIn,
```
elharo, line 58: expiresInSeconds to avoid unit confusion

oauth2_http/javatests/com/google/auth/oauth2/ExternalAccountCredentialsTest.java
```diff
@@ -0,0 +1,346 @@
...
+
+  @Before
+  public void setup() {
+    transportFactory = new MockExternalAccountCredentialsTransportFactory();
+  }
+
+  @Test
+  public void fromStream_identityPoolCredentials() throws IOException {
+    GenericJson json = buildJsonIdentityPoolCredential();
+    TestUtils.jsonToInputStream(json);
```
elharo, line 85: This modifies its argument? That's a weird design, but does not need to be fixed in this PR.

oauth2_http/javatests/com/google/auth/oauth2/ExternalAccountCredentialsTest.java
```diff
@@ -0,0 +1,346 @@
...
+    AccessToken returnedToken =
+        credential.exchangeExternalCredentialForAccessToken(stsTokenExchangeRequest);
+
+    assertEquals(
+        transportFactory.transport.getServiceAccountAccessToken(), returnedToken.getTokenValue());
+  }
+
+  @Test
+  public void exchangeExternalCredentialForAccessToken_throws() throws IOException {
+    final ExternalAccountCredentials credential =
```
elharo, line 229: Google style does not use final on local variables except where required.

oauth2_http/javatests/com/google/auth/oauth2/ExternalAccountCredentialsTest.java
```diff
@@ -0,0 +1,346 @@
...
+        ExternalAccountCredentials.fromJson(buildJsonIdentityPoolCredential(), transportFactory);
+
+    String errorCode = "invalidRequest";
+    String errorDescription = "errorDescription";
+    String errorUri = "errorUri";
+    transportFactory.transport.addResponseErrorSequence(
+        TestUtils.buildHttpResponseException(errorCode, errorDescription, errorUri));
+
+    final StsTokenExchangeRequest stsTokenExchangeRequest =
```
elharo, line 238: Google style does not use final on local variables except where required.

oauth2_http/javatests/com/google/auth/oauth2/IdentityPoolCredentialsTest.java
```diff
@@ -0,0 +1,467 @@
...
+
+  @Test
+  public void retrieveSubjectToken_urlSourcedCredential_throws() {
+    MockExternalAccountCredentialsTransportFactory transportFactory =
+        new MockExternalAccountCredentialsTransportFactory();
+
+    IOException response = new IOException();
+    transportFactory.transport.addResponseErrorSequence(response);
+
+    final IdentityPoolCredentials credential =
```
elharo, line 269: Google style does not use final on local variables except where required.

oauth2_http/javatests/com/google/auth/oauth2/IdentityPoolCredentialsTest.java
```diff
@@ -0,0 +1,467 @@
...
+          e.getMessage());
+    }
+  }
+
+  @Test
+  public void refreshAccessToken_withoutServiceAccountImpersonation() throws IOException {
+    MockExternalAccountCredentialsTransportFactory transportFactory =
+        new MockExternalAccountCredentialsTransportFactory();
+
+    final IdentityPoolCredentials credential =
```
elharo, line 293: Google style does not use final on local variables except where required.

oauth2_http/javatests/com/google/auth/oauth2/StsRequestHandlerTest.java
```diff
@@ -0,0 +1,252 @@
...
+    assertNull(e.getErrorDescription());
+    assertNull(e.getErrorUri());
+  }
+
+  @Test
+  public void exchangeToken_withOptionalParams_throwsException() throws IOException {
+    StsTokenExchangeRequest stsTokenExchangeRequest =
+        StsTokenExchangeRequest.newBuilder("credential", "subjectTokenType").build();
+
+    final StsRequestHandler requestHandler =
```
elharo, line 205: Google style does not use final on local variables except where required.

oauth2_http/java/com/google/auth/oauth2/AwsCredentials.java
```diff
@@ -0,0 +1,349 @@
...
+    return new AwsCredentials.Builder();
+  }
+
+  public static AwsCredentials.Builder newBuilder(AwsCredentials awsCredentials) {
+    return new AwsCredentials.Builder(awsCredentials);
+  }
+
+  public static class Builder extends ExternalAccountCredentials.Builder {
+
+    protected Builder() {}
```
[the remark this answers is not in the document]
lsirac: should be package private

oauth2_http/java/com/google/auth/oauth2/AwsCredentials.java
```diff
@@ -0,0 +1,349 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import javax.annotation.Nullable;
+
+/**
+ * AWS credentials representing a third-party identity for calling Google APIs.
+ *
+ * <p>By default, attempts to exchange the external credential for a GCP access token.
+ */
+public class AwsCredentials extends ExternalAccountCredentials {
```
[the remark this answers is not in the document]
lsirac: no

oauth2_http/java/com/google/auth/oauth2/AwsCredentials.java
```diff
@@ -0,0 +1,349 @@
...
+
+  public static AwsCredentials.Builder newBuilder() {
+    return new AwsCredentials.Builder();
+  }
+
+  public static AwsCredentials.Builder newBuilder(AwsCredentials awsCredentials) {
+    return new AwsCredentials.Builder(awsCredentials);
+  }
+
+  public static class Builder extends ExternalAccountCredentials.Builder {
```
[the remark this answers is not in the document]
lsirac: would break tests

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
```diff
@@ -0,0 +1,318 @@
...
+
+/**
+ * Url-sourced and file-sourced external account credentials.
+ *
+ * <p>By default, attempts to exchange the third-party credential for a GCP access token.
+ */
+public class IdentityPoolCredentials extends ExternalAccountCredentials {
+
+  /**
+   * The IdentityPool credential source. Dictates the retrieval method of the external credential,
```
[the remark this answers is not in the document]
lsirac: This class holds the information we need to retrieve the external credential. I don't quite understand why this wouldn't need to be stored?

oauth2_http/java/com/google/auth/oauth2/StsRequestHandler.java
```diff
@@ -0,0 +1,226 @@
...
+import com.google.api.client.util.GenericData;
+import com.google.common.base.Joiner;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import javax.annotation.Nullable;
+
+/** Implements the OAuth 2.0 token exchange based on https://tools.ietf.org/html/rfc8693. */
+public class StsRequestHandler {
```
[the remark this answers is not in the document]
lsirac: yes - made final

oauth2_http/java/com/google/auth/oauth2/StsTokenExchangeRequest.java
```diff
@@ -0,0 +1,202 @@
...
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import java.util.List;
+import javax.annotation.Nullable;
+
+/**
+ * Defines an OAuth 2.0 token exchange request. Based on
+ * https://tools.ietf.org/html/rfc8693#section-2.1.
+ */
+public class StsTokenExchangeRequest {
```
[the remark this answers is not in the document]
lsirac: made final

oauth2_http/java/com/google/auth/oauth2/StsTokenExchangeResponse.java
```diff
@@ -0,0 +1,136 @@
...
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.List;
+import javax.annotation.Nullable;
+
+/**
+ * Defines an OAuth 2.0 token exchange successful response. Based on
+ * https://tools.ietf.org/html/rfc8693#section-2.2.1.
+ */
+public class StsTokenExchangeResponse {
```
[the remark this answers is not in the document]
lsirac: made final

oauth2_http/javatests/com/google/auth/oauth2/ExternalAccountCredentialsTest.java
lsirac: actually can just be deleted. duplicated code

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
```diff
@@ -0,0 +1,318 @@
...
+    if (scopes != null && !scopes.isEmpty()) {
+      stsTokenExchangeRequest.setScopes(new ArrayList<>(scopes));
+    }
+
+    return exchangeExternalCredentialForAccessToken(stsTokenExchangeRequest.build());
+  }
+
+  @Override
+  public String retrieveSubjectToken() throws IOException {
+    if (identityPoolCredentialSource.credentialSourceType
```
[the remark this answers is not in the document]
lsirac: I'd prefer to not make that kind of refactor at this point, and I think this approach serves its purpose. We've already aligned across platforms, this was part of the design that was approved, etc. We can always refactor this in the future.

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
```diff
@@ -0,0 +1,318 @@
...
+    try {
+      return parseToken(new FileInputStream(new File(credentialFilePath)));
+    } catch (IOException e) {
+      throw new IOException(
+          "Error when attempting to read the subject token from the credential file.", e);
+    }
+  }
+
+  private String parseToken(InputStream inputStream) throws IOException {
+    if (identityPoolCredentialSource.credentialFormatType == CredentialFormatType.TEXT) {
```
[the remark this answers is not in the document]
lsirac: See above

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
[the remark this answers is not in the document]
elharo: It's easier now than it will be in the future. I don't expect the subclasses have to be public. It would be a much cleaner solution.

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
```diff
@@ -0,0 +1,318 @@
...
+      return parseToken(response.getContent());
+    } catch (IOException e) {
+      throw new IOException(
+          String.format("Error getting subject token from metadata server: %s", e.getMessage()), e);
+    }
+  }
+
+  /** Clones the IdentityPoolCredentials with the specified scopes. */
+  @Override
+  public GoogleCredentials createScoped(Collection<String> newScopes) {
```
elharo, line 271: I think this can be declared to return IdentityPoolCredentials thanks to covariant return types

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
```diff
@@ -0,0 +1,318 @@
...
+    return new Builder();
+  }
+
+  public static Builder newBuilder(IdentityPoolCredentials identityPoolCredentials) {
+    return new Builder(identityPoolCredentials);
+  }
+
+  public static class Builder extends ExternalAccountCredentials.Builder {
+
+    protected Builder() {}
```
elharo, line 296: These could probably be private

elharo (commented):
worth considering if any of the public classes and methods should be marked `@Beta`. In the near future it's going to be getting a lot harder to change something once we've shipped it than it has been in the past.

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
lsirac: made package private

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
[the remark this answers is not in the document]
lsirac: I think the overhead of doing it now or in the future is the same, but I do agree that it's a lot less likely to be done in the future if it's not done now. That being said, there's alignment across platforms, etc. I'd be happy to revisit this with you in the future.

oauth2_http/java/com/google/auth/oauth2/AwsCredentials.java
elharo, line 59: does this class need to be public? It looks like the only public methods are overridden or visible for testing, so I'm guessing no. The less API we have to publish and support ever after the better.

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,399 @@
...
+          httpMethod,
+          url,
+          region,
+          requestPayload,
+          additionalHeaders,
+          dates);
+    }
+  }
+
+  static final class AwsDates {
```
elharo, line 341: This feels like it should be an  independent outer class. Is there some reason it needs to be an inner class?

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
```diff
@@ -0,0 +1,455 @@
...
+/**
+ * Base external account credentials class.
+ *
+ * <p>Handles initializing external credentials, calls to STS, and service account impersonation.
+ */
+public abstract class ExternalAccountCredentials extends GoogleCredentials
+    implements QuotaProjectIdProvider {
+
+  /** Base credential source class. Dictates the retrieval method of the external credential. */
+  abstract static class CredentialSource {
```
elharo, line 62: could/should this be an outer class? ExternalAccountCredentials is so large github won't even display it by default, which suggests it might help to break it up.

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
```diff
@@ -0,0 +1,455 @@
...
+
+    protected CredentialSource(Map<String, Object> credentialSourceMap) {
```
elharo, line 64: do you anticipate subclasses outside this package? If not, make it package access instead of protected.

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
```diff
@@ -0,0 +1,455 @@
...
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import javax.annotation.Nullable;
+
+/**
+ * Base external account credentials class.
+ *
+ * <p>Handles initializing external credentials, calls to STS, and service account impersonation.
```
elharo, line 56: Is STS amazon specific? I'm wondering how well this class fits non-AWS 3rd party credentials.  I should take another look at the design doc.

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
elharo: can fromStream be package private too? Otherwise the same issues with class cast exceptions are just on method removed.

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
```diff
@@ -0,0 +1,455 @@
...
+        serviceAccountImpersonationUrl,
+        quotaProjectId,
+        clientId,
+        clientSecret,
+        /* scopes= */ null);
+  }
+
+  private static boolean isAwsCredential(Map<String, Object> credentialSource) {
+    return credentialSource.containsKey("environment_id")
+        && ((String) credentialSource.get("environment_id")).startsWith("aws");
```
elharo, line 271: possible class cast exception here? I'm not sure, but if so you could catch and return false.

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
```diff
@@ -0,0 +1,455 @@
...
+        StsRequestHandler.newBuilder(
+                tokenUrl, stsTokenExchangeRequest, transportFactory.create().createRequestFactory())
+            .build();
+
+    StsTokenExchangeResponse response = requestHandler.exchangeToken();
+    return response.getAccessToken();
+  }
+
+  private static String extractTargetPrincipal(String serviceAccountImpersonationUrl) {
+    // Extract the target principle.
```
elharo, line 298: principle or principal? I can never remember which is which, but either the method name or the comment is incorrect.

oauth2_http/java/com/google/auth/oauth2/StsRequestHandler.java
```diff
@@ -0,0 +1,226 @@
...
+import com.google.api.client.util.GenericData;
+import com.google.common.base.Joiner;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import javax.annotation.Nullable;
+
+/** Implements the OAuth 2.0 token exchange based on https://tools.ietf.org/html/rfc8693. */
+public final class StsRequestHandler {
```
elharo, line 53: any chance this can be non-public?

oauth2_http/java/com/google/auth/oauth2/StsTokenExchangeRequest.java
```diff
@@ -0,0 +1,202 @@
...
+          audience,
+          requestedTokenType);
+    }
+  }
+
+  static class ActingParty {
+    private final String actorToken;
+    private final String actorTokenType;
+
+    public ActingParty(String actorToken, String actorTokenType) {
```
elharo, line 189: These methods don't need to be public

oauth2_http/java/com/google/auth/oauth2/StsTokenExchangeRequest.java
```diff
@@ -0,0 +1,202 @@
...
+          subjectTokenType,
+          actingParty,
+          scopes,
+          resource,
+          audience,
+          requestedTokenType);
+    }
+  }
+
+  static class ActingParty {
```
elharo, line 185: outer class

oauth2_http/java/com/google/auth/oauth2/StsTokenExchangeRequest.java
```diff
@@ -0,0 +1,202 @@
...
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import java.util.List;
+import javax.annotation.Nullable;
+
+/**
+ * Defines an OAuth 2.0 token exchange request. Based on
+ * https://tools.ietf.org/html/rfc8693#section-2.1.
+ */
+public final class StsTokenExchangeRequest {
```
elharo, line 43: Can this be non-pubic?

oauth2_http/java/com/google/auth/oauth2/StsTokenExchangeResponse.java
```diff
@@ -0,0 +1,139 @@
...
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.List;
+import javax.annotation.Nullable;
+
+/**
+ * Defines an OAuth 2.0 token exchange successful response. Based on
+ * https://tools.ietf.org/html/rfc8693#section-2.2.1.
+ */
+public final class StsTokenExchangeResponse {
```
elharo, line 45: Can this be non-public?

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
lsirac: STS is external credential specific. All external credentials will be exchanged for a GCP access token through STS.

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
lsirac: I don't think this makes sense as an external class. It's required for all external credentials and not used in any other context. It's also confusing if you come across it outside of this class.

oauth2_http/java/com/google/auth/oauth2/AwsCredentials.java
lsirac: I'd prefer to keep this public. Developers can initialize an AwsCredential and use it directly, and other credentials follow that pattern. They also need to be able to create an AwsCredential with specific scopes.

oauth2_http/java/com/google/auth/oauth2/AwsRequestSigner.java
```diff
@@ -0,0 +1,399 @@
...
+
+    private static final String X_AMZ_DATE_FORMAT = "yyyyMMdd'T'HHmmss'Z'";
+    private static final String CUSTOM_DATE_FORMAT = "E, dd MMM yyyy HH:mm:ss z";
```
elharo, line 344: why *CUSTOM*_DATE_FORMAT? Is this the HTTP-date format defined by RFC 7231? If so, rename to HTTP_DATE_FORMAT for clarity.

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
```diff
@@ -0,0 +1,318 @@
...
+        headers.putAll(headersMap);
+      }
+
+      // If the format is not provided, we expect the token to be in the raw text format.
+      credentialFormatType = CredentialFormatType.TEXT;
+
+      Map<String, String> formatMap = (Map<String, String>) credentialSourceMap.get("format");
+      if (formatMap != null && formatMap.containsKey("type")) {
+        String type = formatMap.get("type");
+        if (type == null || (!type.equals("text") && !type.equals("json"))) {
```
elharo, line 136: simpler:

```
 if (!"text".equals(type) && !"json".equals(type)) {
```

oauth2_http/java/com/google/auth/oauth2/OAuthException.java
```diff
@@ -0,0 +1,81 @@
...
+    if (errorDescription != null) {
+      sb.append(": ").append(errorDescription);
+    }
+    if (errorUri != null) {
+      sb.append(" - ").append(errorUri);
+    }
+    return sb.toString();
+  }
+
+  public String getErrorCode() {
```
elharo, line 68: since this class is not public, it doesn't need public methods

oauth2_http/java/com/google/auth/oauth2/AwsCredentials.java
```diff
@@ -0,0 +1,349 @@
...
+        throw new IllegalArgumentException(
+            "A regional_cred_verification_url representing the"
+                + " GetCallerIdentity action URL must be specified.");
+      }
+
+      String environmentId = (String) credentialSourceMap.get("environment_id");
+
+      // Environment version is prefixed by "aws". e.g. "aws1".
+      Matcher matcher = Pattern.compile("^(aws)([\\d]+)$").matcher(environmentId);
+      if (!matcher.find()) {
```
elharo, line 97: Is this a multiline string? If not, you probably just want matches() here instead of find(), and you don't need to bind to the beginning and end of the line.

oauth2_http/java/com/google/auth/oauth2/AwsCredentials.java
```diff
@@ -0,0 +1,349 @@
...
+        throw new IllegalArgumentException("Invalid AWS environment ID.");
+      }
+
+      int environmentVersion = Integer.parseInt(matcher.group(2));
```
elharo, line 101: more simply, this looks like the string simply equals the literal "aws1"

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
```diff
@@ -0,0 +1,459 @@
...
+    String audience = (String) json.get("audience");
+    String subjectTokenType = (String) json.get("subject_token_type");
+    String tokenUrl = (String) json.get("token_url");
+    String serviceAccountImpersonationUrl = (String) json.get("service_account_impersonation_url");
+
+    Map<String, Object> credentialSourceMap = (Map<String, Object>) json.get("credential_source");
+
+    // Optional params.
+    String tokenInfoUrl =
+        json.containsKey("token_info_url") ? (String) json.get("token_info_url") : null;
```
elharo, line 238: shouldn't json.get simply return null if there's no such key? Are these containsKey calls necessary?

If they are necessary (because this map doesn't allow null keys) then lins 229-232 need these checks too.

lsirac: removed the containsKey

oauth2_http/java/com/google/auth/oauth2/AwsCredentials.java
lsirac: hmm we have to parse the version anyway for the error message

elharo (approved):

chingor13 (approved):

PR #558, opened 2021-02-04
fix: use covariant return type

### LINKED ISSUE #555: ServiceAccountCredentials.createScoped should return ServiceAccountCredentials
https://github.com/googleapis/google-auth-library-java/blob/master/oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java#L516

method should return ServiceAccountCredentials so callers do not need to cast.

elharo:
fixes #555 
@osi @silvolu @lukesneeringer


chingor13:
Clirr is catching this change as binary compatibility breaking. Source compatibility-wise it passes

PR #601, opened 2021-03-18
fix: fix service account credentials createScopedRequired

arithmetic1728:
Fixes: https://github.com/GoogleCloudPlatform/java-docs-samples/pull/4907#issuecomment-802096084

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
@@ -561,6 +561,12 @@ public class ServiceAccountCredentials extends GoogleCredentials
             fileType, SERVICE_ACCOUNT_FILE_TYPE));
   }
 
+  /** Returns whether the scopes are empty, meaning createScoped must be called before use. */
+  @Override
+  public boolean createScopedRequired() {
+    return scopes.isEmpty();
+  }
+
   /**
    * Refreshes the OAuth2 access token by getting a new access token using a JSON Web Token (JWT).
    */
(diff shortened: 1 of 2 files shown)


oauth2_http/javatests/com/google/auth/oauth2/ServiceAccountCredentialsTest.java
```diff
@@ -479,12 +479,12 @@ public void createdScoped_defaultScopes() throws IOException {
   }
 
   @Test
-  public void createScopedRequired_emptyScopes_false() throws IOException {
+  public void createScopedRequired_emptyScopes_true() throws IOException {
```
elharo, line 482: just createScopedRequired_emptyScopes

arithmetic1728, line 482:
```suggestion
  public void createScopedRequired_emptyScopes() throws IOException {
```

arithmetic1728: done

lesv (approved):

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -561,6 +561,12 @@ public static ServiceAccountCredentials fromStream(
             fileType, SERVICE_ACCOUNT_FILE_TYPE));
   }
 
+  /** Returns whether the scopes are empty, meaning createScoped must be called before use. */
+  @Override
+  public boolean createScopedRequired() {
+    return scopes.isEmpty();
```
bshaffer, line 567: I think without having `defaultScopes.isEmpty()` here as well, the [Client Library for Storage](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/storage/json-api/src/main/java/StorageFactory.java#L50) will never use self-signed JWTs. And there may be more APIs like this.

This can be fixed in a future PR, however.

PR #642, opened 2021-04-26
fix: use orginal url as audience for self signed jwt if scheme or host is null

arithmetic1728:
`getUriForSelfSignedJWT(URI uri)` is used to convert `"https://compute.googleapis.com/compute/v1/projects/"` to `"https://compute.googleapis.com/"`, so if the uri's scheme or host is null (for instance when user provided an uri like `compute.googleapis.com`), it should return the original uri.

### DIFF
--- a/oauth2_http/javatests/com/google/auth/oauth2/ServiceAccountJwtAccessCredentialsTest.java
+++ b/oauth2_http/javatests/com/google/auth/oauth2/ServiceAccountJwtAccessCredentialsTest.java
@@ -183,8 +183,6 @@ public class ServiceAccountJwtAccessCredentialsTest extends BaseSerializationTes
 
   @Test
   public void getUriForSelfSignedJWT_forStaticAudience_returnsURI() {
-    assertNull(ServiceAccountJwtAccessCredentials.getUriForSelfSignedJWT(null));
-
     URI uri = URI.create("compute.googleapis.com");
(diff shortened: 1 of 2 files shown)


oauth2_http/javatests/com/google/auth/oauth2/ServiceAccountJwtAccessCredentialsTest.java
```diff
@@ -181,6 +181,15 @@ public void getUriForSelfSignedJWT() {
...
 
+  @Test
+  public void getUriForSelfSignedJWT_forStaticAudience_returnsURI() {
+    assertNull(ServiceAccountJwtAccessCredentials.getUriForSelfSignedJWT(null));
+
+    URI uri = URI.create("compute.googleapis.com");
+    URI expected = URI.create("compute.googleapis.com");
+    assertEquals(expected, ServiceAccountJwtAccessCredentials.getUriForSelfSignedJWT(uri));
+  }
+
```
chingor13, line 192: This looks like 2 different tests, so let's split into 2.

chingor13 (commented):
Please add more context to the PR and to the commit message that will be created for the changelog rather as to what's actually being fixed.

Also, please get Silvano's stamp on this fix as I'm not sure what the expected behavior is supposed to be.

silvolu (approved):

Neenu1995 (approved):

PR #646, opened 2021-04-29
feat: add pre-emptive async oauth token refreshes [draft]

igorbernstein2:
This is currently a rough sketch and should not be merged. I just wanted to get some feedback here.

The current implementation of oauth refresh offloads the IO to a separate executor, however when a token expires, all calls to getRequestMetadata would hang until the token is refreshed.
This PR is a rough sketch to improve the situation by adding a stale state to token. If a token is within a minute of expiration, the first request to notice this, will spawn a refresh on the executor and immediately return with the existing token. This avoids hourly latency spikes for grpc.

The implementation uses guava's ListenableFutures to manage the async refresh state. Although the apis are marked BetaApi in guava 19, they are GA in guava 30

* The class introduces 3 distinct states:
  * Expired -  the token is not usable
  * Stale - the token is useable, but its time to refresh
  * Fresh - token can be used without any extra effort
* All of the funtionality of getRequestMetadata has been extracted to asyncFetch. The new function will check if the token is unfresh and schedule a refresh using the executor
* asyncFetch uses ListenableFutures to wrap state: if the token is not expired, an immediate future is returned. If the token is expired the future of the refresh task is returned
* A helper refreshAsync 	& finishRefreshAsync are also introduced. They schedule the actual refresh and propagate the side effects
* To handle blocking invocation: the async functionality is re-used but a DirectExecutor is used. All ExecutionErrors are unwrapped. In most cases the stack trace will be preserved because of the DirectExecutor. However if the async & sync methods are interleaved, it's possible that a sync call will await an async refresh task. In this case the callers stacktrace will not be present.

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
@@ -40,18 +40,27 @@ import com.google.common.base.MoreObjects;
 import com.google.common.base.MoreObjects.ToStringHelper;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.Iterables;
-
+import com.google.common.util.concurrent.FutureCallback;
+import com.google.common.util.concurrent.Futures;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.ListenableFutureTask;
+import com.google.common.util.concurrent.MoreExecutors;
 import java.io.IOException;
 import java.io.ObjectInputStream;
+import java.io.Serializable;
 import java.net.URI;
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.Date;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.ServiceLoader;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import javax.annotation.Nullable;
 
 /**
  * Base type for Credentials using OAuth2.
@@ -59,12 +68,12 @@ import java.util.concurrent.Executor;
 public class OAuth2Credentials extends Credentials {
 
   private static final long serialVersionUID = 4556936364828217687L;
-  private static final long MINIMUM_TOKEN_MILLISECONDS = 60000L * 5L;
+  private static final long MINIMUM_TOKEN_MILLISECONDS = TimeUnit.MINUTES.toMillis(5);
+  private static final long REFRESH_MARGIN_MILLISECONDS = MINIMUM_TOKEN_MILLISECONDS + TimeUnit.MINUTES.toMillis(1);
 
   // byte[] is serializable, so the lock variable can be final
-  private final Object lock = new byte[0];
-  private Map<String, List<String>> requestMetadata;
-  private AccessToken temporaryAccess;
+  private volatile OAuthValue value = null;
+  private transient ListenableFuture<OAuthValue> refreshTask;
 
   // Change listeners are not serialized
   private transient List<CredentialsChangedListener> changeListeners;
@@ -97,7 +106,7 @@ public class OAuth2Credentials extends Credentials {
   @Deprecated
   public OAuth2Credentials(AccessToken accessToken) {
     if (accessToken != null) {
-      useAccessToken(accessToken);
+      this.value = OAuthValue.create(accessToken);
     }
   }
 
@@ -117,22 +126,19 @@ public class OAuth2Credentials extends Credentials {
   }
 
   public final AccessToken getAccessToken() {
-    return temporaryAccess;
+    OAuthValue localState = value;
+    if (localState != null) {
+      return localState.temporaryAccess;
+    }
+    return null;
   }
 
   @Override
   public void getRequestMetadata(final URI uri, Executor executor,
       final RequestMetadataCallback callback) {
-    Map<String, List<String>> metadata;
-    synchronized(lock) {
-      if (shouldRefresh()) {
-        // The base class implementation will do a blocking get in the executor.
-        super.getRequestMetadata(uri, executor, callback);
-        return;
-      }
-      metadata = Preconditions.checkNotNull(requestMetadata, "cached requestMetadata");
-    }
-    callback.onSuccess(metadata);
+
+    Futures.addCallback(asyncFetch(executor),
+        new FutureCallbackToMetadataCallbackAdapter(callback), MoreExecutors.directExecutor());
   }
 
   /**
@@ -141,12 +147,7 @@ public class OAuth2Credentials extends Credentials {
    */
   @Override
   public Map<String, List<String>> getRequestMetadata(URI uri) throws IOException {
-    synchronized(lock) {
-      if (shouldRefresh()) {
-        refresh();
-      }
-      return Preconditions.checkNotNull(requestMetadata, "requestMetadata");
-    }
+    return unwrapDirectFuture(asyncFetch(MoreExecutors.directExecutor())).requestMetadata;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
```diff
@@ -208,6 +210,27 @@ private ImpersonatedCredentials initializeImpersonatedCredentials() {
         .build();
   }
 
+  @Override
```
igorbernstein2, line 213: Note to reviewer: before this PR, the blocking version getRequestMetadata() was the primary codepath, so ExternalAccountCredentials only had to override that method. This subclass expected that the async version would just call the sync version. Which I think is a poor assumption since its an implementation detail. I corrected the behavior by overriding both.

Arguably a better approach is to make both getRequestMetadata final and have subclasses use  asyncFetch() as the source of truth, but I didn't want to leak guava to the surface of this library

oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
```diff
@@ -144,75 +152,179 @@ public void getRequestMetadata(
...
+        for (CredentialsChangedListener listener : changeListeners) {
+          listener.onChanged(this);
+        }
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+      } catch (Exception e) {
+        // noop
+      } finally {
+        if (this.refreshTask == finishedTask) {
+          this.refreshTask = null;
```
TimurSadykov, line 269: can we make refreshTask @VisibleForTesting and cover its reset to null this with some tests? To avoid some nasty regressions.

oauth2_http/javatests/com/google/auth/oauth2/OAuth2CredentialsTest.java
```diff
@@ -348,6 +397,211 @@ public void getRequestMetadata_temporaryToken_hasToken() throws IOException {
...
+    // The access token should available once the refresh thread completes
+    TestUtils.assertContainsBearerToken(
+        creds.getRequestMetadata(CALL_URI), refreshedToken.getTokenValue());
+    callback = new MockRequestMetadataCallback();
+    creds.getRequestMetadata(CALL_URI, realExecutor, callback);
+    TestUtils.assertContainsBearerToken(callback.awaitResult(), refreshedToken.getTokenValue());
+  }
+
+  @Test
+  public void getRequestMetadata_singleFlightErrorSharing()
```
TimurSadykov: what is the meaning of the "singleFlightErrorSharing"? The test makes sense, but unfamiliar with the terminology

TimurSadykov (changes requested):
Overall looks good, lets clarify if refresh slight behavior change is blocking. 

Some coverage of refreshTask would be great.

oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
```diff
@@ -144,75 +152,179 @@ public void getRequestMetadata(
...
-      }
-      if (requestMetadata == null) {
-        throw new NullPointerException("requestMetadata");
-      }
-      return requestMetadata;
-    }
+    return unwrapDirectFuture(asyncFetch(MoreExecutors.directExecutor())).requestMetadata;
   }
 
   /** Refresh the token by discarding the cached token and metadata and requesting the new ones. */
```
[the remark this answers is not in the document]
igorbernstein2: yep, I updated the docs. Let me know if its more it makes more sense

oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
igorbernstein2: I updated the existing tests to ensure that its managed properly. Let me know if you had something else in mind

igorbernstein2 (commented):
Thanks for reviewing! I think addressed all of your comments, PTAL

oauth2_http/javatests/com/google/auth/oauth2/OAuth2CredentialsTest.java
igorbernstein2: Internal link is go/singleflight, but it seems to exist in public domain as well. Single flight is a mechanic to coalesce multiple requests into one. A general implementation would allow for batching, but in this case we only have a single operation type that we need to combine

oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
TimurSadykov: looks good, thx!

oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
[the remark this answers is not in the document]
TimurSadykov: The comment looks good, FYI @silvolu to double check that its not a breaking change

oauth2_http/javatests/com/google/auth/oauth2/OAuth2CredentialsTest.java
TimurSadykov: Thanks for pointers :)

paulmordont:
Hey @igorbernstein2 ! Thx a lot for the PR, from my side everything looks good. Looking forward to testing this in the next bigtable client release :)

oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
```diff
@@ -144,75 +152,182 @@ public void getRequestMetadata(
...
+   *
+   * <p>If a fresh token is already available, it will be immediately returned. Otherwise a refresh
+   * will be scheduled using the passed in executor. While a token is being freshed, a stale value
+   * will be returned.
+   */
+  private ListenableFuture<OAuthValue> asyncFetch(Executor executor) {
+    AsyncRefreshResult refreshResult = null;
+
+    // Schedule a refresh as necessary
     synchronized (lock) {
```
paulmordont, line 190: I wonder Is it still necessary to go inside the lock if the token state `Fresh`? Is it viable to just do the following:
```
if (getState() == CacheState.Fresh) {
    return Futures.immediateFuture(value);
}
```

igorbernstein2: Thats good observation, I added the optimization. Thanks!

TimurSadykov (commented):
Looks good to me

TimurSadykov (approved):

oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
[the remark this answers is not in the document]
TimurSadykov: @igorbernstein2 i synced with @silvolu, there is no concern with this change.

igorbernstein2:
I created cl/373364841, I had to hand edit a couple of irrelevant files:
- added missing RunWith on unrelated tests
- added missing deps in BUILD

I requested a global tap presubmit

Also, I updated the codestyle suggestions from critique

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
```diff
@@ -208,6 +210,27 @@ private ImpersonatedCredentials initializeImpersonatedCredentials() {
...
+  public void getRequestMetadata(
+      URI uri, Executor executor, final RequestMetadataCallback callback) {
+    super.getRequestMetadata(
+        uri,
+        executor,
+        new RequestMetadataCallback() {
+          @Override
+          public void onSuccess(Map<String, List<String>> metadata) {
+            metadata = addQuotaProjectIdToRequestMetadata(quotaProjectId, metadata);
```
lsirac, line 222: do we need to do this here? the quotaProjectId is being added in the override below, which seems like is always called

igorbernstein2: This doesn't call the override below. It calls `super.getRequestMetadata`

igorbernstein2: But I did just notice that I had a typo & invoked super.getRequestMetadata twice.

lsirac: ah calling super.getRequestMetadata used to call the override below

igorbernstein2: yea, I dont love the parallel logic paths, and tried to call it out in the comment above, but I dont have a better approach. If you have ideas, lets keep the discussion in https://github.com/googleapis/google-auth-library-java/pull/646/files#r625406829

lsirac (approved):

silvolu (approved):

lesv (approved):

PR #650, opened 2021-05-06
feat: add Id token support for UserCredentials

### LINKED ISSUE #157: Why does UserCredentials use access_token?
When `UserCredentials` obtains a refreshed token from credentials that were loaded from `application_default_credentials.json`, it uses `"access_token"` instead of `"id_token"`.  My Cloud Endpoints setup seems to require a base64 JWT, i.e. `"id_token"`.  

If this is a new use case, I'll be happy to submit a PR to add this capability.

### LINKED ISSUE #468: Add id_token capability on User Credential

### LINKED ISSUE #492: feat: Generate id_token with user credential

TimurSadykov:

Fixes #492  ☕️
Fixes #468 
Fixes #157


oauth2_http/java/com/google/auth/oauth2/IdTokenCredentials.java
```diff
@@ -110,7 +110,18 @@
...
+
+    // if UserCredential is used as id token provider - we need to refresh access token
+    if (this.idTokenProvider instanceof UserCredentials) {
+      try {
+        this.refresh();
+      } catch (IOException ignored) {
+        // ignore because all we need here is to reset the metadata
+      }
+    } else {
+      // target audience is not used for UserCredentials
```
silvolu, line 122: technically is "can't be used"

oauth2_http/java/com/google/auth/oauth2/UserCredentials.java
```diff
@@ -210,6 +196,33 @@ public AccessToken refreshAccessToken() throws IOException {
     return new AccessToken(accessToken, new Date(expiresAtMilliseconds));
   }
 
+  /**
+   * Returns a Google ID Token from the refresh token response.
+   *
+   * @param targetAudience the aud: field the IdToken should include. This is not unused for
```
silvolu, line 202: "This is not used", or "This can't be used" would make it more clear that is not a library choice but a limitation of the flow.

oauth2_http/java/com/google/auth/oauth2/UserCredentials.java
```diff
@@ -210,6 +196,33 @@ public AccessToken refreshAccessToken() throws IOException {
...
+    GenericData responseData = doRefreshAccessToken();
+    String idTokenKey = "id_token";
+    if (responseData.containsKey(idTokenKey)) {
+      String idTokenString =
+          OAuth2Utils.validateString(responseData, idTokenKey, PARSE_ERROR_PREFIX);
+      return IdToken.create(idTokenString);
+    }
+
+    throw new IOException(
+        "UserCredentials instance cannot refresh id token unless you are logged"
```
silvolu, line 221: That page is not very clear about this, I think we need to write something like

"UserCredentials can obtain an id token only when authentication has been obtained through gcloud running 'gcloud auth login --update-adc' or 'gcloud auth application-default login'. The latter form would not work for Cloud Run, but would still generate an id token.

oauth2_http/javatests/com/google/auth/oauth2/UserCredentialsTest.java
```diff
@@ -699,6 +705,54 @@ public void onFailure(Throwable exception) {
     assertTrue("Should have run onSuccess() callback", success.get());
   }
 
+  @Test
+  public void IdTokenCredentials_SdkClientId_correct() throws IOException {
```
silvolu, line 709: I'm not sure I understand the tests: the ability of getting an id token doesn't depend on the Client ID, depends on the scopes requested.

gcloud always requests the scopes to obtain an id token, but the client id would be different if for example auth has been obtained with gcloud auth application-default login.

The requirement for the gcloud client id is to authenticate testing cloud run services.

silvolu (changes requested):

oauth2_http/java/com/google/auth/oauth2/UserCredentials.java
TimurSadykov: updated

oauth2_http/javatests/com/google/auth/oauth2/UserCredentialsTest.java
TimurSadykov: that makes sense, thanks!

oauth2_http/java/com/google/auth/oauth2/IdTokenCredentials.java
TimurSadykov: fixed

oauth2_http/java/com/google/auth/oauth2/UserCredentials.java
TimurSadykov: Thanks for suggestion, updated the exception

silvolu (approved):

oauth2_http/javatests/com/google/auth/oauth2/UserCredentialsTest.java
```diff
@@ -722,6 +722,8 @@ public void IdTokenCredentials_WithUserEmailScope_success() throws IOException {
     IdTokenCredentials tokenCredential =
         IdTokenCredentials.newBuilder().setIdTokenProvider(credentials).build();
 
+    tokenCredential.getRequestMetadata();
```
elharo, line 725: This looks like a no-op. What's going on here?

elharo: OK, need a comment to that effect to keep future devs from removing this since this is a non-standard naming convention and confusing

TimurSadykov: added assert for null and a comment for the line

elharo (approved):

PR #572, opened 2021-02-16
feat: add self signed jwt support

arithmetic1728:
This PR adds self signed jwt support (https://google.aip.dev/auth/4111). Googlers see this go/yoshi-self-signed-jwt for more details. If user uses service account credentials, and doesn't provide custom endpoint/scopes/audience, then self signed jwt should be used. Since custom endpoint is now supported (cl/356594013) and users cannot provide default audience for `ServiceAccountCredentials`, **we automatically use self signed jwt if scopes are not provided by the users**.

### Details
1. Scopes. 
We add a new `defaultScopes` (for client lib) to `GoogleCredentials` to distinguish with the existing `scopes` (for users), and updated the subclasses. 
- For `AppEngineCredentials`, `ComputeEngineCredentials` classes, `defaultScopes` will used  if `scopes` is not provided, there are no other changes.
- For `ServiceAccountCredentials`, if `scopes` is not provided, then a `ServiceAccountJWTAccessCredentials` instance will be created inside `ServiceAccountCredentials`, and will be used for self signed jwt.
- For other type of credentials where scopes are not used, `defaultScopes` has no effect.

2. Simplified self signed jwt audience. Googlers see [go/simpler-JWT-token-audience](https://docs.google.com/document/d/1gdXBtOFogIWOcGUlZSScCqEjIe8cQ_ihSMzXwyQU0mo/edit#) 
When self signed jwt is used for service account credentials, in `getRequestMetadata(uri)`, we replace the uri with its `http(s)://{host_name}/` prefix. For instance, http client normally passes uri like `https://compute.googleapis.com/compute/v1/projects/`, but the correct audience should be `https://compute.googleapis.com/`. 

#### Follow up PRs
It will be trivial changes to gax-java and client libs to integrate this feature.

This [PR](https://github.com/googleapis/gax-java/pull/1302/files) shows how gax-java integrates the feature. gax-java just needs to pass `defaultScopes` to the auth lib.

This [PR](https://github.com/arithmetic1728/java-pubsub/pull/2) shows how GAPIC client integrates the feature. GAPIC client just needs to pass the default scopes via `setDefaultScopes` instead of `setScopesToApply`.

PR #689, opened 2021-06-22
feat: allow scopes for self signed jwt

arithmetic1728:
This PR allows self signed jwt to use scopes. 
AIP: https://google.aip.dev/auth/4111
googlers see: go/yoshi-self-signed-jwt-phase-2

In ServiceAccountCredentials, this PR now uses JwtCredentials instead of ServiceAccountJwtCredentials. As a result, this PR reverted the changes made to ServiceAccountJwtCredentials in #572 #642 so the ServiceAccountJwtCredentials is the same as what it was before phrase 1. 

The current behavior for ServiceAccountCredentials is:
```
if (hasScopes):
    if (useJwtAccessWithScope):
        // create a self signed JWT with "scope" set to the scope
    else:
        // call oauth token endpoint
else:
        // create a self signed JWT with modified uri as the audience
``` 

Follow up PRs:
(1) gax-java: https://github.com/googleapis/gax-java/pull/1420
(2) gapic-generator-java: it will be a very simple change, same as https://github.com/arithmetic1728/java-kms/pull/2/files. Since Gapic clients will always sets scopes, once self signed JWT is enabled in the future, it will always use self signed JWT with scope claim.

This PR has been tested with cloudkms: https://github.com/arithmetic1728/java-kms/pull/2/files

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
@@ -1016,8 +1016,8 @@ public class ServiceAccountCredentials extends GoogleCredentials
               + " providing uri to getRequestMetadata.");
     }
 
-    // If scopes are provided but we cannot use self signed JWT, then use scopes in the normal oauth
-    // way.
+    // If scopes are provided but we cannot use self signed JWT, then use scopes to get access
+    // token.
     if (!createScopedRequired() && !useJwtAccessWithScope) {
       return super.getRequestMetadata(uri);
     }
--- a/oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
@@ -110,7 +110,7 @@ public class ServiceAccountCredentials extends GoogleCredentials
   private final Collection<String> defaultScopes;
   private final String quotaProjectId;
   private final int lifetime;
-  private final boolean alwaysUseJwtAccess;
+  private final boolean useJWTAccessWithScope;
 
   private transient HttpTransportFactory transportFactory;
 
@@ -134,7 +134,7 @@ public class ServiceAccountCredentials extends GoogleCredentials
    *     most 43200 (12 hours). If the token is used for calling a Google API, then the value should
    *     be at most 3600 (1 hour). If the given value is 0, then the default value 3600 will be used
    *     when creating the credentials.
-   * @param alwaysUseJwtAccess whether self signed JWT should be always used.
+   * @param useJWTAccessWithScope whether self signed JWT with scopes should be always used.
    */
   ServiceAccountCredentials(
       String clientId,
@@ -149,7 +149,7 @@ public class ServiceAccountCredentials extends GoogleCredentials
       String projectId,
       String quotaProjectId,
       int lifetime,
-      boolean alwaysUseJwtAccess) {
+      boolean useJWTAccessWithScope) {
     this.clientId = clientId;
     this.clientEmail = Preconditions.checkNotNull(clientEmail);
     this.privateKey = Preconditions.checkNotNull(privateKey);
@@ -170,7 +170,7 @@ public class ServiceAccountCredentials extends GoogleCredentials
       throw new IllegalStateException("lifetime must be less than or equal to 43200");
     }
     this.lifetime = lifetime;
-    this.alwaysUseJwtAccess = alwaysUseJwtAccess;
+    this.useJWTAccessWithScope = useJWTAccessWithScope;
   }
 
   /**
@@ -692,7 +692,7 @@ public class ServiceAccountCredentials extends GoogleCredentials
         projectId,
         quotaProjectId,
         lifetime,
-        alwaysUseJwtAccess);
+        useJWTAccessWithScope);
   }
 
   /**
@@ -709,13 +709,13 @@ public class ServiceAccountCredentials extends GoogleCredentials
   }
 
   /**
-   * Clones the service account with a new alwaysUseJwtAccess value.
+   * Clones the service account with a new useJWTAccessWithScope value.
    *
-   * @param alwaysUseJwtAccess whether self signed JWT should be used
-   * @return the cloned service account credentials with the given alwaysUseJwtAccess
+   * @param useJWTAccessWithScope whether self signed JWT with scopes should be used
+   * @return the cloned service account credentials with the given useJWTAccessWithScope
    */
-  public ServiceAccountCredentials createWithAlwaysUseJwtAccess(boolean alwaysUseJwtAccess) {
-    return this.toBuilder().setAlwaysUseJwtAccess(alwaysUseJwtAccess).build();
+  public ServiceAccountCredentials createWithUseJWTAccessWithScope(boolean useJWTAccessWithScope) {
+    return this.toBuilder().setUseJWTAccessWithScope(useJWTAccessWithScope).build();
   }
 
   @Override
@@ -733,7 +733,7 @@ public class ServiceAccountCredentials extends GoogleCredentials
         projectId,
         quotaProjectId,
         lifetime,
-        alwaysUseJwtAccess);
+        useJWTAccessWithScope);
   }
 
   public final String getClientId() {
@@ -781,8 +781,8 @@ public class ServiceAccountCredentials extends GoogleCredentials
     return lifetime;
   }
 
-  public boolean getAlwaysUseJwtAccess() {
-    return alwaysUseJwtAccess;
+  public boolean getUseJWTAccessWithScope() {
+    return useJWTAccessWithScope;
   }
 
   @Override
@@ -843,7 +843,7 @@ public class ServiceAccountCredentials extends GoogleCredentials
         defaultScopes,
         quotaProjectId,
         lifetime,
-        alwaysUseJwtAccess);
+        useJWTAccessWithScope);
   }
 
   @Override
@@ -859,7 +859,7 @@ public class ServiceAccountCredentials extends GoogleCredentials
         .add("serviceAccountUser", serviceAccountUser)
         .add("quotaProjectId", quotaProjectId)
         .add("lifetime", lifetime)
-        .add("alwaysUseJwtAccess", alwaysUseJwtAccess)
+        .add("useJWTAccessWithScope", useJWTAccessWithScope)
         .toString();
   }
 
@@ -879,7 +879,7 @@ public class ServiceAccountCredentials extends GoogleCredentials
         && Objects.equals(this.defaultScopes, other.defaultScopes)
         && Objects.equals(this.quotaProjectId, other.quotaProjectId)
         && Objects.equals(this.lifetime, other.lifetime)
-        && Objects.equals(this.alwaysUseJwtAccess, other.alwaysUseJwtAccess);
+        && Objects.equals(this.useJWTAccessWithScope, other.useJWTAccessWithScope);
   }
 
   String createAssertion(JsonFactory jsonFactory, long currentTime, String audience)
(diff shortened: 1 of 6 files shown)


Neenu1995 (approved):

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -950,14 +985,19 @@ public void getRequestMetadata(
   /** Provide the request metadata by putting an access JWT directly in the metadata. */
   @Override
   public Map<String, List<String>> getRequestMetadata(URI uri) throws IOException {
-    if (scopes.isEmpty() && defaultScopes.isEmpty() && uri == null) {
-      throw new IOException(
-          "Scopes and uri are not configured for service account. Either pass uri"
-              + " to getRequestMetadata to use self signed JWT, or specify the scopes"
-              + " by calling createScoped or passing scopes to constructor.");
+    if (createScopedRequired()) {
```
TimurSadykov, line 988: i think we don't want to use this property here. It seems like not correct at the moment, because CreateScoped might not be actually required and Scopes could be empty at the same time. You can see functional tests examples as the reference. First couple 200 test cases will return true for createScopedRequired. The reason is that, as I understand, ESF will set scope based on audience. 

Therefore, I think that property is subject to change in the future.

oauth2_http/java/com/google/auth/oauth2/ServiceAccountJwtAccessCredentials.java
```diff
@@ -332,35 +331,17 @@ public boolean hasRequestMetadataOnly() {
     return true;
   }
 
-  /**
-   * Self signed JWT uses uri as audience, which should have the "https://{host}/" format. For
```
TimurSadykov, line 336: why this no longer needed?

TimurSadykov (changes requested):

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -492,7 +484,8 @@ static ServiceAccountCredentials fromPkcs8(
         serviceAccountUser,
         projectId,
         quotaProject,
-        DEFAULT_LIFETIME_IN_SECONDS);
+        DEFAULT_LIFETIME_IN_SECONDS,
+        false);
```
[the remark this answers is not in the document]
arithmetic1728: This feature is disabled by default because it doesn't work for all client libraries. Gapic clients will set this flag to `true` to enable the feature.

oauth2_http/java/com/google/auth/oauth2/ServiceAccountJwtAccessCredentials.java
arithmetic1728: This was added by me in phrase 1, so I could create a ServiceAccountJwtAccessCredentials inside service account credentials to do the actual self signed jwt work. In this PR I switched to use JwtCredentials instead, so I reverted the changes made to ServiceAccountJwtAccessCredentials.

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -937,11 +949,34 @@ String createAssertionForIdToken(
...
+        .build();
+  }
+
   @Override
   public void getRequestMetadata(
       final URI uri, Executor executor, final RequestMetadataCallback callback) {
-    if (jwtCredentials != null && uri != null) {
-      jwtCredentials.getRequestMetadata(uri, executor, callback);
+    if (alwaysUseJwtAccess) {
+      // This will call getRequestMetadata(URI uri), which handles self signed JWT logic.
```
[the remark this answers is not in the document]
arithmetic1728: executor is for async, here we just do a blocking call to create a self signed jwt for better efficiency, similar as what ServiceAccountJwtAccessCredentials does:

https://github.com/googleapis/google-auth-library-java/blob/dfe118c261aadf137a3cf47a7acb9892c7a6db4d/oauth2_http/java/com/google/auth/oauth2/ServiceAccountJwtAccessCredentials.java#L353-L358

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
arithmetic1728: This is the original behavior of the library. I would like to keep the original behavior in this PR so it doesn't cause any problems.

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -698,7 +691,8 @@ public GoogleCredentials createScoped(
         serviceAccountUser,
         projectId,
         quotaProjectId,
-        lifetime);
+        lifetime,
+        alwaysUseJwtAccess);
```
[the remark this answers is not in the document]
arithmetic1728: The constructor has default visibility, I think it is only visible to this package. So it won't be a breaking change to the users.

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -728,7 +732,8 @@ public GoogleCredentials createDelegated(String user) {
         user,
         projectId,
         quotaProjectId,
-        lifetime);
+        lifetime,
+        alwaysUseJwtAccess);
```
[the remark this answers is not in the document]
arithmetic1728: I don't think so. See the comment above.

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -109,9 +110,9 @@
   private final Collection<String> defaultScopes;
   private final String quotaProjectId;
   private final int lifetime;
+  private final boolean alwaysUseJwtAccess;
```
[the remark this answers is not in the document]
arithmetic1728: done

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
arithmetic1728: yep you are right, either scopes or uri work.

arithmetic1728: Nope. I would like to revert to the commit before phrase 1. In other words, always use scopes and never use uri.

https://github.com/googleapis/google-auth-library-java/blob/497d4e739cf83876b77d1eda8c9618968cf8165a/oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java#L414-L418

The integration test is not correct, it should be updated. It works because it uses self signed jwt with a modified url as audience (used the `https://<api>.googleapis.com/` part from the url). This is what I introduced in phrase 1. However this logic is wrong, url may be a regional endpoint or any url unrelated to audience, there is no way to derive the audience from uri. This is also the reason why I reverted the url modification logic in ServiceAccountJwtAccessCredentials in this PR. 
 
In other words, we should always scopes (never use uri) for service account credentials for Java. In this PR if `useJwtAccessWithScope` is `true`, I still included the `uri` logic because it won't be used by client libraries - GAPIC clients always set scopes we will always use scopes.

arithmetic1728: I updated the PR to only use scopes and never use uri, and updated the functional tests.

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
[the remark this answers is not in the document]
TimurSadykov: maybe add a comment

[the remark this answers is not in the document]
arithmetic1728: done

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -109,9 +110,9 @@
   private final Collection<String> defaultScopes;
   private final String quotaProjectId;
   private final int lifetime;
+  private final boolean useJWTAccessWithScope;
```
TimurSadykov, line 113: nit: this should be useJwtAccessWithScope. Sorry, I misstyped my suggestion :) per Jon, if used within name, JWT should be Jwt.

arithmetic1728: yea, updated the name.

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
TimurSadykov: Thanks for explaining the change! However, the fact that those functional tests will fail with the change - means this is a breaking change, right? People could have used it this way already.

arithmetic1728: added uri back

oauth2_http/javatests/com/google/auth/oauth2/ServiceAccountCredentialsTest.java
```diff
@@ -1387,7 +1461,11 @@ private void verifyJwtAccess(Map<String, List<String>> metadata) throws IOExcept
         JsonWebSignature.parse(GsonFactory.getDefaultInstance(), assertion);
     assertEquals(CLIENT_EMAIL, signature.getPayload().getIssuer());
     assertEquals(CLIENT_EMAIL, signature.getPayload().getSubject());
-    assertEquals(JWT_AUDIENCE, signature.getPayload().getAudience());
+    if (expectedScopeClaim != null) {
+      assertEquals(expectedScopeClaim, signature.getPayload().get("scope"));
```
TimurSadykov, line 1465: add check that aud is not present

oauth2_http/javatests/com/google/auth/oauth2/ServiceAccountCredentialsTest.java
```diff
@@ -1387,7 +1461,11 @@ private void verifyJwtAccess(Map<String, List<String>> metadata) throws IOExcept
...
+    } else {
+      assertEquals(JWT_AUDIENCE, signature.getPayload().getAudience());
```
TimurSadykov, line 1467: add check that scope is not present

oauth2_http/javatests/com/google/auth/oauth2/ServiceAccountJwtAccessCredentialsTest.java
```diff
@@ -172,29 +171,6 @@ public void getAuthenticationType_returnsJwtAccess() throws IOException {
     assertEquals(credentials.getAuthenticationType(), "JWTAccess");
   }
 
-  @Test
```
TimurSadykov, line 175: restore these tests for JWTCredential? Since you have moved the logic there

TimurSadykov (changes requested):

oauth2_http/javatests/com/google/auth/oauth2/ServiceAccountJwtAccessCredentialsTest.java
arithmetic1728: These tests are moved to ServiceAccountCredentialsTest.

oauth2_http/javatests/com/google/auth/oauth2/ServiceAccountCredentialsTest.java
arithmetic1728: done

oauth2_http/javatests/com/google/auth/oauth2/ServiceAccountCredentialsTest.java
arithmetic1728: done

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -950,17 +1009,31 @@ public void getRequestMetadata(
...
+          "Scopes and uri are not configured for service account. Specify the scopes"
+              + " by calling createScoped or passing scopes to constructor or"
+              + " providing uri to getRequestMetadata.");
     }
-    if (jwtCredentials != null && uri != null) {
-      return jwtCredentials.getRequestMetadata(uri);
-    } else {
+
+    // If scopes are provided but we cannot use self signed JWT, then use scopes in the normal oauth
+    // way.
```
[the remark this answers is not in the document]
arithmetic1728: done

TimurSadykov (approved):

PR #699, opened 2021-07-22
fix: use source credential expiration when STS does not return expires_in

lsirac:
For the CAB flow, STS only returns expires_in if the source access token belongs to a service account.
For other source credential types, we can copy the source credentials expiration as the generated downscoped token
will always have the same expiration time as the source credentials.

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/DownscopedCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/DownscopedCredentials.java
@@ -129,7 +129,21 @@ public final class DownscopedCredentials extends OAuth2Credentials {
             .setInternalOptions(credentialAccessBoundary.toJson())
             .build();
 
-    return handler.exchangeToken().getAccessToken();
+    AccessToken downscopedAccessToken = handler.exchangeToken().getAccessToken();
+
+    // The STS endpoint will only return the expiration time for the downscoped token if the
+    // original access token
+    // represents a service account.
+    // The downscoped token's expiration time will always match the source credential expiration.
+    // When no expires_in is returned, we can copy the source credential's expiration time.
+    if (downscopedAccessToken.getExpirationTime() == null) {
+      AccessToken sourceAccessToken = this.sourceCredential.getAccessToken();
+      if (sourceAccessToken.getExpirationTime() != null) {
+        return new AccessToken(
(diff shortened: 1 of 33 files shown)


oauth2_http/javatests/com/google/auth/oauth2/DownscopedCredentialsTest.java
```diff
@@ -107,11 +109,35 @@ public void refreshAccessToken() throws IOException {
         "urn:ietf:params:oauth:token-type:access_token", query.get("requested_token_type"));
   }
 
+  @Test
+  public void refreshAccessToken_userCredentials_expectExpiresInCopied() throws IOException {
```
TimurSadykov, line 113: add doc explaining the behavior that you are trying to test... from PR description :)

oauth2_http/javatests/com/google/auth/oauth2/StsRequestHandlerTest.java
```diff
@@ -248,4 +248,30 @@ public void run() throws Throwable {
             });
     assertEquals(e, thrownException);
   }
+
+  @Test
+  public void exchangeToken_noExpiresInReturned() throws IOException {
```
TimurSadykov, line 253: Also document why expires_in might not be returned?

TimurSadykov (changes requested):

oauth2_http/java/com/google/auth/oauth2/DownscopedCredentials.java
```diff
@@ -129,7 +129,21 @@ public AccessToken refreshAccessToken() throws IOException {
...
+    AccessToken downscopedAccessToken = handler.exchangeToken().getAccessToken();
+
+    // The STS endpoint will only return the expiration time for the downscoped token if the
+    // original access token
+    // represents a service account.
+    // The downscoped token's expiration time will always match the source credential expiration.
+    // When no expires_in is returned, we can copy the source credential's expiration time.
+    if (downscopedAccessToken.getExpirationTime() == null) {
+      AccessToken sourceAccessToken = this.sourceCredential.getAccessToken();
+      if (sourceAccessToken.getExpirationTime() != null) {
```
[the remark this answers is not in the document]
lsirac: Yes. They should still be able to use the downscoped token (e.g. can consume via OAuth2Credentials).

PR #702, opened 2021-07-27
feat: downscoping with credential access boundaries

lsirac:
See go/cab-client. This feature is publicly documented [here](https://cloud.google.com/iam/docs/downscoping-short-lived-credentials). 

Summary:

- Adds a new DownscopedCredentials class that enables the ability to downscope, or restrict, the IAM permissions that a short-lived credential can use for Cloud Storage. This is done by defining a CredentialAccessBoundary which specifies the upper bound of permissions the downscoped credential will be able to access. 
- OAuth2CredentialsWithRefresh enables access token refresh via a developer defined refresh handler.
- With CAB, STS may not always return an expires_in. The STS utility has been updated to reflect this. When not returned, the expires_in is copied from the source credential, when available.
- Includes integration tests with a one time use setup script (already ran).
- Samples/documentation will be provided in a separate PR.

### DIFF
--- a/scripts/downscoping-with-cab-setup.sh
+++ b/scripts/downscoping-with-cab-setup.sh
@@ -1,0 +1,96 @@
+#!/bin/bash
+
+# Copyright 2021 Google LLC
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are
+# met:
+#
+#    * Redistributions of source code must retain the above copyright
+# notice, this list of conditions and the following disclaimer.
+#    * Redistributions in binary form must reproduce the above
+# copyright notice, this list of conditions and the following disclaimer
+# in the documentation and/or other materials provided with the
+# distribution.
+#    * Neither the name of Google LLC nor the names of its
+# contributors may be used to endorse or promote products derived from
+# this software without specific prior written permission.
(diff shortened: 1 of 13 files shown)


scripts/downscoping-with-cab-setup.sh
```diff
@@ -0,0 +1,96 @@
+#!/bin/bash
```
elharo, line 1: When/by whom or what is this script run? 

Doing this in bash is hacky. Not a total blocker for purely test code, but Java code would be preferable.

lsirac: This is a convenience script that makes it easy for us to port the integration tests to a new project or quickly set everything up if someone goes into pantheon and makes changes. It is similar to the script we had for the Workload Identity Federation integration tests. We decided to use a script as doing it in the test code means we may have to deal with potential intermittent test failures and recovery. You can take a look at go/cab-client-integration-testing if you want to read about our approach. 

It has already been run, so you don't need to run it. If you want to change the configuration, you can run it again. Ideally, this script will never even need to be executed again.

TimurSadykov (approved):
LGTM

Neenu1995 (approved):

PR #729, opened 2021-09-07
feat: add support for Workforce Pools

lsirac:
See go/workforce-pools-client-support.


TimurSadykov:
@lsirac unrelated to this feature, but I see there is a transportFactoryClassName that is never used. Is it expected?

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
```diff
@@ -325,19 +326,30 @@ static ExternalAccountCredentials fromJson(
...
-        quotaProjectId,
-        clientId,
-        clientSecret,
-        /* scopes= */ null,
-        /* environmentProvider= */ null);
+
+    // Validate that the userProject parameter, if present, is for a Workforce configuration.
+    // This doesn't validate the if the workforce audience is valid, only if it follows the
+    // workforce pattern.
+    Pattern workforceAudiencePattern =
```
TimurSadykov, line 333: This should be within the relevant class right? Otherwise this check will not work if we create it via public constructor

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
```diff
@@ -325,19 +326,30 @@ static ExternalAccountCredentials fromJson(
...
+        Pattern.compile("^.+/locations/.+/workforcePools/.+/providers/.+$");
+    if (userProject != null && !workforceAudiencePattern.matcher(audience).matches()) {
+      throw new IOException(
+          "The workforce_pool_user_project parameter should only be provided for a Workforce Pool configuration.");
+    }
+
+    return IdentityPoolCredentials.newBuilder()
```
TimurSadykov, line 340: Return either IndentityPoolCredential or WorkforcePoolCredential.. 

If done separately, we can explicitly implement and document the condition where user auth takes priority over service. Right now it is not easy to find out.

oauth2_http/javatests/com/google/auth/oauth2/IdentityPoolCredentialsTest.java
```diff
@@ -326,6 +327,38 @@ public void refreshAccessToken_withoutServiceAccountImpersonation() throws IOExc
     assertEquals(transportFactory.transport.getAccessToken(), accessToken.getTokenValue());
   }
 
+  @Test
```
TimurSadykov, line 330: A test with clientId and userProject validating that userProject gets priority? With separate credential this can be a ExternalCredential test.

TimurSadykov (changes requested):

oauth2_http/javatests/com/google/auth/oauth2/IdentityPoolCredentialsTest.java
lsirac: Client ID/secret are not currently used.

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
lsirac: Yep - moved.

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
```diff
@@ -155,39 +154,32 @@ private boolean hasHeaders() {
...
+        builder.clientId,
+        builder.clientSecret,
+        builder.scopes,
+        builder.environmentProvider);
+    this.identityPoolCredentialSource = (IdentityPoolCredentialSource) builder.credentialSource;
+    this.workforcePoolUserProject = builder.workforcePoolUserProject;
+  }
+
+  @Nullable
+  public String getWorkforcePoolUserProject() {
```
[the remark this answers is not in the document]
lsirac: I think adding a new credentials class for this one field is a little overkill. IdentityPool is generic and we thought would handle the workforce case initially (see go/workforce-pools-client-support why we need this field with workforce but not workload). 

In the most common use-case, the credential itself is not used directly (ADC). Developers are generating a workforce config via gcloud and settings their ADC env variable to point to it - they are not interfacing with the credentials class directly, so I think this is OK.

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
```diff
@@ -202,6 +194,15 @@ public AccessToken refreshAccessToken() throws IOException {
       stsTokenExchangeRequest.setScopes(new ArrayList<>(scopes));
     }
 
+    // If the workforcePoolUserProject is set, we need to pass it to STS via the the internal
+    // options param.
+    if (workforcePoolUserProject != null && !workforcePoolUserProject.isEmpty()) {
```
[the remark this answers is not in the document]
lsirac: Can you elaborate on this?

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
lsirac: See comment above.

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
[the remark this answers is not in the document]
TimurSadykov: It is not for the field.. it is for all the conditions you are adding. I just used the field for the comment. 

The main motivation is that it seems like we can expect more and this area is evolving.

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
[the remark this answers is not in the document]
TimurSadykov: its a one way to move out this code into workforce credential if we go ahead and add one.

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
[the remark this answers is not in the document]
lsirac: For basic workforce support, this is pretty much it. IdentityPoolCredentials was created to be generic and I think this is the way to go for this change.

This is also what was proposed in go/workforce-pools-client-support and what was decided on for all languages. If we want to add a new credential type, we'd have to do this in all the other languages for parity.

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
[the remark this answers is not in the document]
TimurSadykov: @lsirac One simple alternative is to define properties like isWorkForcePoolMode (and use the line above as a get implementation) and isIdentityPoolMode and document the difference. And use those in conditional statements. Otherwise there is only indirect reference to workforce pool.

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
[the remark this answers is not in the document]
TimurSadykov: An alternative suggested in a different comment

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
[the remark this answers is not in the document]
lsirac: Ok, I added an isWorkforcePoolConfiguration() method with some documentation.

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
```diff
@@ -202,6 +200,15 @@ public AccessToken refreshAccessToken() throws IOException {
...
```
TimurSadykov, line 205: replace with isWorkforcePoolConfiguration?

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
```diff
@@ -269,22 +276,24 @@ private String getSubjectTokenFromMetadataServer() throws IOException {
     }
   }
 
+  /**
+   * Returns whether or not the current configuration is for Workforce Pools (which enable 3p user
+   * identities, rather than workloads).
```
TimurSadykov, line 281: nit: Maybe you can be i bit more crisp, specifically saying that if this true - workloads will be disabled.

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
```diff
@@ -269,22 +276,24 @@ private String getSubjectTokenFromMetadataServer() throws IOException {
...
+   */
+  private boolean isWorkforcePoolConfiguration() {
```
TimurSadykov, line 283: Should be public so developers can easily distinguish between two configurations.

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
```diff
@@ -269,22 +276,24 @@ private String getSubjectTokenFromMetadataServer() throws IOException {
...
+    // This doesn't validate the if the audience is valid, only if it follows the
+    // workforce pattern.
+    Pattern workforceAudiencePattern =
+        Pattern.compile("^.+/locations/.+/workforcePools/.+/providers/.+$");
+    return workforcePoolUserProject != null
```
TimurSadykov, line 288: check non-empty as well?

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
lsirac: Going to keep it like this for now

TimurSadykov (approved):
LGTM

Neenu1995 (approved):

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
```diff
@@ -155,39 +155,37 @@ private boolean hasHeaders() {
...
+    this.workforcePoolUserProject = builder.workforcePoolUserProject;
+
+    if (workforcePoolUserProject != null && !isWorkforcePoolConfiguration()) {
+      throw new IllegalArgumentException(
+          "The workforce_pool_user_project parameter should only be provided for a Workforce Pool configuration.");
+    }
+  }
+
+  @Nullable
+  public String getWorkforcePoolUserProject() {
```
chingor13, line 187: does this need to be public?

chingor13 (commented):
Just a question on whether we need to expose those workforce pool methods?

PR #749, opened 2021-09-28
fix: timing of stale token refreshes on ComputeEngine

igorbernstein2:
ComputeEngine metadata server has its own token caching mechanism that will return a cached token
until the last 5 minutes of its expiration. This has a negative interaction with stale token refreshes because 
stale token refresh kicks in T-6mins until T-5mins. This will cause every stale refresh to return the same 
stale token.

This PR updates the timing for ComputeEngineCredentials to start a stale refresh at T-4mins and consider 
the token expired at T-3 mins. The implementation is a bit noisy because it includes a change OAuth2Credentials 
to make the thresholds configureable and now that we targeting java8, I migrated to using java8 time data types.

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
@@ -50,6 +50,7 @@ import java.io.InputStream;
 import java.io.ObjectInputStream;
 import java.net.SocketTimeoutException;
 import java.net.UnknownHostException;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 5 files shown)


oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -71,6 +72,9 @@
 public class ComputeEngineCredentials extends GoogleCredentials
     implements ServiceAccountSigner, IdTokenProvider {
```
silvolu, line 74: Could you add some comments on why we had to change the margins specifically for GCE?

Neenu1995 (approved):

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
igorbernstein2: done

TimurSadykov (approved):
LGTM

PR #750, opened 2021-09-28
feat: update retries and implement Retryable

### LINKED ISSUE #626: RequestMetadataCallback.onFailure() should provide a type-safe method to retrieve the root cause of the failure
This caused issue: [googleapis/gax-java#965](https://github.com/googleapis/gax-java/issues/965) which is being investigated by grpc team here: https://github.com/grpc/grpc-java/issues/6808

## Description
The following is what happens.

1. If refresh token is invalid, then auth lib will throw an `HttpResponseException` like the following

    ```
    com.google.api.client.http.HttpResponseException: 400 Bad Request
    POST https://oauth2.googleapis.com/token
    {
      "error": "invalid_request",
      "error_description": "Could not determine client ID from request."
    }
    ```

1. [auth lib then passes the exception](https://github.com/googleapis/google-auth-library-java/blob/master/credentials/java/com/google/auth/Credentials.java#L114) to a `callback`

    ```
    protected final void blockingGetToCallback(URI uri, RequestMetadataCallback callback) {
        Map<String, List<String>result;
        try {
          result = getRequestMetadata(uri);
        } catch (Throwable e) {
          callback.onFailure(e);   <========= here
          return;
        }
        callback.onSuccess(result);
    }
    ```

1. callback is handled by the following code in grpc-auth [GoogleAuthLibraryCallCredentials.java](https://github.com/grpc/grpc-java/blob/master/auth/src/main/java/io/grpc/auth/GoogleAuthLibraryCallCredentials.java#L138)

    ```
    public void onFailure(Throwable e) {
        if (e instanceof IOException) {         <=============== here
          // Since it's an I/O failure, let the call be retried with UNAVAILABLE.
          applier.fail(Status.UNAVAILABLE
              .withDescription("Credentials failed to obtain metadata")
              .withCause(e));
        } else {
          applier.fail(Status.UNAUTHENTICATED
              .withDescription("Failed computing credential metadata")
              .withCause(e));
        }
    }
    ```

Note that `HttpResponseException` is subclass of `IOException`, but we shouldn't retry on `HttpResponseException`. So the correct implementation would be:

```
public void onFailure(Throwable e) {
        if ((e instanceof IOException) && !(e instanceof HttpResponseException)) {         <=============== here
          // Since it's an I/O failure, let the call be retried with UNAVAILABLE.
          applier.fail(Status.UNAVAILABLE
              .withDescription("Credentials failed to obtain metadata")
              .withCause(e));
        } else {
          applier.fail(Status.UNAUTHENTICATED
              .withDescription("Failed computing credential metadata")
              .withCause(e));
        }
      }
```

**Describe the solution you'd like**
`HttpResponseException` is an implementation detail of google-auth-library-java. We'd like a typed exception returned here with properties for the error context.

```
public class AuthenticationException {
  public AuthErrorCode getError() {
    // returns AuthErrorCode.INVALID_REQUEST for the example above.
  }

  public String getErrorDescription() {
    // returns "Could not determine client ID from request." for the example above.
  }
}

public enum AuthErrorCode {
  UNAVAILABLE,
  UNAUTHENTICATED,
  INVALID_REQUEST
  ...
}
```

**Describe alternatives you've considered**
gRPC could modify the exception handling to check if the exception is a `HttpResponseException`. However, this is an implementation detail of google-auth-library-java and potentially not a reliable indicator that the request has failed permanently.

TimurSadykov:
This is PR fixes retries to token endpoint.
For more details on the feature: go/auth-correct-retry

Fixes #626


oauth2_http/pom.xml
```diff
@@ -129,10 +129,12 @@
     <dependency>
       <groupId>com.google.http-client</groupId>
       <artifactId>google-http-client</artifactId>
+      <version>1.40.1-SNAPSHOT</version>
```
TimurSadykov, line 132: temporary for POC

credentials/pom.xml
```diff
@@ -54,6 +54,16 @@
...
     </dependency>
+      <dependency>
+        <groupId>com.google.api-client</groupId>
+        <artifactId>google-api-client</artifactId>
+        <version>1.32.1</version>
+      </dependency>
+      <dependency>
+        <groupId>com.google.http-client</groupId>
+        <artifactId>google-http-client</artifactId>
+        <version>1.40.1-SNAPSHOT</version>
```
TimurSadykov, line 65: temporary for POC

arithmetic1728 (approved):

arithmetic1728 (approved):

oauth2_http/java/com/google/auth/oauth2/GoogleAuthException.java
```diff
@@ -0,0 +1,157 @@
...
+   * Constructor with all parameters
+   *
+   * @param isRetryable A retry status for the related HTTP request
+   * @param retryCount A number of retries performed for the related HTTP request
+   * @param message The detail message (which is saved for later retrieval by the {@link
+   *     #getMessage()} method)
+   * @param cause The cause (which is saved for later retrieval by the {@link #getCause()} method).
+   *     (A null value is permitted, and indicates that the cause is nonexistent or unknown.)
+   */
+  public GoogleAuthException(boolean isRetryable, int retryCount, String message, Throwable cause) {
```
chingor13, line 57: Please avoid public constructors - they cause extra surface area changes. Instead, opt for a builder.

oauth2_http/java/com/google/auth/oauth2/GoogleAuthException.java
```diff
@@ -0,0 +1,157 @@
...
+   * Creates an instance of the exception based on {@link HttpResponseException} and a custom error
+   * message.
+   *
+   * @see #createWithTokenEndpointResponseException(HttpResponseException, String)
+   * @param responseException an instance of {@link HttpResponseException}
+   * @param message The detail message (which is saved for later retrieval by the {@link
+   *     #getMessage()} method)
+   * @return new instance of {@link GoogleAuthException}
+   */
+  public static GoogleAuthException createWithTokenEndpointResponseException(
```
chingor13, line 119: Does this need to be public?

oauth2_http/java/com/google/auth/oauth2/GoogleAuthException.java
```diff
@@ -0,0 +1,157 @@
...
+
+  /**
+   * Creates an instance of the exception based on {@link HttpResponseException} returned by Google
+   * token endpoint. It uses response status code information to populate the {@code #isRetryable}
+   * property and a number of performed attempts to populate the {@code #retryCount} property
+   *
+   * @param responseException an instance of {@link HttpResponseException}
+   * @return new instance of {@link GoogleAuthException}
+   */
+  public static GoogleAuthException createWithTokenEndpointResponseException(
```
chingor13, line 141: Does this need to be public?

credentials/java/com/google/auth/Retryable.java
```diff
@@ -0,0 +1,49 @@
...
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package com.google.auth;
+
+// an interface to identify retryable errors
+public interface Retryable {
```
chingor13, line 35: I don't see how this interface is adding value unless the caller is checking `implements` vs `instanceof` in the catch block.

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -278,6 +277,17 @@ public GoogleCredentials createScoped(String... scopes) {
...
   }
 
+  /**
+   * If the credentials support automatic retries, creates a copy of the identity with the provided
+   * retry strategy
+   *
+   * @param defaultRetriesEnabled a flag enabling or disabling default retries
+   * @return GoogleCredentials with the new default retries configuration.
+   */
+  public GoogleCredentials createWithCustomRetryStrategy(boolean defaultRetriesEnabled) {
```
chingor13, line 287: This method name is awkward with the boolean value. I would expect to be providing a retry strategy as the parameter from reading the method name.

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -576,36 +592,42 @@ public AccessToken refreshAccessToken() throws IOException {
 
     HttpRequestFactory requestFactory = transportFactory.create().createRequestFactory();
     HttpRequest request = requestFactory.buildPostRequest(new GenericUrl(tokenServerUri), content);
+
+    if (this.defaultRetriesEnabled) {
```
chingor13, line 596: Does it make sense to just allow configuring the number of retries and using the default if no configured value is set?

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -485,7 +500,8 @@ static ServiceAccountCredentials fromPkcs8(
         projectId,
         quotaProject,
         DEFAULT_LIFETIME_IN_SECONDS,
-        false);
+        false,
+        true);
```
lesv, line 504: Since you modified it, it really is too long to not use a builder here.

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -692,7 +725,8 @@ public GoogleCredentials createScoped(
         projectId,
         quotaProjectId,
         lifetime,
-        useJwtAccessWithScope);
+        useJwtAccessWithScope,
+        defaultRetriesEnabled);
```
lesv, line 729: needs a builder here.

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -733,7 +767,8 @@ public GoogleCredentials createDelegated(String user) {
...
```
lesv, line 771: and here

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -1219,7 +1265,8 @@ public ServiceAccountCredentials build() {
           projectId,
           quotaProjectId,
           lifetime,
-          useJwtAccessWithScope);
+          useJwtAccessWithScope,
+          defaultRetriesEnabled);
```
lesv, line 1269: and here.  (Hopefully I'll actually get to `ServiceAccountCredentials()` to give a better comment.

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -149,7 +162,8 @@
       String projectId,
       String quotaProjectId,
       int lifetime,
-      boolean useJwtAccessWithScope) {
+      boolean useJwtAccessWithScope,
+      boolean defaultRetriesEnabled) {
```
lesv, line 166: This really needs to be done w/ a builder.  Way too many params.

oauth2_http/java/com/google/auth/oauth2/GoogleAuthException.java
TimurSadykov: This exception was in the Credentials package in previous iteration, therefore ended up public. Made exception itself private but not the rest ) Thanks!

credentials/java/com/google/auth/Retryable.java
TimurSadykov: This is to be leveraged in gRPC code and custom clients that we control once we let them know that their retries most likely: 
1. incorrect
2. duplicate

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
TimurSadykov: As explained in the doc - we don't want to enhance retry functionality. Instead, we want to disable it for as many clients as possible.

Therefore we keep retries and fix them to the spec. But if someone wants custom retry strategy - guidance is to disable library retries and retry on client level. That guidance will be published as AIP, now on review.

lesv (changes requested):
I need to spend a bit more time w/ your testing, but I'm generally favorable.  If I didn't comment on Jeff's comment's, I probably agree with them.  Even were I disagree, I'll defer to him if he insists.

`ServiceAccountCredentials` constructor has way too many parameters and should be made into a builder pattern - especially since your breaking any contract anyway.

I want to spend a bit more time thinking through your tests, but I believe you've implemented what I read in the design docs.

credentials/java/com/google/auth/Retryable.java
chingor13: Would you not just catch the GoogleAuthException directly rather than a generic IOException and then check whether it  implements Retryable?

oauth2_http/java/com/google/auth/oauth2/GoogleAuthException.java
TimurSadykov: nope, removed

oauth2_http/java/com/google/auth/oauth2/GoogleAuthException.java
TimurSadykov: nope, leftover, thanks!

credentials/java/com/google/auth/Retryable.java
TimurSadykov: IOException + Retryable is preferable at least for gRPC because Retryable is part of Credentials that gRPC mostly use. They don't want to import specifics with oauth2 package.

oauth2_http/java/com/google/auth/oauth2/GoogleAuthException.java
TimurSadykov: Removed public, keeping local constructor.

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
TimurSadykov: It is not perfect, but still fits. I tried to come up with something better but did not :) If you have suggestion - please let me know. Also, this interface is part of proposed AIP: [go/aip4117](https://goto.google.com/aip4117)

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -111,66 +115,58 @@
...
-   *     authority to the service account.
-   * @param projectId the project used for billing
-   * @param quotaProjectId the project used for quota and billing purposes. May be null.
-   * @param lifetime number of seconds the access token should be valid for. The value should be at
-   *     most 43200 (12 hours). If the token is used for calling a Google API, then the value should
-   *     be at most 3600 (1 hour). If the given value is 0, then the default value 3600 will be used
-   *     when creating the credentials.
-   * @param useJwtAccessWithScope whether self signed JWT with scopes should be always used.
+   * @param builder A builder for {@link ServiceAccountCredentials} See {@link
+   *     ServiceAccountCredentials.Builder} List of fields being set: clientId: client ID of the
```
TimurSadykov, line 126: will move to the builder

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
lesv: `enableRetries()` ??

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -111,66 +115,58 @@
...
-      String projectId,
-      String quotaProjectId,
-      int lifetime,
-      boolean useJwtAccessWithScope) {
-    this.clientId = clientId;
-    this.clientEmail = Preconditions.checkNotNull(clientEmail);
-    this.privateKey = Preconditions.checkNotNull(privateKey);
-    this.privateKeyId = privateKeyId;
-    this.scopes = (scopes == null) ? ImmutableSet.<String>of() : ImmutableSet.copyOf(scopes);
+  ServiceAccountCredentials(ServiceAccountCredentials.Builder builder) {
```
lesv, line 143: I responded yes when you asked about this yesterday from my phone.  And I'm ok w/ it if you insist, but really the `.build()` method should return a `ServiceAccountCredential`.  Your choice. see go/ej3e-2 (final few paragraphs)

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
TimurSadykov: you mean createWithDefaultRetriesEnabled?

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
TimurSadykov: fixed

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
TimurSadykov: fixed

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
TimurSadykov: yep, fixed

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
TimurSadykov: fixed

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
TimurSadykov: fixed

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -276,18 +267,14 @@ public static ServiceAccountCredentials fromPkcs8(
...
-        null,
-        null,
-        null,
-        null,
-        null);
+    ServiceAccountCredentials.Builder builder =
+        ServiceAccountCredentials.newBuilder()
+            .setClientId(clientId)
+            .setClientEmail(clientEmail)
+            .setPrivateKeyId(privateKeyId)
```
lesv, line 274: Looks like it should be `privateKeyPkcs8`

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -242,18 +237,14 @@ public static ServiceAccountCredentials fromPkcs8(
...
```
lesv, line 244: Looks like it should be `privateKeyPkcs8`

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -209,18 +205,17 @@ static ServiceAccountCredentials fromJson(
...
-        transportFactory,
-        tokenServerUriFromCreds,
-        null,
-        projectId,
-        quotaProjectId);
+    ServiceAccountCredentials.Builder builder =
+        ServiceAccountCredentials.newBuilder()
+            .setClientId(clientId)
+            .setClientEmail(clientEmail)
+            .setPrivateKeyId(privateKeyId)
```
lesv, line 212: `privateKeyPkcs8`

lesv (changes requested):
I'm sending this early as I'm going to get lunch.

I found some stuff that you will want to change.  I'm suggesting change to the form of the builder pattern that I ok'd yesterday - apologies - I should n't type on my phone.  I'm ok if you don't want to do it.   I'm unattached to my suggestions on names.

I'll pick up where I left off in an hour or so.

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
lesv: Ah, your just passing it to the constructor directly.

lesv (approved):
Ignore my earlier comments.  LGTM

arithmetic1728 (approved):

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
TimurSadykov: kept the name as is

credentials/java/com/google/auth/Retryable.java
```diff
@@ -0,0 +1,49 @@
...
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package com.google.auth;
+
+// an interface to identify retryable errors
```
ejona86, line 34: `/** Retryable is an interface to identify retryable errors. */`

ejona86 (commented):
drive-by comment

PR #754, opened 2021-10-01
fix: disabling self-signed jwt for domain wide delegation

TimurSadykov:
The self-signed JWT is not supported with domain delegation. Therefore disabling the self-signed JWT in that case.

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
@@ -1018,8 +1018,8 @@ public class ServiceAccountCredentials extends GoogleCredentials
 
     // If scopes are provided but we cannot use self signed JWT, then use scopes to get access
     // token.
-    if (!createScopedRequired() && !useJwtAccessWithScope
-        || serviceAccountUser != null && serviceAccountUser.length() > 0) {
+    if ((!createScopedRequired() && !useJwtAccessWithScope)
+        || (serviceAccountUser != null && serviceAccountUser.length() > 0)) {
       return super.getRequestMetadata(uri);
     }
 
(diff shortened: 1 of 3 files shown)


oauth2_http/javatests/com/google/auth/oauth2/TokenVerifierTest.java
```diff
@@ -242,8 +242,6 @@ public void verifyRs256TokenWithLegacyCertificateUrlFormat()
   @Test
   public void verifyServiceAccountRs256Token()
       throws TokenVerifier.VerificationException, IOException {
-    HttpTransportFactory httpTransportFactory =
```
TimurSadykov, line 245: just some unused var removed

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -1018,7 +1018,8 @@ public void getRequestMetadata(
 
     // If scopes are provided but we cannot use self signed JWT, then use scopes to get access
     // token.
-    if (!createScopedRequired() && !useJwtAccessWithScope) {
+    if (!createScopedRequired() && !useJwtAccessWithScope
+        || serviceAccountUser != null && serviceAccountUser.length() > 0) {
```
arithmetic1728, line 1022: somehow I think the logic would be
```
// We don't use self signed JWT approach if `useJwtAccessWithScope` is false, or domain wide
// delegation is used.
if (!useJwtAccessWithScope || (serviceAccountUser != null && serviceAccountUser.length() > 0)) {
    return super.getRequestMetadata(uri);
}
```
Most likely I wrote this code, but I don't remember why `!createScopedRequired()` is here, lol

arithmetic1728 (approved):

Neenu1995 (approved):

Issue #756, opened 2021-10-01
Migrate from JUnit4 to JUnit5


Captain1653:
The library uses JUnit4 . JUnit5 is stable. Is it worth migrating? I can do it. It's interesting for me.

It can be something like that:
- update maven-surfire-plugin (if necessary)
- update junit version
- run old tests with vintage junit runner
- refactor old tests step by step

Kind regards

suztomo:
HI @Captain1653 . Thank you for filling the issue. Just curious. Is this (this repository's using JUnit4) blocking some of your activity?

Captain1653:
No, it doesn't block my activity. It's only my opinion and I can do it. Does it make sense to start ?

Capstan:
This renders this package incapable of being imported into Google's own monorepo due to trailing support for JUnit5 in bazel Java.

PR #760, opened 2021-10-08
Migration tests in appengine module from JUnit4 to JUnit5 (#756)

Captain1653:
It belongs to #756  . It's the first iteration. That's why I don't use `fixed` notation for the issue.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -61,6 +61,7 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.google.http.version>1.40.1</project.google.http.version>
     <project.junit.version>4.13.2</project.junit.version>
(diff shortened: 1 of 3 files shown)


Captain1653:
@chanseokoh Thank you for your advice. Will my commit be squashed? Right?

pom.xml
```diff
@@ -61,6 +61,7 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.google.http.version>1.40.1</project.google.http.version>
     <project.junit.version>4.13.2</project.junit.version>
```
chanseokoh, line 63: Don't forget to remove JUnit 4 in a follow-up when it's no longer needed!

PR #770, opened 2021-10-13
fix: service account impersonation with workforce credentials

lsirac:
While service account impersonation is not commonly used with workforce
pool configurations, there is a bug where the workforcePoolUserProject is not being set on the source credential.

This also lets us align with other languages who added the workforcePoolUserProject to the base class, instead of IdentityPoolCredentials (though it is the only one that currently supports it).

The bug itself is fixed by the refactor. By moving workforcePoolUserProject to be set in the base constructor, it is set before initializeImpersonatedCredentials() is called. A copy of the source credential is then made that has the workforcePoolUserProject set. Prior to this it was not set and the impersonation call resulted in a 403.

### DIFF
--- a/oauth2_http/pom.xml
+++ b/oauth2_http/pom.xml
@@ -61,7 +61,7 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-failsafe-plugin</artifactId>
-        <version>3.0.0-M4</version>
+        <version>3.0.0-M5</version>
         <configuration>
           <forkedProcessTimeoutInSeconds>1200</forkedProcessTimeoutInSeconds>
           <reportNameSuffix>sponge_log</reportNameSuffix>
@@ -70,13 +70,6 @@
             <include>**/functional/*.java</include>
           </includes>
         </configuration>
-        <dependencies>
-          <dependency>
-            <groupId>org.apache.maven.surefire</groupId>
-            <artifactId>surefire-junit47</artifactId>
-            <version>3.0.0-M4</version>
-          </dependency>
-        </dependencies>
       </plugin>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>


TimurSadykov (commented):
Some general comments for now. Can you please highlight the bugfix itself? Otherwise refactoring hides it

oauth2_http/javatests/com/google/auth/oauth2/ExternalAccountCredentialsTest.java
```diff
@@ -400,18 +540,15 @@ void exchangeExternalCredentialForAccessToken_throws() throws IOException {
...
-            "subjectTokenType",
-            STS_URL,
-            new TestCredentialSource(new HashMap<String, Object>()),
-            "tokenInfoUrl",
-            /* serviceAccountImpersonationUrl= */ null,
-            "quotaProjectId",
-            /* clientId= */ null,
-            /* clientSecret= */ null,
-            /* scopes= */ null);
+        (TestExternalAccountCredentials)
```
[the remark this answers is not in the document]
lsirac: Yep, the ExternalAccountCredentials builder is abstract.

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
```diff
@@ -155,37 +154,15 @@ private boolean hasHeaders() {
...
-    this.workforcePoolUserProject = builder.workforcePoolUserProject;
-
-    if (workforcePoolUserProject != null && !isWorkforcePoolConfiguration()) {
-      throw new IllegalArgumentException(
-          "The workforce_pool_user_project parameter should only be provided for a Workforce Pool configuration.");
-    }
   }
 
   @Nullable
   public String getWorkforcePoolUserProject() {
```
TimurSadykov, line 164: remove this exposing parent method directly

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
```diff
@@ -304,17 +261,14 @@ public static Builder newBuilder(IdentityPoolCredentials identityPoolCredentials
...
-    @Nullable private String workforcePoolUserProject;
-
     Builder() {}
 
     Builder(IdentityPoolCredentials credentials) {
       super(credentials);
-      setWorkforcePoolUserProject(credentials.getWorkforcePoolUserProject());
     }
 
     public Builder setWorkforcePoolUserProject(String workforcePoolUserProject) {
```
TimurSadykov, line 270: remove this as well?

TimurSadykov (changes requested):
Looks good, couple small comments

PR #789, opened 2021-11-07
chore: java format update

TimurSadykov:


BenWhitehead:
v1.12.0 of google-java-format requires a minimum of java 11, all of our build tooling currently has a minimum of java 8 so we can't merge this.

TimurSadykov:
@BenWhitehead Yes, that was a purpose of the PR (it is incomplete), to switch lint to jre11. Effectively, the bump is only for linter, it does not affect anything else. It still validates older Java code, just JRE has to be 11 per: https://github.com/google/google-java-format/issues/529

BenWhitehead:
In that case, you'll likely need to update the base template for all java project in the synthtool repo here https://github.com/googleapis/synthtool/blob/master/synthtool/gcp/templates/java_library/.github/workflows/ci.yaml#L49

PR #816, opened 2021-12-29
chore: downstream check for all libraries in single job

Neenu1995:
Test core java libraries in java client libraries before release. Feature/code changes introduced in each PR can be tested on downstream java client libraries by adding `downstream-check:run` label. For now, unit tests are being run.

Next phase involves running samples and integration tests.

Note: Some of these tests are expected to fail due to individual client library conditions and do block the merging of this PR.


.kokoro/downstream-client-library-check.sh
```diff
@@ -0,0 +1,123 @@
...
+REPO=$1
+## Get the directory of the build script
+scriptDir=$(realpath $(dirname "${BASH_SOURCE[0]}"))
+## cd to the parent directory, i.e. the root of the git repo
+cd ${scriptDir}/..
+
+# Make java-core artifacts available for 'mvn validate' at the bottom
+mvn install -DskipTests=true -Dmaven.javadoc.skip=true -Dgcloud.download.skip=true -B -V -q
+
+# Read the current version of this java-core in the POM. Example version: '0.116.1-alpha-SNAPSHOT'
```
suztomo, line 33: I guess this is not java-core any more.

.kokoro/downstream-client-library-check.sh
```diff
@@ -0,0 +1,123 @@
...
+AUTH_VERSION_POM=pom.xml
+# Namespace (xmlns) prevents xmllint from specifying tag names in XPath
+AUTH_VERSION=`sed -e 's/xmlns=".*"//' ${AUTH_VERSION_POM} | xmllint --xpath '/project/version/text()' -`
```
suztomo, line 36: As of now, I see this script is specific to auth library (from the variable name "AUTH_VERSION)). But I believe this script is beneficial to Google HTTP client, GAX, and all other libraries in the shared dependencies BOM. What do you think about making it generic?

suztomo (commented):
It's nice to see the 132 checks ran.

.kokoro/downstream-client-library-check.sh
Neenu1995: I will update it to be more generic.

.kokoro/downstream-client-library-check.sh
Neenu1995: Fixed

lesv (approved):

lesv (approved):

PR #868, opened 2022-02-23
fix: update library docs

### LINKED ISSUE #836: Library description does not provide pointers to key terms
Current description is hard to navigate and missing some useful description. We want to improve that to help users to have a chance to self-discover actual use-case for the library

1. Short descriptions for the main concepts like Application Default Credentials. This should provide better value than enumeration of classes right now
2. Missing table of contents, impossible to navigate effectively 
3. Suboptimal structure, where most important artefact description is hidden in the middle of the doc
4. CI links are outdated
5. Still says version is unstable

TimurSadykov:
1. Adding table of contents
2. Adding description of ADC with links instead of class enumeration
3. Switching status to stable
4. remove java7 CI link

Fixes #836  ☕️


README.md
```diff
@@ -53,64 +70,27 @@ libraryDependencies += "com.google.auth" % "google-auth-library-oauth2-http" % "
...
-    AppEngineCredentials.newBuilder()
-        .setScopes(...)
-        .setAppIdentityService(appIdentityService)
-        .build();
-```
+### Application Default Credentials
 
-**Important: `com.google.auth.appengine.AppEngineCredentials` is a separate class from
-`com.google.auth.oauth2.AppEngineCredentials`.**
+This library provides an implementation of [Application Default Credentials](https://cloud.google.com/docs/authentication/getting-started)
```
lesv, line 77: ADC isn't mentioned on that page.  https://google.aip.dev/auth/4110 explains it at length.

I also asked for repairs to that page.

lesv: You might also want to mention https://google.aip.dev/auth

lesv (approved):
a comment for the readme

README.md
TimurSadykov: yep, thanks, originally I did not like the link either but forgot about it :)

PR #893, opened 2022-03-16
fix: update branding in ExternalAccountCredentials

hegemonic:
These changes align the Javadoc comments with the branding that Google uses externally:

+ STS -> Security Token Service
+ GCP -> Google Cloud
+ Remove references to a Google-internal token type

- [ ] Ensure the tests and linter pass: Tests are failing, but I don't think that was caused by the changes in this PR
- [ ] Code coverage does not decrease (if any source code was changed): n/a
- [ ] Appropriate docs were updated (if necessary): n/a


TimurSadykov (approved):
LGTM

Neenu1995 (approved):

PR #901, opened 2022-04-01
test: test LTS compatibility

Neenu1995:
http:1.41.5
guava:31.1-android
autovalue:1.9
auto-value-annotation:1.9


TimurSadykov:
@Neenu1995 i think this is no longer needed

PR #908, opened 2022-04-07
feat: Adds Pluggable Auth support (WIF)

lsirac:
See go/pluggable-auth-design. 

Summary:
- Adds a new PluggableAuthCredentials class that enables the ability for developers to use their own executable to retrieve subject tokens for Workload Identity Federation (instead of a file/URL).
- ADC is extended to support these new credentials. 
- Adds a handler that deals with calling the executables and parsing their response in the expected format. 
- Documentation/changes to the README will be provided in a separate PR.


Neenu1995 (approved):

oauth2_http/java/com/google/auth/oauth2/PluggableAuthHandler.java
```diff
@@ -0,0 +1,260 @@
...
+    abstract InternalProcessBuilder redirectErrorStream(boolean redirectErrorStream);
+
+    abstract Process start() throws IOException;
+  }
+
+  /**
+   * The default implementation that wraps {@link ProcessBuilder} for creating and managing a
+   * process.
+   */
+  static final class DefaultProcessBuilder extends InternalProcessBuilder {
```
TimurSadykov, line 72: What is the value of this class over using ProcessBuilder directly?

TimurSadykov (changes requested):
Most importantly - lets explore how to keep the impersonatedCredentials final. 
Tests review in progress meanwhile.

oauth2_http/java/com/google/auth/oauth2/PluggableAuthHandler.java
lsirac: I will answer this in the comment above as the answer is the same. Resolving.

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
```diff
@@ -97,7 +99,7 @@ abstract static class CredentialSource {
 
   protected transient HttpTransportFactory transportFactory;
 
-  @Nullable protected final ImpersonatedCredentials impersonatedCredentials;
+  @Nullable protected ImpersonatedCredentials impersonatedCredentials;
```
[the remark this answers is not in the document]
lsirac: Since it is being called in the base class constructor, the underlying credentials are being initialized before the handler is being set. We need to call the method again after setting the handler in the subclass.

oauth2_http/java/com/google/auth/oauth2/PluggableAuthHandler.java
```diff
@@ -0,0 +1,260 @@
...
+/**
+ * Internal handler for retrieving 3rd party tokens from user defined scripts/executables for
+ * workload identity federation.
+ *
+ * <p>See {@link PluggableAuthCredentials}.
+ */
+final class PluggableAuthHandler implements ExecutableHandler {
+
+  /** An interface for creating and managing a process. */
+  abstract static class InternalProcessBuilder {
```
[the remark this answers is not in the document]
lsirac: ProcessBuilder is final (so it can't be mocked) and it does not implement any interface. Creating an abstract class and allowing concrete implementations to be specified lets us test these changes. 

DefaultProcessBuilder is the implementation we use that wraps the ProcessBuilder methods.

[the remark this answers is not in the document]
TimurSadykov: Can you document that a bit more explicitly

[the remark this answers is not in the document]
TimurSadykov: Also, please move it below, since it mostly infra stuff

pom.xml
```diff
@@ -169,6 +169,7 @@
           <version>3.4.0</version>
           <configuration>
             <source>7</source>
```
TimurSadykov, line 171: why this?

oauth2_http/java/com/google/auth/oauth2/ExecutableResponse.java
```diff
@@ -0,0 +1,206 @@
...
+        this.subjectToken = (String) json.get("id_token");
+      }
+      if (subjectToken == null || subjectToken.isEmpty()) {
+        throw new PluggableAuthException(
+            "INVALID_EXECUTABLE_RESPONSE",
+            "The executable response does not contain a valid token.");
+      }
+    } else {
+      // Error response must contain both an error code and message.
+      this.errorCode = (String) json.get("code");
```
TimurSadykov, line 99: we want to try-catch and surface it if otherwise?

TimurSadykov (approved):
Minor comment, LGTM

pom.xml
lsirac: This isn't part of this PR?

PR #927, opened 2022-06-08
docs(samples): added auth samples and tests

Sita04:
Adding samples for the authentication docs.


Chizzy09 (approved):

Chizzy09:
45449b220588773b03c70fcfc287d92e708972e9

samples/snippets/src/main/java/IdTokenFromImpersonatedCredentials.java
```diff
@@ -0,0 +1,159 @@
...
+    // depending on the level of access you need.
+    // Example: The following scope lets you view and manage Pub/Sub topics and subscriptions.
+    // For more information, see: https://developers.google.com/identity/protocols/oauth2/scopes
+    String scope = "https://www.googleapis.com/auth/pubsub";
+
+    // The service name for which the id token is requested. Service name refers to the
+    // logical identifier of an API service, such as "pubsub.googleapis.com".
+    String targetAudience = "pubsub.googleapis.com";
+
+    // The service account name of the limited-privilege account for whom the credential is created.
```
piaxc, line 55: I am in the process of changing the names of the service accounts used in this flow. But my changes aren't approved yet, so maybe hold off until they are? CL/452139026, I cc'ed you on the CL.

samples/snippets/src/main/java/IdTokenFromImpersonatedCredentials.java
```diff
@@ -0,0 +1,159 @@
...
+public class IdTokenFromImpersonatedCredentials {
+
+  public static void main(String[] args) throws IOException, GeneralSecurityException {
+    // TODO(Developer): Replace the below variables before running the code.
+
+    // Your Google Cloud project id.
+    String projectId = "your-google-cloud-project-id";
+
+    // Path to the service account json credential file.
+    String jsonCredentialPath = "path-to-json-credential-file";
```
piaxc, line 43: There shouldn't be any service account key path involved in this flow.

samples/snippets/src/main/java/IdTokenFromImpersonatedCredentials.java
```diff
@@ -0,0 +1,159 @@
...
+    // Initialize the IAMCredentials service with the source credential and scope.
+    IAMCredentials service = null;
+    try {
+      service = initService(jsonCredentialPath, scope);
+    } catch (IOException | GeneralSecurityException e) {
+      System.out.println("Unable to initialize service: \n" + e);
+      return;
+    }
+
+    // delegates: The chained list of delegates required to grant the final accessToken.
```
piaxc, line 79: I've also separated out the delegation flow from the single service account flow. It doesn't make sense to create an ID token from a delegation chain, because there's no authz included for ID tokens. So let's just stick to a single privilege-bearing (current term, subject to change) service account (no delegation chain).

samples/snippets/src/main/java/IdTokenFromImpersonatedCredentials.java
```diff
@@ -0,0 +1,159 @@
...
+    GenerateIdToken idToken =
+        service
+            .projects()
+            .serviceAccounts()
+            .generateIdToken(
+                String.format(
+                    "projects/%s/serviceAccounts/%s", projectId, impersonatedServiceAccount),
+                idTokenRequest);
+
+    // Verify the obtained id token. This is done at the receiving end of the OIDC endpoint.
```
piaxc, line 109: I won't be using this the way the page is currently structured. Please exclude when you add region tags.

samples/snippets/src/main/java/IdTokenFromImpersonatedCredentialsREST.java
```diff
@@ -0,0 +1,167 @@
...
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+public class IdTokenFromImpersonatedCredentialsREST {
+
+  public static void main(String[] args) throws IOException, GeneralSecurityException {
+    // TODO(Developer): Replace the below variables before running the code.
+    // Path to the service account json credential file.
+    String jsonCredentialPath = "path-to-json-credential-file";
```
piaxc, line 44: Again, not sure how the service account key fits in here...

samples/snippets/src/main/java/IdTokenFromMetadataServer.java
```diff
@@ -0,0 +1,128 @@
...
+import java.util.Collections;
+
+public class IdTokenFromMetadataServer {
+
+  public static void main(String[] args) throws IOException, GeneralSecurityException {
+    // TODO(Developer): Replace the below variables before running the code.
+
+    // The service name for which the id token is requested. Service name refers to the
+    // logical identifier of an API service, such as "pubsub.googleapis.com".
+    String targetAudience = "pubsub.googleapis.com";
```
piaxc, line 43: Pubsub does not require an id token. This should be something that the developer fills in, and it should be a url, not a service name.

samples/snippets/src/main/java/IdTokenFromMetadataServer.java
```diff
@@ -0,0 +1,128 @@
...
+
+    getIdTokenFromMetadataServer(targetAudience);
+  }
+
+  // Every VM stores its metadata on a metadata server. You can query for default VM metadata,
+  // such as the VM's host name, instance ID, and service account information programmatically
+  // from within a VM.
+  // Here, we query the service account information from the Metadata server exposed by the
+  // ComputeEngine and use that information to obtain an id token.
+  // Appengine 2nd Generation, Cloud Run or even Kubernetes engine's also expose a
```
piaxc, line 53: Again, let's keep the documentation in the docs. It's easier to keep up to date and there's no issue with diverging stories.

samples/snippets/src/main/java/IdTokenFromMetadataServer.java
```diff
@@ -0,0 +1,128 @@
...
+  // metadata server.
+  // For AppEngine, see:
+  // https://cloud.google.com/appengine/docs/standard/java/accessing-instance-metadata#identifying_which_metadata_endpoint_to_use
+  // For CloudRun container instance, see:
+  // https://cloud.google.com/run/docs/container-contract#metadata-server
+  public static void getIdTokenFromMetadataServer(String targetAudience)
+      throws GeneralSecurityException, IOException {
+
+    // Optionally, you can also set scopes in computeEngineCredentials.
+    ComputeEngineCredentials computeEngineCredentials = ComputeEngineCredentials.create();
```
piaxc, line 63: Please use ADC for authn for this flow.

samples/snippets/src/main/java/IdTokenFromServiceAccount.java
```diff
@@ -0,0 +1,106 @@
...
+import java.util.List;
+import java.util.concurrent.ExecutionException;
+
+public class IdTokenFromServiceAccount {
+
+  public static void main(String[] args)
+      throws IOException, ExecutionException, InterruptedException, GeneralSecurityException {
+    // TODO(Developer): Replace the below variables before running the code.
+    // Path to the service account json credential file.
+    String jsonCredentialPath = "path-to-json-credential-file";
```
piaxc, line 39: Can't use this snippet, need not to use service account key.

piaxc (commented):
Hi Sita, feel free to grab a slot on my calendar if you want to discuss any of my comments!

samples/snippets/src/main/java/AuthenticateExplicit.java
```diff
@@ -0,0 +1,89 @@
...
+import com.google.auth.oauth2.GoogleCredentials;
+import com.google.cloud.storage.Bucket;
+import com.google.cloud.storage.Storage;
+import com.google.cloud.storage.StorageOptions;
+import com.google.common.collect.Lists;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.security.GeneralSecurityException;
+
+public class AuthenticateExplicit {
```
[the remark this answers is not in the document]
Sita04: You're right. I thought of obtaining the review before adding the region tags.

samples/snippets/src/main/java/AuthenticateExplicit.java
```diff
@@ -0,0 +1,89 @@
...
+    String projectId = "your-google-cloud-project-id";
+
+    // Path to the service account json credential file.
+    String jsonCredentialPath = "path-to-json-credential-file";
+
+    // Provide the scopes that you might need to request to access Google APIs,
+    // depending on the level of access you need.
+    // Example: The following scope lets you view and manage Pub/Sub topics and subscriptions.
+    // For more information, see: https://developers.google.com/identity/protocols/oauth2/scopes
+    String scope = "https://www.googleapis.com/auth/devstorage.full_control";
```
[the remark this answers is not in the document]
Sita04: Modified it. Thanks.

samples/snippets/src/main/java/AuthenticateImplicitWithAdc.java
```diff
@@ -0,0 +1,80 @@
...
+ */
+
+import com.google.cloud.compute.v1.Instance;
+import com.google.cloud.compute.v1.InstancesClient;
+import java.io.IOException;
+
+public class AuthenticateImplicitWithAdc {
+
+  public static void main(String[] args) throws IOException {
+    // TODO(Developer):
```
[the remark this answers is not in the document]
Sita04: PTAL at the refactored approach and comments.

samples/snippets/src/main/java/IdTokenFromImpersonatedCredentialsREST.java
Sita04: Refactored the code. PTAL.

samples/snippets/src/main/java/AuthenticateImplicitWithAdc.java
```diff
@@ -0,0 +1,80 @@
...
+  // 2. Explicit authentication by specifying the service account
+  // 3. Authentication with service account credentials obtained from metadata server, like,
+  // Compute Engine or App Engine etc.,
+  // 4. Bring your own (BYO) access token
+  // 5. Using API keys (for libraries that support)
+  //
+  // In this snippet, we demonstrate "Implicit authentication with ADC".
+  // ADC - Application Default Credentials
+  // When interacting with Google Cloud Client libraries, the library can auto-detect the
+  // credentials to use, if the "APPLICATION_DEFAULT_CREDENTIALS" is set.
```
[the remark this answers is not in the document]
piaxc: Maybe I'm not understanding how the GitHub review process works-- but I'm still seeing this sentence. Am I looking at an old version or something?

samples/snippets/src/main/java/AuthenticateImplicitWithAdc.java
```diff
@@ -0,0 +1,80 @@
...
+  // This APPLICATION_DEFAULT_CREDENTIALS is an environment variable/ configuration.
```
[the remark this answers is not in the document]
piaxc: This too-- are you still working on this? I'm confused...

samples/snippets/src/main/java/IdTokenFromImpersonatedCredentials_OAuth.java
```diff
@@ -0,0 +1,113 @@
+/*
+ * Copyright 2022 Google Inc.
```
piaxc, line 2: I'm not going to review this snippet right now. We need to decide how and where to incorporate it. Right now, it doesn't really fit the flow in either the authentication docs or the IAM docs. But it's good we have it. Can you put it somewhere where we can return to it, but it's not yet visible to users?

samples/snippets/src/main/java/AuthenticateExplicit.java
```diff
@@ -0,0 +1,72 @@
...
+    //  1. Replace the below variable.
+    //  2. Make sure you have the necessary permission to list storage buckets "storage.buckets.list"
+    String projectId = "your-google-cloud-project-id";
+
+    // Provide the scopes that you might need to request to access Google APIs,
+    // depending on the level of access you need.
+    // For more information, see: https://developers.google.com/identity/protocols/oauth2/scopes
+    // The best practice is to use the cloud-wide scope and use IAM to narrow the permissions.
+    // https://cloud.google.com/docs/authentication#authorization_for_services
+    String scope = "https://www.googleapis.com/auth/cloud-platform";
```
piaxc, line 38: Change to a generic scope (not an actual one, a dummy).

samples/snippets/src/main/java/IdTokenFromMetadataServer.java
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  // Use the Google Cloud metadata server in the Cloud Run (or AppEngine or Kubernetes etc.,)
+  // environment to create an identity token and add it to the HTTP request as part of an
+  // Authorization header.
+  public static void getIdTokenFromMetadataServer(String url)
+      throws IOException {
+    // Construct the GoogleCredentials object which obtains the default configuration from your
+    // working environment.
+    // Optionally, you can also set scopes.
```
piaxc, line 50: Scope doesn't make sense for ID tokens, remove.

samples/snippets/src/main/java/IdTokenFromMetadataServer.java
```diff
@@ -0,0 +1,79 @@
...
+        .setTargetAudience(url)
+        // Setting the ID token options.
+        .setOptions(Arrays.asList(Option.FORMAT_FULL, Option.LICENSES_TRUE))
+        .build();
+
+    // Get the ID token.
+    System.out.printf("Generated ID token: %s", idTokenCredentials.refreshAccessToken().getTokenValue());
+
+    // Make an authenticated HTTP request with the idTokenCredentials.
+    makeAuthenticatedRequest(idTokenCredentials, url);
```
piaxc, line 64: Would this be needed to achieve the objective of generating an ID token, or is it more of a test showing the ID token being used? If the latter, I'm thinking we should remove it to avoid confusion. LMKWYT. At the least, provide comments making it super clear that this is to demonstrate usage and they would use a different call.

samples/snippets/src/main/java/AuthenticateExplicit.java
```diff
@@ -0,0 +1,72 @@
...
+
+public class AuthenticateExplicit {
+
+  public static void main(String[] args) throws IOException, GeneralSecurityException {
+    // TODO(Developer):
+    //  1. Replace the below variable.
+    //  2. Make sure you have the necessary permission to list storage buckets "storage.buckets.list"
+    String projectId = "your-google-cloud-project-id";
+
+    // Provide the scopes that you might need to request to access Google APIs,
```
piaxc, line 33: Maybe continue with numbers, since you started them above?
// 3.

samples/snippets/src/main/java/AuthenticateImplicitWithAdc.java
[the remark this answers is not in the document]
piaxc: Looks good

samples/snippets/src/main/java/AuthenticateImplicitWithAdc.java
```diff
@@ -0,0 +1,52 @@
...
+
+import com.google.cloud.compute.v1.Instance;
+import com.google.cloud.compute.v1.InstancesClient;
+import java.io.IOException;
+
+public class AuthenticateImplicitWithAdc {
+
+  public static void main(String[] args) throws IOException {
+    // TODO(Developer):
+    //  1. Before running this sample, authenticate using ADC as mentioned in:
```
piaxc, line 25: // 1. Before running this sample, set up ADC as described in https://cloud.google.com/docs/authentication/external/set-up-adc
// 2. (same)
// 3. Make sure that the user account or service account that you are using has the required permissions. For this sample, you must have "compute.instances.list".

samples/snippets/src/main/java/AuthenticateImplicitWithAdc.java
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  // When interacting with Google Cloud Client libraries, the library can auto-detect the
+  // credentials to use.
+  // ADC detection is independent of the client library and language and works with all Cloud Client
+  // libraries.
+  public static void authenticateImplicitWithAdc(String project) throws IOException {
+
+    String zone = "us-central1-a";
+    // This snippet demonstrates how to initialize Cloud Compute Engine and list instances.
```
piaxc, line 40: "...list instances. Replace the client created below with the client required for your application."

b/199979085

samples/snippets/src/main/java/IdTokenFromServiceAccount.java
piaxc: And, Jeff said it would be ok to leave in snippets that use service account keys with the proper language, so reviewing the comments.

samples/snippets/src/main/java/IdTokenFromServiceAccount.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.Arrays;
+import java.util.List;
+import java.util.concurrent.ExecutionException;
+
+public class IdTokenFromServiceAccount {
+
+  public static void main(String[] args)
+      throws IOException, ExecutionException, InterruptedException, GeneralSecurityException {
+    // TODO(Developer): Replace the below variables before running the code.
+    //  Using Service account key is discouraged. Please consider alternate approaches first.
```
piaxc, line 32: Using service account keys introduces risk; they are long-lived, and can be used by anyone that obtains the key. Proper rotation and storage reduce this risk but do not eliminate it. For these reasons, you should consider an alternative approach that does not use a service account key. Several alternatives to service account keys are described here:
https://cloud.google.com/docs/authentication/external/set-up-adc

samples/snippets/src/main/java/IdTokenFromServiceAccount.java
```diff
@@ -0,0 +1,77 @@
...
+    // Path to the service account json credential file.
+    String jsonCredentialPath = "path-to-json-credential-file";
+
+    // Provide the scopes that you might need to request to access Google APIs,
```
piaxc, line 36: Scopes don't really make sense for ID tokens-- can this be done without them?

If not, use this verbiage:

When you are accessing Google Cloud services, it is a best practice to use a scope that includes all Google Cloud services (https://www.googleapis.com/auth/cloud-platform) and use IAM for authorizing access to resources.

samples/snippets/src/main/java/IdTokenFromServiceAccount.java
```diff
@@ -0,0 +1,77 @@
...
+    // depending on the level of access you need.
+    // For more information, see: https://developers.google.com/identity/protocols/oauth2/scopes
+    // The best practice is to use the cloud-wide scope and use IAM to narrow the permissions.
+    // https://cloud.google.com/docs/authentication#authorization_for_services
+    String scope = "https://www.googleapis.com/auth/cloud-platform";
+
+    // The service name for which the id token is requested. Service name refers to the
+    // logical identifier of an API service, such as "pubsub.googleapis.com".
+    String targetAudience = "iap.googleapis.com";
```
piaxc, line 45: This should be a url, like in the other sample: 
String url = "http://www.abc.com";

piaxc (changes requested):
Thanks Sita!

samples/snippets/src/main/java/AuthenticateExplicit.java
```diff
@@ -0,0 +1,72 @@
...
+    // depending on the level of access you need.
+    // For more information, see: https://developers.google.com/identity/protocols/oauth2/scopes
+    // The best practice is to use the cloud-wide scope and use IAM to narrow the permissions.
```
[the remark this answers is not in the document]
Sita04: Done. Modified. PTAL.

samples/snippets/src/main/java/AuthenticateImplicitWithAdc.java
Sita04: This makes it much clearer to the customers on what to replace. Thanks Pia.

samples/snippets/src/main/java/IdTokenFromServiceAccount.java
Sita04: Added the comment. Thanks.

samples/snippets/src/main/java/IdTokenFromServiceAccount.java
Sita04: Yes, we can omit the scopes. They don't add a lot of value to the sample as such. 
Modified. PTAL.

samples/snippets/src/main/java/AuthenticateImplicitWithAdc.java
Sita04: Thanks. Modified.

samples/snippets/src/main/java/AuthenticateExplicit.java
Sita04: The above ones are TODO comments and they are separate from the parameter descriptions. Hence, I didn't continue the numbers.

To make it clear, I've formatted with a newline. PTAL.

samples/snippets/src/main/java/IdTokenFromImpersonatedCredentials_IAM.java
```diff
@@ -0,0 +1,105 @@
+/*
```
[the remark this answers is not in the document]
Sita04: More context on why the sample will be deleted:

This sample uses the older IAM Apiary library. And this might not be the best way to obtain a IdToken. Rather, use the OAuth library which provides the appropriate methods for the same.

samples/snippets/src/main/java/IdTokenFromImpersonatedCredentials_OAuth.java
Sita04: We can leave it as such in the repo and not add a region tag. That way it'll not be picked up by the snippet bot and hence will not be visible in cgc. wdyt?

samples/snippets/src/main/java/IdTokenFromMetadataServer.java
Sita04: Saw Salmaan's comment on this. 
Modified the comment. thanks.

samples/snippets/src/main/java/IdTokenFromMetadataServer.java
Sita04: Thought about it and I think you are right in wanting to exclude them. We do not want to unnecessarily provide more information than that's required to achieve the objective. Hence, removed the testing part from the samples.

Thanks Pia for your thoughts on this.

samples/snippets/pom.xml
```diff
@@ -0,0 +1,101 @@
...
+      <groupId>com.auth0</groupId>
+      <artifactId>jwks-rsa</artifactId>
+      <version>0.18.0</version>
+    </dependency>
+
+<!--    GCloud dependency-->
+    <dependency>
+      <artifactId>google-cloud-compute</artifactId>
+      <groupId>com.google.cloud</groupId>
+      <version>1.8.1</version>
```
kolea2, line 77: can we remove this version number since we are using the BOM? Also, I would swap the order on `groupId` and `artifactId`

samples/snippets/src/main/java/VerifyNonGoogleIdToken.java
```diff
@@ -0,0 +1,100 @@
...
+      System.out.println("Could not verify Signature: " + e.getMessage());
+      isVerified = false;
+    }
+
+    if (isVerified) {
+      System.out.println("Id token verified.");
+      return;
+    }
+    System.out.println("Unable to verify ID token.");
+  }
```
chingor13, line 98: Our TokenVerifier class supports loading a [public key explicitly](https://github.com/googleapis/google-auth-library-java/blob/ef17ed8bf2c91e7bd66d26f039f029de6af3814a/oauth2_http/java/com/google/auth/oauth2/TokenVerifier.java#L238) or [via a certificates URL](https://github.com/googleapis/google-auth-library-java/blob/ef17ed8bf2c91e7bd66d26f039f029de6af3814a/oauth2_http/java/com/google/auth/oauth2/TokenVerifier.java#L215).

We just default to loading from Google's public certificates endpoint.

samples/snippets/src/main/java/AuthenticateImplicitWithAdc.java
```diff
@@ -0,0 +1,56 @@
...
+  // credentials to use.
+  public static void authenticateImplicitWithAdc(String project) throws IOException {
+
+    String zone = "us-central1-a";
+    // This snippet demonstrates how to list instances.
+    // *NOTE*: Replace the client created below with the client required for your application.
+    //
+    // Note that the credentials are not specified when constructing the client.
+    // Hence, the client library will look for credentials using ADC.
+    try (InstancesClient instancesClient = InstancesClient.create()) {
```
kurtisvg, line 46: Please. make sure to include the comment from the style guide when demonstrating how to init clients: https://googlecloudplatform.github.io/samples-style-guide/#clients

kurtisvg (changes requested):

samples/snippets/pom.xml
Sita04: Done. Thanks.

samples/snippets/src/main/java/AuthWithCredentialsFromMetadataServer.java
```diff
@@ -0,0 +1,74 @@
...
+      System.out.println(bucket.toString());
+    }
+    System.out.println("Authentication complete.");
+  }
+
+  // Initialize the Storage client by getting the credentials
+  // from a Metadata server.
+  private static Storage initService(String projectId) {
+    // Explicitly request the credentials from the ComputeEngine metadata server.
+    GoogleCredentials credentials = ComputeEngineCredentials.create();
```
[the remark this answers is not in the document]
Sita04: I agree.. modified to use ADC and also added a comment.

@kurtisvg we didn't want to demonstrate scopes as we thought it had a limited usecase for GCP and primarily only workspace products need it. @piaxc can you weigh in on the "scopes" part?

samples/snippets/src/main/java/AuthWithCredentialsFromMetadataServer.java
```diff
@@ -0,0 +1,74 @@
...
+    // This snippet demonstrates how to initialize Cloud Storage and list buckets.
+    // Note that the credentials are requested from the ComputeEngine metadata server.
+    Storage storage = initService(project);
+
+    System.out.println("Buckets:");
+    Page<Bucket> buckets = storage.list();
+    for (Bucket bucket : buckets.iterateAll()) {
+      System.out.println(bucket.toString());
+    }
+    System.out.println("Authentication complete.");
```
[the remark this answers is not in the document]
Sita04: Yes, seems like a good idea. PTAL.

samples/snippets/src/main/java/IdTokenFromImpersonatedCredentials.java
Sita04: The delegates parameter is mandatory in the "Impersonated Credentials" API, unlike the other one we used from IAM library. 
Hence, I am unable to remove it. Added clarifying comments stating that it's optional. wdyt?

samples/snippets/src/main/java/IdTokenFromImpersonatedCredentials.java
```diff
@@ -0,0 +1,96 @@
...
+        "name@project.service.gserviceaccount.com";
+
+    getIdTokenUsingOAuth2(impersonatedServiceAccount, scope, targetAudience);
+  }
+
+  // Use a service account (SA1) to impersonate as another service account (SA2) and obtain id token
+  // for the impersonated account.
+  // To obtain token for SA2, SA1 should have the "roles/iam.serviceAccountTokenCreator" permission
+  // on SA2.
+  public static void getIdTokenUsingOAuth2(String impersonatedServiceAccount, String scope,
```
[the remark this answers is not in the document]
Sita04: Nope, this param also looks mandatory to me. 
Unless, we switch this approach for the IAM one.

samples/snippets/src/main/java/AuthenticateImplicitWithAdc.java
Sita04: Sure, thanks.

samples/snippets/src/main/java/AuthWithCredentialsFromMetadataServer.java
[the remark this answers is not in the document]
piaxc: To chingor13@'s comment: the idea here was to specifically get credentials from the Metadata server. This is in keeping with previous samples that have been provided, although this sample won't be included in the docs. Sita04@, how will this differ from the implicit flow now? Maybe we could just remove it?

kurtisvg@: for Cloud APIs, the best practice as communicated by the IAM team is to use the cloud-wide scope "https://www.googleapis.com/auth/cloud-platform" and use IAM only to restrict access. That is the guidance I'm providing in the Authentication documentation. Maybe we could provide more color?

For Cloud APIs, you can let the library use the default scope of <code>https://www.googleapis.com/auth/cloud-platform</code> because IAM provides fine-grained access control.

Or something like that.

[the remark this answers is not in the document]
Sita04: I think it's best to remove the sample as there is very minute difference between this one and the ADC explicit.
I'll add an extra comment in that sample that @chingor13 mentioned "GoogleCredentials.getApplicationDefault() will give you ComputeEngineCredentials if you are on a GCE (or other metadata server supported environments)." ?

samples/snippets/src/main/java/AuthWithCredentialsFromMetadataServer.java
```diff
@@ -0,0 +1,74 @@
...
+    authWithCredentialsFromMetadataServer(projectId);
+  }
+
+  // In this snippet, we demonstrate "Authentication with account credentials
+  // obtained from a metadata server".
+  public static void authWithCredentialsFromMetadataServer(String project) {
+
+    // This snippet demonstrates how to initialize Cloud Storage and list buckets.
+    // Note that the credentials are requested from the ComputeEngine metadata server.
+    Storage storage = initService(project);
```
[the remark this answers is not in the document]
Sita04: This sample has been deleted. 
But applied your suggestion to another sample "AuthenticateExplicit".
Thanks.

samples/snippets/src/main/java/VerifyNonGoogleIdToken.java
Sita04: Thanks Jeff. Would you recommend we also demonstrate using that?

samples/snippets/src/main/java/IdTokenFromImpersonatedCredentials.java
piaxc: Could we just say 
// delegates: Unused for this call
or something like that? I'm concerned that the subject of chaining is very complex, and that folks will spend a lot of time trying to parse and understand what identities need to have what permission on what-- only to get to the bottom and find out it's unused. Why put them through that?

piaxc (approved):
LGTM, thanks Sita!

samples/snippets/src/main/java/VerifyGoogleIdToken.java
```diff
@@ -0,0 +1,72 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// [START auth_cloud_verify_google_idtoken]
+
+import com.auth0.jwk.JwkException;
+import com.google.api.client.googleapis.auth.oauth2.GoogleIdToken;
+import com.google.api.client.googleapis.auth.oauth2.GoogleIdToken.Payload;
+import com.google.api.client.googleapis.auth.oauth2.GoogleIdTokenVerifier;
```
chingor13, line 22: This verifier is from google-api-java-client and was not being maintained. It also had a CVE until recently where we [weren't actually validating the signature](https://github.com/googleapis/google-oauth-java-client/issues/786).

We should instead use the token verifier from this library (like https://github.com/GoogleCloudPlatform/java-docs-samples/blob/96c4f10bf01683d0e0a5b2a44ba4bee1ac5fc507/iap/src/main/java/com/example/iap/VerifyIapRequestHeader.java#L60-L73)

samples/snippets/src/main/java/VerifyNonGoogleIdToken.java
chingor13: I would recommend that we use our first-party implementation and not require extra third-party dependencies.

samples/snippets/src/main/java/AuthenticateExplicit.java
```diff
@@ -0,0 +1,72 @@
...
+    // https://www.googleapis.com/auth/cloud-platform, because IAM is used to provide fine-grained
+    // permissions for Cloud.
+    // If you need to provide a scope, specify it as follows:
+    // GoogleCredentials credentials = GoogleCredentials.getApplicationDefault()
+    //     .createScoped(scope);
+    // For more information on scopes to use,
+    // see: https://developers.google.com/identity/protocols/oauth2/scopes
+
+    // Construct the Storage client.
+    Storage storage =
```
kurtisvg, line 57: This should use a try/catch and include the warning about making sure to close: disingenuous: https://googlecloudplatform.github.io/samples-style-guide/#clients

kurtisvg (approved):

samples/snippets/src/main/java/AuthenticateExplicit.java
```diff
@@ -0,0 +1,77 @@
...
+public class AuthenticateExplicit {
+
+  public static void main(String[] args) throws IOException, GeneralSecurityException {
+    // TODO(Developer):
+    //  1. Replace the project variable below.
+    //  2. Make sure you have the necessary permission to list storage buckets "storage.buckets.list"
+
+    String projectId = "your-google-cloud-project-id";
+
+    // If you are authenticating to a Cloud API, you do not need to specify a scope;
```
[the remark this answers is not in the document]
Sita04: I've taken a middle ground here and have demonstrated both (with and wo scopes) in the sample with comments on where to use them. PTAL.

samples/snippets/src/main/java/AuthenticateExplicit.java
Sita04: I am not sure of this as many of the storage samples look like they don't follow the auto close try-catch..
https://github.com/googleapis/java-storage/blob/main/samples/snippets/src/main/java/com/example/storage/bucket/ListBuckets.java

samples/snippets/src/main/java/VerifyGoogleIdToken.java
Sita04: Done. I've made the change you suggested and also removed the third party dependency.
PTAL.

chingor13 (approved):
2 non-blocking comments

Sita04:
> 2 non-blocking comments

Thanks Jeff! Included your comments in the sample for further clarification.

PR #948, opened 2022-07-21
fix: Override AwsSecurityCredentials

edjeffreys:
See https://issuetracker.google.com/issues/238911014

When using a Web Identity token file (AWS_WEB_IDENTITY_TOKEN_FILE) to authenticate with AWS, environment variables AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, and AWS_SESSION_TOKEN are not set.

While the existing implementation then makes a request to Amazon's EC2 metadata service endpoint; when the Java instance is running in a pod in EKS, the metadata endpoint does not return metadata for the pod, but for the underlying EKS node.

To work around this, we can optionally set these variables in the credential builder instead of looking up the env variables or polling the metadata endpoint.


TimurSadykov:
@lsirac does this look good to you?

TimurSadykov:
Sorry for a delay, i'll look into it by the end of the week

edjeffreys:
No problem, I've got a workaround in place which is doing the trick for now.

lsirac:
Hey @edjeffreys - can you help me understand how you are retrieving the AWS security creds?

edjeffreys:
Essentially, we're running our workload on Kubernetes using [IAM Roles for Service Accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html), which means that we cannot get the credentials from IMDS, and the credentials aren't stored in ENV vars.

`WebIdentityTokenCredentialsProvider` exists as part of Amazon's `DefaultAWSCredentialsProviderChain` to retrieve temporary credentials.

So here, I'm attempting to fetch the credentials with an arbitrary provider, in my case `WebIdentityCredentialsProvider`, and provide them to google auth manually when building AwsCredentials.

pwalczak:
@edjeffreys, just for curiosity - if the token is rotated, are you going to handle this situation in client code? Means you are going to create new credential builder with the new values being set?

edjeffreys:
This has broken my workaround https://github.com/googleapis/google-auth-library-java/pull/1079/files - where I was providing a custom IMDS implementation to provide the correct credentials, I can no longer.

Is there a proper fix in the pipeline to support AWS IRSA credentials?

edjeffreys:
Any update on this @TimurSadykov? With the new validation checking that IMDS address is not set to anything custom, I can no longer fetch credentials using our workaround custom IMDS sidecar, unless I stay on an old version of google auth.

PR #949, opened 2022-07-22
churn: add auth team as codeowners

TimurSadykov:
This is to automate adding auth team folks to PRs to review auth bits


TimurSadykov:
not needed

Issue #929, opened 2022-06-13
Backport the fix to the public key caching for the IdTokenVerifier


TimurSadykov:
Current public key fetch on the IdTokenVerified can cache empty entry for 1 hour if request fails for any reason. We want to backport the fix that keep old entries active during failures: https://github.com/googleapis/google-oauth-java-client/pull/892

TimurSadykov:
Dropped the priority since this library throws errors and users can retry effectively. Fix is in progress, ETA Jul 12

TimurSadykov:
A fix got checked in

PR #967, opened 2022-08-09
fix: updates IdTokenVerifier so that it does not cache a failed public key response 

TimurSadykov:
avoiding caching an empty entry in case of public key request failure
avoiding caching a corrupted response of a public key request

Fixes the #929  ☕️

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/TokenVerifier.java
+++ b/oauth2_http/java/com/google/auth/oauth2/TokenVerifier.java
@@ -314,17 +314,13 @@ public class TokenVerifier {
     public Map<String, PublicKey> load(String certificateUrl) throws Exception {
       HttpTransport httpTransport = httpTransportFactory.create();
       JsonWebKeySet jwks;
-      try {
-        HttpRequest request =
-            httpTransport
-                .createRequestFactory()
-                .buildGetRequest(new GenericUrl(certificateUrl))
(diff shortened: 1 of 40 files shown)


oauth2_http/java/com/google/auth/oauth2/TokenVerifier.java
```diff
@@ -314,17 +314,13 @@ public static class JsonWebKey {
...
-            httpTransport
-                .createRequestFactory()
-                .buildGetRequest(new GenericUrl(certificateUrl))
-                .setParser(OAuth2Utils.JSON_FACTORY.createJsonObjectParser());
-        HttpResponse response = request.execute();
-        jwks = response.parseAs(JsonWebKeySet.class);
-      } catch (IOException io) {
-        return ImmutableMap.of();
-      }
+      HttpRequest request =
```
sai-sunder-s, line 317: Do try catch like before but throw a good exception instead of swallowing?

clundin25 (approved):

oauth2_http/java/com/google/auth/oauth2/TokenVerifier.java
TimurSadykov: we don't change it and throw the same thing - so there is no need. 

In the old library, in comparison, we have a logger and at least can trace before rethrowing

PR #971, opened 2022-08-09
fix: Add Test to validate 0x20 in token

sai-sunder-s:
Adds a test for a reported condition where 0x20 character causes an exception


oauth2_http/java/com/google/auth/oauth2/IdToken.java
```diff
@@ -77,6 +78,7 @@ public static IdToken create(String tokenValue) throws IOException {
    * @return returns com.google.auth.oauth2.IdToken
    */
   public static IdToken create(String tokenValue, JsonFactory jsonFactory) throws IOException {
+    tokenValue = tokenValue.trim();
```
TimurSadykov, line 81: add a test to catch regressions?

sai-sunder-s: just adding test to validate that 0x20 is handled by java lib if it is part of the token in an expected way

TimurSadykov (approved):
LGTM

PR #975, opened 2022-08-11
fix: disable logging from http requests and responses.

clundin25:
The contents of the Auth library requests contains sensitive data
that should not be logged.

Fixes https://github.com/googleapis/google-oauth-java-client/issues/878 ☕️


TimurSadykov (changes requested):
let's cancel this one

PR #983, opened 2022-08-19
fix: add retries to public key fetch

### LINKED ISSUE #930: Improve public key fetching for IdTokenVerifier
Current implementation of the public key fetching does refresh public keys every hour. If public certs refresh fails during refresh - ID token validation is blocked until refresh succeeds. Normally we expect that to be rare and transient, but we want to try to avoid it altogether. 

Alternative solution is to check certificate field for expiration date and refresh when the date is close. Potentially we want to consider a combination of two.

TimurSadykov:
An improvement to public key fetch to cover potential transient errors with retries.

Fixes #930  ☕️

### DIFF
--- a/.github/.OwlBot.lock.yaml
+++ b/.github/.OwlBot.lock.yaml
@@ -13,5 +13,5 @@
 # limitations under the License.
 docker:
   image: gcr.io/cloud-devrel-public-resources/owlbot-java:latest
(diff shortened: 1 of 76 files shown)


sai-sunder-s (approved):

aitemour79 (commented):
OK help my I like seber  robot

aitemour79:
Ok

aitemour79:
Ok

aitemour79:
Come facio con voi

PR #984, opened 2022-08-19
fix: Populate HttpHeaders fields with List<String> instead of String.

tjgq:
IdentityPoolCredentials#getSubjectTokenFromMetadataServer calls
HttpHeaders.putAll to set request headers. The latter sets its fields through
reflective access: well-known headers such as Content-Type or Authentication
have dedicated fields of type List<String>, while remaining headers go into a
Map<String, Object> grab bag. However, we attempt to set every header to a
String, which causes a crash for well-known headers.

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
@@ -51,7 +51,9 @@ import java.nio.file.LinkOption;
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import javax.annotation.Nullable;
@@ -230,7 +232,10 @@ public class IdentityPoolCredentials extends ExternalAccountCredentials {
 
     if (identityPoolCredentialSource.hasHeaders()) {
       HttpHeaders headers = new HttpHeaders();
-      headers.putAll(identityPoolCredentialSource.headers);
+      for (Map.Entry<String, String> entry : identityPoolCredentialSource.headers.entrySet()) {
+        // HttpHeaders expects List<String> instead of String.
+        headers.put(entry.getKey(), Collections.singletonList(entry.getValue()));
+      }
       request.setHeaders(headers);
     }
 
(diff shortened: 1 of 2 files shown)


aitemour79:
OK good

TimurSadykov:
@lsirac Please take a look

lsirac:
Hey @tjgq, I don't understand why this change is necessary. Can you elaborate?

tjgq:
> Hey @tjgq, I don't understand why this change is necessary. Can you elaborate?

Suppose you instantiate an `IdentityPoolCredentials` from a JSON file that contains an `Authorization` header, e.g. `{"header": {"Authorization": "Bearer xxx"}}`. This will populate the `headers` member variable, which is currently a `Map<String, String>`.

Later, when `IdentityPoolCredentials#getSubjectTokenFromMetadataServer` is called, the headers will be copied into an `HttpHeaders` object via `HttpHeaders#putAll(...)` where the argument is the `headers` member variable populated above.

If you look at the implementation of `HttpHeaders#putAll` (which is actually in the superclass, `GenericData#putAll`), you'll see that it uses reflection to find a method with an appropriate name for each header. For example, for an `Authorization` header, it will attempt to find a method called `setAuthorization` and call it with the header value, which is a `String`. But `HttpHeaders#setAuthorization` expects a `List<String>` argument, so the reflective call fails.

You can see what the stack trace for this failure looks like in https://github.com/bazelbuild/bazel/issues/15639#issue-1264351778. You can also reproduce it yourself by running `IdentityPoolCredentialsTest` as changed by this PR, without making the corresponding changes to `IdentityPoolCredentials`.

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
```diff
@@ -84,7 +86,7 @@ enum CredentialFormatType {
     private String credentialLocation;
 
     @Nullable private String subjectTokenFieldName;
-    @Nullable private Map<String, String> headers;
+    @Nullable private Map<String, List<String>> headers;
```
lsirac, line 89: Let's keep this as Map<String, String>

tjgq: Done.

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
```diff
@@ -230,7 +232,10 @@ private String getSubjectTokenFromMetadataServer() throws IOException {
 
     if (identityPoolCredentialSource.hasHeaders()) {
       HttpHeaders headers = new HttpHeaders();
-      headers.putAll(identityPoolCredentialSource.headers);
+      for (Map.Entry<String, String> entry : identityPoolCredentialSource.headers.entrySet()) {
+        // HttpHeaders expects List<String> instead of String.
```
lsirac, line 236: I can't reproduce the issue you've outlined in this PR. Also put is (String, Object)?

PR #990, opened 2022-08-29
docs(samples): modified comments in the samples and minor refactor

Sita04:
Changes:

1. Modified Auth implicit sample to use GCS instead of GCE.
2. Modified comments
3. Minor variable value refactoring

### DIFF
--- a/samples/snippets/src/main/java/AuthenticateImplicitWithAdc.java
+++ b/samples/snippets/src/main/java/AuthenticateImplicitWithAdc.java
@@ -40,7 +40,6 @@ public class AuthenticateImplicitWithAdc {
   // credentials to use.
   public static void authenticateImplicitWithAdc(String project) throws IOException {
 
-    // This snippet demonstrates how to list storage buckets.
     // *NOTE*: Replace the client created below with the client required for your application.
     // Note that the credentials are not specified when constructing the client.
     // Hence, the client library will look for credentials using ADC.
--- a/samples/snippets/src/main/java/AuthenticateImplicitWithAdc.java
+++ b/samples/snippets/src/main/java/AuthenticateImplicitWithAdc.java
@@ -47,11 +47,7 @@ public class AuthenticateImplicitWithAdc {
     //
     // Initialize client that will be used to send requests. This client only needs to be created
     // once, and can be reused for multiple requests.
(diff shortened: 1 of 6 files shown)


samples/snippets/src/main/java/AuthenticateImplicitWithAdc.java
```diff
@@ -37,24 +40,21 @@ public static void main(String[] args) throws IOException {
   // credentials to use.
   public static void authenticateImplicitWithAdc(String project) throws IOException {
 
-    String zone = "us-central1-a";
-    // This snippet demonstrates how to list instances.
+    // This snippet demonstrates how to list storage buckets.
```
piaxc, line 43: Please remove-- while strictly true, it's not really what the snippet is about (adc).

samples/snippets/src/main/java/VerifyGoogleIdToken.java
```diff
@@ -26,9 +26,8 @@ public static void main(String[] args) {
     // The Google ID token to verify.
     String idToken = "id-token";
 
-    // The service name for which the id token is requested. Service name refers to the
-    // logical identifier of an API service, such as "pubsub.googleapis.com".
-    String targetAudience = "pubsub.googleapis.com";
+    // The service name for which the id token is requested.
```
piaxc, line 29: This is interesting. If you are validating an ID token, you should be checking what is in the token against what you expect, right? So wouldn't this be something like 'The expected target, for example, "https://example.com"'?

piaxc (changes requested):

samples/snippets/src/main/java/AuthenticateImplicitWithAdc.java
Sita04: done.

samples/snippets/src/main/java/VerifyGoogleIdToken.java
Sita04: We are not doing a manual check here. Instead we let the API verify by providing the details that we used during creating the ID token.
Hence, I would prefer to leave it as such.

piaxc (approved):
Thanks Sita!

lesv (approved):

PR #1002, opened 2022-09-21
Proposal on how to fix issue #692

### LINKED ISSUE #692: Possible race condition in credential refresh
We are seeing a possible race condition around credential refresh. We use the following code in an integration test suite to fetch OAuth2 tokens:

```java
    GoogleCredentials credentials = ...;
    try {
      if (forceRefresh) {
        credentials.refresh();
      } else {
        credentials.getRequestMetadata();
      }

      AccessToken token = credentials.getAccessToken();
      return token.getTokenValue();
    } catch (IOException e) {
      throw new RuntimeException(e);
    }
```

We expect the `credentials` to contain a valid access token at the end of the if-else block. However, our tests [intermittently fail](https://github.com/firebase/firebase-admin-java/runs/3025628758?check_suite_focus=true) by throwing a `NullPointerException` at `token.getTokenValue()`. Any thoughts on what might be going on here? What should we do to ensure that `getAccessToken()` always returns a non-null value?

Note: In the failing tests, `forceRefresh` has been set to `false`. So it should execute the `credentials.getRequestMetadata()` path.

johanblumenberg:
## Questionnaire

Fixes #692

## Problem

There is a race condition here https://github.com/googleapis/google-auth-library-java/blob/429b481bbedb9cd96ce7a27cc9391eb8c661030a/oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java#L252

When `task` is completed, two things happen:
- This listener executes, which sets the current `value`: https://github.com/googleapis/google-auth-library-java/blob/429b481bbedb9cd96ce7a27cc9391eb8c661030a/oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java#L261
- This code is unblocked, which signals to the caller that the refresh was completed: https://github.com/googleapis/google-auth-library-java/blob/429b481bbedb9cd96ce7a27cc9391eb8c661030a/oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java#L181

When more than one thread is involved, there is a race condition in which of these that get to execute first. Most of the time the listener executes first, and everything works. But now and then the listener is delayed, and that means that whoever is waiting for `refresh()` to complete will read the old token.

## Logs

Adding a sleep in the listener shows that the race condition exists: https://github.com/veritru/google-auth-library-java/commit/551513b233339b952ec40366e12d448e8c7f7661

This produces the following logs:

```
2022-09-21 15:07:35.288 INFO --- [pool-6-thread-1] OAuth2Credentials : Starting refresh task
2022-09-21 15:07:35.289 INFO --- [pool-6-thread-2] OAuth2Credentials : refresh()
2022-09-21 15:07:35.308 INFO --- [pool-6-thread-1] OAuth2Credentials : Refresh task done
2022-09-21 15:07:35.308 INFO --- [pool-6-thread-2] OAuth2Credentials : refresh() done
2022-09-21 15:07:40.309 INFO --- [pool-6-thread-1] OAuth2Credentials : Update token info
```

The "refresh() done" log is the point where the caller would get and use the access token. It does not wait for the listener to actually update the current access token.

## Solution Proposal

The proposed solution is to move the update of `value` to be done before the task is completed.

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
@@ -254,7 +254,13 @@ public class OAuth2Credentials extends Credentials {
               new Callable<OAuthValue>() {
                 @Override
                 public OAuthValue call() throws Exception {
-                  return OAuthValue.create(refreshAccessToken(), getAdditionalHeaders());
+                  OAuthValue result = OAuthValue.create(refreshAccessToken(), getAdditionalHeaders());
+
+                  synchronized (lock) {
+                    value = result;
+                  }
+
+                  return result;
                 }
               });
 
@@ -281,12 +287,9 @@ public class OAuth2Credentials extends Credentials {
   private void finishRefreshAsync(ListenableFuture<OAuthValue> finishedTask) {
     synchronized (lock) {
       try {
-        this.value = finishedTask.get();
         for (CredentialsChangedListener listener : changeListeners) {
           listener.onChanged(this);
         }
-      } catch (InterruptedException e) {
-        Thread.currentThread().interrupt();
       } catch (Exception e) {
         // noop
       } finally {


oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
```diff
@@ -254,7 +254,13 @@ private AsyncRefreshResult getOrCreateRefreshTask() {
               new Callable<OAuthValue>() {
                 @Override
                 public OAuthValue call() throws Exception {
-                  return OAuthValue.create(refreshAccessToken(), getAdditionalHeaders());
+                  OAuthValue result = OAuthValue.create(refreshAccessToken(), getAdditionalHeaders());
+
```
kurtisvg, line 258: Would it be better to move `finishRefreshAsync` up onto this line?

johanblumenberg: Yes, that might be better.

I don't know if there is any listener that relies on the task being completed, so I didn't want to change it.

clundin25:
LGTM, do you mind signing the CLA and updating to a conventional commit?

oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
```diff
@@ -254,7 +254,13 @@ private AsyncRefreshResult getOrCreateRefreshTask() {
...
+                  synchronized (lock) {
```
clundin25, line 259: Is this lock necessary? Only one thread should be doing work on this future right?

clundin25:
There is a test failure, once that is resolved I think this can be merged.

https://github.com/googleapis/google-auth-library-java/blob/257071aeb39c4441bd152813d727f83d433f346f/oauth2_http/javatests/com/google/auth/oauth2/OAuth2CredentialsTest.java#L625 For someone reason it seems to me that the task is not getting set back to NULL.

clundin25:
Thanks @johanblumenberg for the fix. We are eager to get this merged so I have forked your PR here https://github.com/googleapis/google-auth-library-java/pull/1031 in order to resolve the test failures.

I will make sure you are credited if that is the PR that lands.

johanblumenberg:
> Thanks @johanblumenberg for the fix. We are eager to get this merged so I have forked your PR here #1031 in order to resolve the test failures.

@clundin25 Perfect, thanks a lot

oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
johanblumenberg: The lock was taken in the `finishRefreshAsync()` method when modifying the `value` variable. I wasn't sure why, if it was needed to protect the `value` variable or something else. But you are probably right, the lock is not necessary here.

TimurSadykov:
Thanks for suggestion, closing as we are about to fix with another PR: https://github.com/googleapis/google-auth-library-java/pull/1031

PR #1031, opened 2022-09-29
fix: Resolve race condition reported in #692

clundin25:
Fork of https://github.com/googleapis/google-auth-library-java/pull/1002 with test fixes.

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
@@ -254,9 +254,12 @@ public class OAuth2Credentials extends Credentials {
               new Callable<OAuthValue>() {
                 @Override
                 public OAuthValue call() throws Exception {
-                  OAuthValue result =
-                      OAuthValue.create(refreshAccessToken(), getAdditionalHeaders());
-                  value = result;
+                  OAuthValue result = OAuthValue.create(refreshAccessToken(), getAdditionalHeaders());
+
+                  synchronized (lock) {
+                    value = result;
+                  }
+
                   return result;
                 }
               });
@@ -284,18 +287,9 @@ public class OAuth2Credentials extends Credentials {
   private void finishRefreshAsync(ListenableFuture<OAuthValue> finishedTask) {
     synchronized (lock) {
       try {
-        // The future MUST be polled to make sure the credential is still valid before
-        // invoking the listeners. If the credential has been revoked we expect an exception to
-        // occur here and NOOP. Without this task MAY revive revoked credentials.
-        //
-        // We do not care to get the Future's value. The value should be updated in the Future
-        // itself.
-        finishedTask.get();
         for (CredentialsChangedListener listener : changeListeners) {
           listener.onChanged(this);
         }
-      } catch (InterruptedException e) {
-        Thread.currentThread().interrupt();
       } catch (Exception e) {
         // noop
       } finally {
--- a/oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
@@ -300,8 +300,6 @@ public class OAuth2Credentials extends Credentials {
    * extremely unlikely.
    */
   private static <T> T unwrapDirectFuture(ListenableFuture<T> future) throws IOException {
-    future = Futures.transform(future, Functions.identity(), MoreExecutors.directExecutor());
-
     try {
       return future.get();
     } catch (InterruptedException e) {
--- a/oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
@@ -282,7 +282,7 @@ public class OAuth2Credentials extends Credentials {
       } catch (Exception e) {
         // noop
       } finally {
-        if (this.refreshTask != null && this.refreshTask.task == finishedTask) {
+        if (this.refreshTask != null && this.refreshTask.getTask() == finishedTask) {
           this.refreshTask = null;
         }
       }
@@ -616,6 +616,10 @@ public class OAuth2Credentials extends Credentials {
             }
           },
           MoreExecutors.directExecutor());
+    }
+
+    public ListenableFutureTask<OAuthValue> getTask() {
+      return this.task;
     }
 
     public void run() {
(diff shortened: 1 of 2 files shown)


johanblumenberg (approved):
looks good

I tried it in our product as well and it seems to work

petedmarsh:
@johanblumenberg @igorbernstein2 @TimurSadykov 

I've also been affected by this race condition and I would like to see a fix (this one looks along the right lines to me at least), but I think there may be one thing missing which is making the `refreshTask` `volatile` too as in:

https://github.com/googleapis/google-auth-library-java/pull/1046

oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
```diff
@@ -307,7 +299,9 @@ private void finishRefreshAsync(ListenableFuture<OAuthValue> finishedTask) {
    * thread of whatever executor the async call used. This doesn't affect correctness and is
    * extremely unlikely.
    */
-  private static <T> T unwrapDirectFuture(Future<T> future) throws IOException {
+  private static <T> T unwrapDirectFuture(ListenableFuture<T> future) throws IOException {
+    future = Futures.transform(future, Functions.identity(), MoreExecutors.directExecutor());
```
igorbernstein, line 303: This shouldn’t be necessary. You are already enforcing ordering in the refreshtask

oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
```diff
@@ -290,7 +282,7 @@ private void finishRefreshAsync(ListenableFuture<OAuthValue> finishedTask) {
       } catch (Exception e) {
         // noop
       } finally {
-        if (this.refreshTask == finishedTask) {
+        if (this.refreshTask != null && this.refreshTask.task == finishedTask) {
```
igorbernstein, line 285: Why do you need to break encapsulation here?

oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
```diff
@@ -582,6 +576,53 @@ void executeIfNew(Executor executor) {
...
+  class RefreshTask extends AbstractFuture<OAuthValue> implements Runnable {
+    private final ListenableFutureTask<OAuthValue> task;
+    private final RefreshTaskListener listener;
+
+    RefreshTask(ListenableFutureTask<OAuthValue> task, RefreshTaskListener listener) {
+      this.task = task;
+      this.listener = listener;
+
+      // Update Credential state first
+      task.addListener(listener, MoreExecutors.directExecutor());
```
igorbernstein, line 602: I think you should either in-line both callbacks or move all the callback logic into the listener class.

oauth2_http/javatests/com/google/auth/oauth2/OAuth2CredentialsTest.java
```diff
@@ -873,6 +876,83 @@ public void serialize() throws IOException, ClassNotFoundException {
...
+                new RefreshTaskListener(task) {
+                  @Override
+                  public void run() {
+                    try {
+                      // Sleep before setting accessToken to new accessToken. Refresh should not
+                      // complete before this, and the accessToken is `null` until it is.
+                      Thread.sleep(500);
+                      super.run();
+                    } catch (Exception e) {
+                      fail("Unexpected error.");
```
igorbernstein, line 912: I would pass the unexpected exception to the fail so that we don’t leave the person debugging the failure in suspense

oauth2_http/javatests/com/google/auth/oauth2/OAuth2CredentialsTest.java
```diff
@@ -873,6 +876,83 @@ public void serialize() throws IOException, ClassNotFoundException {
     assertSame(deserializedCredentials.clock, Clock.SYSTEM);
   }
 
+  @Test
+  public void UpdateTokenValueBeforeWake() throws IOException, InterruptedException {
```
igorbernstein, line 880: Nit: methods should always use lowerCamelCase

igorbernstein: Also this test is very specific.

I wonder if it would make sense to add a smoke test that creates 500 creds and have 1000 threads race to refresh and get() the creds

oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
clundin25: The refresh task isn't tracked but the future performing the refresh is, which is wrapped by the refresh task.

Since there is only one refresh task at a time could this check be removed?

oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
clundin25: Not familiar with what this line does.

Sounds like I can go ahead and delete it?

oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
clundin25: In order to inject a sleep to exercise the race condition I wrote it like this. Agreed the logic could be better but I am not very familiar with Java

clundin25: That said I would happily accept a cleaner approach!

oauth2_http/javatests/com/google/auth/oauth2/OAuth2CredentialsTest.java
clundin25: This test makes sure there is no regression of the race condition.

I think smoke tests could make sense but I do not like how flakey / hard to debug they can be. 

We could add one later, that doesn't run with the unit test suite

TimurSadykov (approved):
LGTM, forgot to post few nit comments

oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
TimurSadykov: added the getter )

TimurSadykov:
@nnl-1 Please share what kind of client are you using where you see the issue and how you check the dependency tree?

nnl-1:
> @nnl-1 Please share what kind of client are you using where you see the issue and how you check the dependency tree?

I'm using this dep. https://mvnrepository.com/artifact/com.google.cloud.sql/postgres-socket-factory. The dep. tree command is 'mvn dependency:tree'.

nnl-1:
ok, now I see it — https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/commit/c6df99f3e894d95ffe66b0b129085ae6777f1667. Thanks

PR #1033, opened 2022-09-29
fix: race condition of credentials

igorbernstein2:


clundin25:
Closed in favor of https://github.com/googleapis/google-auth-library-java/pull/1031

PR #1046, opened 2022-10-07
Make OAuth2Credentials#refreshTask volatile

petedmarsh:
The synchronization on the lock is not enough to ensure that all threads have latest value of "refreshTask" which means it is possible for two refresh tasks to run at once:

    // OAuth2Credentials#L247
    synchronized (lock) {
      if (refreshTask != null) {
        return new AsyncRefreshResult(refreshTask, false);
      }

This is similar to why this.value has to be volatile.


petedmarsh:
I was wrong the synchronized block enough ok as `refreshTask` is only ever read or set in there.

https://jenkov.com/tutorials/java-concurrency/java-happens-before-guarantee.html#the-java-synchronized-visibility-guarantee

PR #1054, opened 2022-10-14
feat: Introduce Environment Variable for Quota Project Id

sai-sunder-s:
Introduce a new environment variable "GOOGLE_CLOUD_QUOTA_PROJECT" that can be used to specify a quota project id. This quota project will be used to override the quota project from the credential detected during ADC generation if another quota project was not provided explicitly.
This change also introduces a helper method to set the quota project from environment variable on the credential explicitly.


TimurSadykov:
not needed anymore

PR #1082, opened 2022-11-05
feat: Introduce Environment Variable for Quota Project Id

sai-sunder-s:
Introduce a new environment variable "GOOGLE_CLOUD_QUOTA_PROJECT" that can be used to specify a quota project id. This quota project will be used to override the quota project from the credential detected during ADC generation.


oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java
```diff
@@ -223,6 +225,12 @@ private final GoogleCredentials getDefaultCredentialsUnsynchronized(
       credentials = tryGetComputeCredentials(transportFactory);
     }
 
+    String quotaProjectFromEnv = this.getEnv(QUOTA_PROJECT_ENV_VAR);
+
+    if (quotaProjectFromEnv != null && !quotaProjectFromEnv.isEmpty()) {
+      credentials = credentials.toBuilder().setQuotaProjectId(quotaProjectFromEnv).build();
```
TimurSadykov, line 231: please define a createWithQuotaProject in GoogleCredentials to encapsulate this

oauth2_http/javatests/com/google/auth/oauth2/DefaultCredentialsProviderTest.java
```diff
@@ -326,6 +328,32 @@ public void getDefaultCredentials_envUser_providesToken() throws IOException {
...
+        UserCredentialsTest.writeUserStream(
+            USER_CLIENT_ID, USER_CLIENT_SECRET, REFRESH_TOKEN, QUOTA_PROJECT);
+    TestDefaultCredentialsProvider testProvider = new TestDefaultCredentialsProvider();
+    String userPath = tempFilePath("user.json");
+    testProvider.addFile(userPath, userStream);
+    testProvider.setEnv(DefaultCredentialsProvider.CREDENTIAL_ENV_VAR, userPath);
+    MockTokenServerTransportFactory transportFactory = new MockTokenServerTransportFactory();
+    transportFactory.transport.addClient(USER_CLIENT_ID, USER_CLIENT_SECRET);
+    transportFactory.transport.addRefreshToken(REFRESH_TOKEN, ACCESS_TOKEN);
+
```
TimurSadykov, line 343: get credential before env is set and check quota is not present or different

oauth2_http/javatests/com/google/auth/oauth2/DefaultCredentialsProviderTest.java
```diff
@@ -326,6 +328,32 @@ public void getDefaultCredentials_envUser_providesToken() throws IOException {
...
+    // Validate that quota from env overrides the value from json
+    testProvider.setEnv(
+        DefaultCredentialsProvider.QUOTA_PROJECT_ENV_VAR, QUOTA_PROJECT_FROM_ENVIRONMENT);
+    GoogleCredentials credentials = testProvider.getDefaultCredentials(transportFactory);
+    assertEquals(QUOTA_PROJECT_FROM_ENVIRONMENT, credentials.getQuotaProjectId());
+
+    // Validate that if user sets quota, env and json value not used
+    credentials = credentials.toBuilder().setQuotaProjectId(QUOTA_PROJECT_EXPLICIT).build();
```
TimurSadykov, line 351: Add a test case where: 
- env variable for quota is set
- user creates credential from file himself, the file contains quota (ideally)
- check quota is not affected by env variable

oauth2_http/javatests/com/google/auth/oauth2/DefaultCredentialsProviderTest.java
sai-sunder-s: yeah wrote the test exactly like that first. Then realized that there is a cache. So we will get back same cred even if env var is set.

oauth2_http/javatests/com/google/auth/oauth2/DefaultCredentialsProviderTest.java
sai-sunder-s: createWithQuotaProject method needs to be called explicitly to set quota from env. With current implementation we need a environment provider to be sent explicitly for using custom env source.
Since currently there is no way to pass in the environmentprovider while class initialization, an effective test is not possible.

If we can use a mocking library instead like mentioned in https://stackoverflow.com/questions/8168884/how-to-test-code-dependent-on-environment-variables-using-junit, then such tests will be easy.
We can achieve same by adding EnvironmentProvider as a property/field in GoogleCredentials but then there will be a complicated change to add it to all builders, constructors etc. Not sure it is worth doing that.

oauth2_http/javatests/com/google/auth/oauth2/DefaultCredentialsProviderTest.java
TimurSadykov: What cache? Then this test is not effective?

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -184,6 +187,31 @@ public static GoogleCredentials fromStream(
...
+
+    return this;
+  }
+
+  /**
+   * Creates a credential with quota project from environment if present
+   *
+   * @return credential with quota project from envoronment
+   */
+  public GoogleCredentials createWithQuotaProject() {
```
TimurSadykov, line 211: This need to take the quota as a parameter and ADC should pass it from env. This should encapsulate checking the value and building the credential with it.

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -184,6 +187,31 @@ public static GoogleCredentials fromStream(
...
+  /**
+   * Creates a credential with quota project from environment if present
+   *
+   * @param environmentProvider a provider used to get environment values
+   * @return credential with quota project from envoronment
+   */
+  public GoogleCredentials createWithQuotaProject(EnvironmentProvider environmentProvider) {
+    String quotaProjectFromEnv = environmentProvider.getEnv(QUOTA_PROJECT_ENV_VAR);
+
+    if (quotaProjectFromEnv != null && !quotaProjectFromEnv.isEmpty()) {
```
TimurSadykov, line 199: This method is not needed

TimurSadykov (changes requested):

oauth2_http/javatests/com/google/auth/oauth2/DefaultCredentialsProviderTest.java
sai-sunder-s: Cache is [here](https://github.com/googleapis/google-auth-library-java/blob/main/oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java#L119). This test calls getDefaultCredentials only once. The second check is just seeing if user can update (new copy) the quota project on the credential to get the value they want.

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
sai-sunder-s: Doing only null check. If user wants to set an empty value, they should be able to I think.

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -211,7 +238,17 @@ protected GoogleCredentials() {
...
   public GoogleCredentials(AccessToken accessToken) {
+    this(accessToken, null);
+  }
+
+  /**
+   * Constructor with explicit access token and quota project.
+   *
+   * @param accessToken initial or temporary access token
+   */
+  public GoogleCredentials(AccessToken accessToken, @Nullable String quotaProjectId) {
```
TimurSadykov, line 249: you can remove this now we have createWithQuotaProject

TimurSadykov: Since you have to pass the value from every credential anyways, you better call the method. So you keep interface clean as it was.

sai-sunder-s: UserCredentials constructor is passed in quotaprojectId. So we need this constructor in base class.

oauth2_http/clirr-ignored-differences.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- see http://www.mojohaus.org/clirr-maven-plugin/examples/ignored-differences.html -->
+<differences>
+  <difference>
+    <differenceType>6001</differenceType>
+    <className>com/google/auth/oauth2/ExternalAccountCredentials/Builder</className>
```
chingor13, line 6:
```suggestion
    <className>com/google/auth/oauth2/ExternalAccountCredentials$Builder</className>
```

oauth2_http/javatests/com/google/auth/oauth2/DefaultCredentialsProviderTest.java
sai-sunder-s: Added Integration test for this as discussed

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -184,6 +187,20 @@ public static GoogleCredentials fromStream(
             fileType, USER_FILE_TYPE, SERVICE_ACCOUNT_FILE_TYPE));
   }
 
+  /**
+   * Creates a credential with quota project from environment if present
+   *
+   * @param environmentProvider a provider used to get environment values
+   * @return credential with quota project from envoronment
+   */
+  public GoogleCredentials createWithQuotaProject(String quotaProject) {
```
chingor13, line 196: This could use a unit test.

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -200,9 +217,23 @@ static Map<String, List<String>> addQuotaProjectIdToRequestMetadata(
     return Collections.unmodifiableMap(newRequestMetadata);
   }
 
+  @Override
+  protected Map<String, List<String>> getAdditionalHeaders() {
+    Map<String, List<String>> headers = super.getAdditionalHeaders();
+    if (quotaProjectId != null) {
```
chingor13, line 223: Should this use `this.getQuotaProjectId()`. What if a child class overrides how it stores/returns the quota project id?

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -211,7 +242,12 @@ protected GoogleCredentials() {
    * @param accessToken initial or temporary access token
    */
   public GoogleCredentials(AccessToken accessToken) {
-    super(accessToken);
+    this(accessToken, null);
+  }
+
+  public GoogleCredentials(Builder builder) {
```
chingor13, line 248: Does this need to be public?

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -232,6 +269,10 @@ public Builder toBuilder() {
     return new Builder(this);
   }
 
+  public String getQuotaProjectId() {
```
chingor13, line 272: Should this have an `@Override` annotation as it implements a interface method?

oauth2_http/javatests/com/google/auth/oauth2/functional/FTQuotaProjectId.java
```diff
@@ -0,0 +1,61 @@
...
+import com.google.auth.TestUtils;
+import com.google.auth.oauth2.GoogleCredentials;
+import java.io.IOException;
+import org.junit.Test;
+
+public final class FTQuotaProjectId {
+
+  @Test
+  public void adcQuotaFromEnv() throws IOException {
+    GoogleCredentials credentials = GoogleCredentials.getApplicationDefault();
```
chingor13, line 46: I'm not a big fan of this test where the code does not control the success of this test.

chingor13 (changes requested):

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
sai-sunder-s: Made it protected

oauth2_http/javatests/com/google/auth/oauth2/functional/FTQuotaProjectId.java
```diff
@@ -0,0 +1,60 @@
+/*
+ * Copyright 2021 Google LLC
```
TimurSadykov, line 2: nit: 2022

oauth2_http/clirr-ignored-differences.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- see http://www.mojohaus.org/clirr-maven-plugin/examples/ignored-differences.html -->
+<differences>
+  <difference>
+    <differenceType>6001</differenceType>
+    <className>com/google/auth/oauth2/ExternalAccountCredentials$Builder</className>
```
TimurSadykov, line 6: why we need to do this?

TimurSadykov (commented):
just a question on clirr

oauth2_http/javatests/com/google/auth/oauth2/functional/FTQuotaProjectId.java
sai-sunder-s: Removed. Moved this as an assert on env var to the other test

oauth2_http/javatests/com/google/auth/oauth2/functional/FTQuotaProjectId.java
```diff
@@ -0,0 +1,61 @@
...
+  public void explicitCredentialQuotaFromJson() throws IOException {
+    GenericJson json = new GenericJson();
+    json.put("client_id", "clientId");
+    json.put("client_secret", "clientSecret");
+    json.put("refresh_token", "refreshToken");
+    json.put("quota_project_id", "quota_from_file");
+    json.put("type", "authorized_user");
+    GoogleCredentials credentials = GoogleCredentials.fromStream(TestUtils.jsonToInputStream(json));
+    assertEquals("quota_from_file", credentials.getQuotaProjectId());
+  }
```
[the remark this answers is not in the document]
sai-sunder-s: Renamed test as Discussed offline

oauth2_http/clirr-ignored-differences.xml
sai-sunder-s: `protected quotaProjectId` was removed from builder of external account creds because a `protected quotaProjectId` was added to its base class. Got a build error saying change in public interface but we really did not change. Same thing is available from parent. So adding exception,

TimurSadykov: Got it, i was not sure why it is only that credential. Looks good.

oauth2_http/javatests/com/google/auth/oauth2/functional/FTQuotaProjectId.java
TimurSadykov: Sorry, for a late comment. I think we better restore the original test. It is not perfect but it adds value by validating the quota project override behavior "in the wild".

The second test only checks that override does not work for explicit value.

TimurSadykov (approved):
auth bits looks good

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -184,6 +187,20 @@ public static GoogleCredentials fromStream(
...
   }
 
+  /**
+   * Creates a credential with quota project from environment if present
+   *
+   * @param environmentProvider a provider used to get environment values
+   * @return credential with quota project from envoronment
+   */
+  public GoogleCredentials createWithQuotaProject(String quotaProject) {
+    if (quotaProject == null) {
```
chingor13, line 197: Looking at the test, this behavior seems a bit error prone. Calling this will `null` with neither clear the entry or give me a new instance.

Either way, this behavior should be described in the javadoc.

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
chingor13: This new constructor still looks public

chingor13 (changes requested):

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
chingor13: nvm, I was looking at an old commit I guess

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
sai-sunder-s: Updated the behavior to set null for the quota project as well. This was lingering behavior from an itr where env var was read here. Thanks for catching

chingor13 (approved):

lsirac (approved):

PR #1088, opened 2022-11-10
fix: Revert race condition test temporarily

clundin25:


oauth2_http/javatests/com/google/auth/oauth2/OAuth2CredentialsTest.java
```diff
@@ -876,91 +872,6 @@ public void serialize() throws IOException, ClassNotFoundException {
     assertSame(deserializedCredentials.clock, Clock.SYSTEM);
   }
 
-  @Test
```
TimurSadykov, line 879: No need to delete, you can just add the @Ignore attribute temporarily

TimurSadykov (changes requested):
let's not delete )

PR #1089, opened 2022-11-10
fix: empty string check for aws url validation

sai-sunder-s:
Validate aws metadata urls only if they are not null or empty. It is possible to fetch tokens successfully even if metadata urls are not provided but required environment variables are set. If both metadata urls and env vars are not provided, we will throw an exception.

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/AwsCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/AwsCredentials.java
@@ -137,18 +137,21 @@ public class AwsCredentials extends ExternalAccountCredentials {
       validateMetadataServerUrlIfAny(this.imdsv2SessionTokenUrl, "imdsv2_session_token_url");
     }
 
-    private static void validateMetadataServerUrlIfAny(String urlString, String nameInConfig) {
-      if (urlString != null && !urlString.isEmpty()) {
-        try {
-          URL url = new URL(urlString);
-          String host = url.getHost();
-          if (!host.equals("169.254.169.254") && !host.equals("[fd00:ec2::254]")) {
-            throw new IllegalArgumentException(
-                String.format("Invalid host %s for %s.", host, nameInConfig));
-          }
-        } catch (MalformedURLException malformedURLException) {
-          throw new IllegalArgumentException(malformedURLException);
+    @VisibleForTesting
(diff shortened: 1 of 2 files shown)


TimurSadykov (changes requested):

oauth2_http/java/com/google/auth/oauth2/AwsCredentials.java
```diff
@@ -137,18 +137,21 @@ private void validateMetadataServerUrls() {
...
+    @VisibleForTesting
+    static void validateMetadataServerUrlIfAny(String urlString, String nameInConfig) {
+      if (urlString == null || urlString.trim().length() == 0) {
+        return;
+      }
+
+      try {
+        URL url = new URL(urlString);
+        String host = url.getHost();
+        if (!host.equals("169.254.169.254") && !host.equals("[fd00:ec2::254]")) {
```
TimurSadykov, line 149: can host have different casing?

oauth2_http/javatests/com/google/auth/oauth2/AwsCredentialsTest.java
```diff
@@ -489,6 +501,66 @@ public void getAwsSecurityCredentials_fromEnvironmentVariablesWithToken() throws
...
+    AwsSecurityCredentials credentials =
+        testAwsCredentials.getAwsSecurityCredentials(EMPTY_METADATA_HEADERS);
+
+    assertEquals("awsAccessKeyId", credentials.getAccessKeyId());
+    assertEquals("awsSecretAccessKey", credentials.getSecretAccessKey());
+    assertEquals("awsSessionToken", credentials.getToken());
+  }
+
+  @Test
+  public void validateMetadataServerUrlIfAny_validUrls() {
```
TimurSadykov, line 528: validOrEmptyUrls

oauth2_http/java/com/google/auth/oauth2/AwsCredentials.java
sai-sunder-s: Apparently both lower and upper case are acceptable but as per RFC, they should be lower case https://www.rfc-editor.org/rfc/rfc5952#section-4.3

TimurSadykov (approved):
LGTM, except i would allow uppercase host

Issue #1001, opened 2022-09-19
Improve GCE residency check by leveraging SMBIOS 


TimurSadykov:
Current GCE residency check depends on network stack initialization and therefore error-prone.
With this fix, library will check unique environment variables to detect GCE residency.

For more details: go/auth-library-noping

PR #1092, opened 2022-11-11
feat: add smbios check for GCE residency detection

TimurSadykov:
Current GCE residency check depends on network stack initialization and therefore error-prone.
With this fix, library will check unique environment variables to detect GCE residency.

For more details: go/auth-library-noping

Fixes https://github.com/googleapis/google-auth-library-java/issues/1001


oauth2_http/javatests/com/google/auth/oauth2/DefaultCredentialsProviderTest.java
```diff
@@ -176,6 +181,64 @@ public void getDefaultCredentials_noCredentials_singleGceTestRequest() {
...
+    testProvider.setProperty("os.name", "Linux");
+    String productFilePath = SMBIOS_PATH_LINUX;
+    InputStream productStream = new ByteArrayInputStream("test".getBytes());
+    testProvider.addFile(productFilePath, productStream);
+
+    assertFalse(ComputeEngineCredentials.checkStaticGceDetection(testProvider));
+  }
+
+  @Test
+  public void getDefaultCredentials_static_linux() throws IOException {
```
sai-sunder-s, line 196: These tests could be simplified to a single test with a loop going over a map of inputs.

sai-sunder-s (changes requested):

oauth2_http/javatests/com/google/auth/oauth2/DefaultCredentialsProviderTest.java
TimurSadykov: in this case explicit test cases preferred for readability

clundin25 (approved):

oauth2_http/javatests/com/google/auth/oauth2/DefaultCredentialsProviderTest.java
sai-sunder-s: I think it will still be easy to understand if we have keyvalue pair of input and expected output. But up to you.

sai-sunder-s (approved):

PR #1100, opened 2022-11-17
fix: AwsCredentials should not call metadata server if security creds and region are retrievable through environment vars

lsirac:
Fixes an issue introduced by the AWS IMDSv2 changes where the metadata server is being called even if everything is retrievable through the defined environment variables.

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/AwsCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/AwsCredentials.java
@@ -63,6 +63,13 @@ import javax.annotation.Nullable;
  */
 public class AwsCredentials extends ExternalAccountCredentials {
 
+  // Supported environment variables.
+  static final String AWS_REGION = "AWS_REGION";
+  static final String AWS_DEFAULT_REGION = "AWS_DEFAULT_REGION";
+  static final String AWS_ACCESS_KEY_ID = "AWS_ACCESS_KEY_ID";
+  static final String AWS_SECRET_ACCESS_KEY = "AWS_SECRET_ACCESS_KEY";
+  static final String AWS_SESSION_TOKEN = "AWS_SESSION_TOKEN";
+
   static final String AWS_IMDSV2_SESSION_TOKEN_HEADER = "x-aws-ec2-metadata-token";
   static final String AWS_IMDSV2_SESSION_TOKEN_TTL_HEADER = "x-aws-ec2-metadata-token-ttl-seconds";
   static final String AWS_IMDSV2_SESSION_TOKEN_TTL = "300";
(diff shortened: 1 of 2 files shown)


oauth2_http/java/com/google/auth/oauth2/AwsCredentials.java
```diff
@@ -266,6 +270,39 @@ private String buildSubjectToken(AwsRequestSignature signature)
...
+    }
+    return false;
+  }
+
+  private boolean canRetrieveSecurityCredentialsFromEnvironment() {
+    // Check if both AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY are available.
+    List<String> keys = ImmutableList.of("AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY");
+    for (String env : keys) {
+      String value = getEnvironmentProvider().getEnv(env);
+      if (value == null || value.isEmpty()) {
```
sai-sunder-s, line 292: maybe do value==null || value.trim().length() > 0 to handle whitespace strings?

oauth2_http/java/com/google/auth/oauth2/AwsCredentials.java
```diff
@@ -266,6 +270,39 @@ private String buildSubjectToken(AwsRequestSignature signature)
...
+        // Region available.
+        return true;
+      }
+    }
+    return false;
+  }
+
+  private boolean canRetrieveSecurityCredentialsFromEnvironment() {
+    // Check if both AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY are available.
+    List<String> keys = ImmutableList.of("AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY");
```
sai-sunder-s, line 289: Make these env var names constants since they are used in multiple places now and maybe can refer to these constants in tests as well.

sai-sunder-s (changes requested):

oauth2_http/java/com/google/auth/oauth2/AwsCredentials.java
lsirac: I made them constant but did not change the tests. If someone changes them its better that the tests fail.

oauth2_http/javatests/com/google/auth/oauth2/AwsCredentialsTest.java
```diff
@@ -774,6 +862,98 @@ public void credentialSource_missingRegionalCredVerificationUrl() {
...
+                  .setHttpTransportFactory(transportFactory)
+                  .setCredentialSource(buildAwsImdsv2CredentialSource(transportFactory))
+                  .setEnvironmentProvider(environmentProvider)
+                  .build();
+      assertTrue(awsCredential.shouldUseMetadataServer());
+    }
+  }
+
+  @Test
+  public void shouldUseMetadataServer_missingAwsSecretAccessKey() {
```
sai-sunder-s, line 934: Consider adding test when both accesskeyid and secretaccesskey are missing. Also when all 4 env vars are missing.

sai-sunder-s (approved):

PR #1107, opened 2022-11-30
feat: Expose scopes granted by user

sai-sunder-s:


oauth2_http/java/com/google/auth/oauth2/AccessToken.java
```diff
@@ -43,6 +43,7 @@ public class AccessToken implements Serializable {
 
   private final String tokenValue;
   private final Long expirationTimeMillis;
+  private final String scopes;
```
TimurSadykov, line 46: Should this be a List<String>?

oauth2_http/javatests/com/google/auth/oauth2/UserAuthorizerTest.java
```diff
@@ -219,10 +226,69 @@ public void getCredentials_nullUserId_throws() throws IOException {
...
+            .setAccessToken(accessToken1)
+            .setHttpTransportFactory(transportFactory)
+            .build();
+
+    authorizer.storeCredentials(USER_ID, originalCredentials);
+
+    UserCredentials credentials1 = authorizer.getCredentials(USER_ID);
+
+    assertEquals(REFRESH_TOKEN, credentials1.getRefreshToken());
+    assertEquals(accessTokenValue1, credentials1.getAccessToken().getTokenValue());
```
TimurSadykov, line 261: check there were no scopes before?

TimurSadykov (approved):
LGTM

oauth2_http/java/com/google/auth/oauth2/UserAuthorizer.java
```diff
@@ -343,15 +360,22 @@ public void storeCredentials(String userId, UserCredentials credentials) throws
...
     AccessToken accessToken = credentials.getAccessToken();
     String acessTokenValue = null;
+    String scopes = null;
     Date expiresBy = null;
     if (accessToken != null) {
       acessTokenValue = accessToken.getTokenValue();
       expiresBy = accessToken.getExpirationTime();
+      List<String> grantedScopes = accessToken.getScopes();
+
+      if (grantedScopes != null) {
```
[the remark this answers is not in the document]
sai-sunder-s: List can be serialized but while deserializing it gets complicated. It is not possible to do an instance of check for List<String>. I get this error: "Cannot perform instanceof check against parameterized type List<String>. Use the form List<?> instead since further generic type information will be erased at runtime"

There are ways to deserialize like using Gson and seems to be another library called jackson that can help. But there is no need to complicate I think.

oauth2_http/javatests/com/google/auth/oauth2/UserAuthorizerTest.java
sai-sunder-s: There would be scopes before as well. They come as part of both first token response and future refresh responses as well.

oauth2_http/java/com/google/auth/oauth2/UserAuthorizer.java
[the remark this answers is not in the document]
TimurSadykov: Its just had to be List ) Generic does not exist at runtime )

PR #1125, opened 2023-01-07
fix: accessToken scopes clean serialization and default as empty list

TimurSadykov:
Removing extra processing from access token copes serialization 
Making empty list of scopes as default, no null

The original change that introduces Scopes into AccessToken [is not shipped yet](https://github.com/googleapis/google-auth-library-java/pull/1119), therefore this change does not change the existing behavior


oauth2_http/java/com/google/auth/oauth2/UserAuthorizer.java
```diff
@@ -205,8 +206,8 @@ public UserCredentials getCredentials(String userId) throws IOException {
     Long expirationMillis =
         OAuth2Utils.validateLong(tokenJson, "expiration_time_millis", TOKEN_STORE_ERROR);
     Date expirationTime = new Date(expirationMillis);
-    String scopes =
-        OAuth2Utils.validateOptionalString(
+    List<String> scopes =
+        OAuth2Utils.validateOptionalListString(
```
wangyutongg, line 210: For the existing token with string-typed scopes in the token store, this validation will fail, and they need to refresh to get a new access token?

TimurSadykov: The feature is not shipped yet, it was checked in late last year. So we don't change the behavior yet.

TimurSadykov: added this detail to the change description

oauth2_http/java/com/google/auth/oauth2/OAuth2Utils.java
```diff
@@ -168,6 +169,20 @@ static String validateOptionalString(Map<String, Object> map, String key, String
...
+      Map<String, Object> map, String key, String errorPrefix) throws IOException {
+    Object value = map.get(key);
+    if (value == null) {
+      return null;
+    }
+    if (!(value instanceof List)) {
+      throw new IOException(
+          String.format(VALUE_WRONG_TYPE_MESSAGE, errorPrefix, "List<String>", key));
+    }
+    return (List<String>) value;
```
sai-sunder-s, line 183: There is no type check for `List<String>`. Check is for `List`. What happens if `value` is something like `List<Int>`

sai-sunder-s (approved):

oauth2_http/java/com/google/auth/oauth2/UserAuthorizer.java
wangyutongg: This makes sense as we do not need to worry about breaking existing customers.

wangyutongg (approved):
LGTM

PR #1131, opened 2023-01-19
fix: ensure both refreshMargin and expirationMargin are set when using OAuth2CredentialsWithRefresh

lsirac:
Fixes b/263058140

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
@@ -152,12 +152,14 @@ public class OAuth2Credentials extends Credentials {
   }
 
   /** Returns the credentials' refresh margin. */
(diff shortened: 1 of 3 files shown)


oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
```diff
@@ -151,6 +151,16 @@ public final AccessToken getAccessToken() {
     return null;
   }
 
+  /** Returns the credentials' refresh margin. */
```
TimurSadykov, line 154: We don't need those, except for testing? VisibleForTesting attribute for variables is preferable. We better limit the public surface unless required.

TimurSadykov (changes requested):
overall LGTM, just a comment on public properties

TimurSadykov (approved):

PR #1140, opened 2023-01-26
chore: add joint ownership on 3pi files to aion sdk

TimurSadykov:
add joint ownership on 3pi files to aion sdk


.github/sync-repo-settings.yaml
```diff
@@ -81,3 +81,5 @@ permissionRules:
     permission: admin
   - team: googleapis-auth
     permission: push
+  - team: aion-sdk
```
TimurSadykov, line 84: i wonder why this shows... as it is already in main

PR #1146, opened 2023-02-03
feat: Add PKCE to 3LO exchange.

clundin25:
go/client-auth-java-oauth-pkce

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/DefaultPKCEProvider.java
+++ b/oauth2_http/java/com/google/auth/oauth2/DefaultPKCEProvider.java
@@ -39,7 +39,7 @@ import java.util.Base64;
 public class DefaultPKCEProvider implements PKCEProvider {
   private String codeVerifier;
   private CodeChallenge codeChallenge;
-  final int MAX_CODE_VERIFIER_LENGTH = 127;
+  private static final int MAX_CODE_VERIFIER_LENGTH = 127;
 
   private class CodeChallenge {
     private String codeChallenge;
--- a/oauth2_http/java/com/google/auth/oauth2/DefaultPKCEProvider.java
+++ b/oauth2_http/java/com/google/auth/oauth2/DefaultPKCEProvider.java
@@ -1,0 +1,103 @@
+/*
+ * Copyright 2023, Google Inc. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met:
+ *
+ *    * Redistributions of source code must retain the above copyright
+ * notice, this list of conditions and the following disclaimer.
+ *    * Redistributions in binary form must reproduce the above
+ * copyright notice, this list of conditions and the following disclaimer
+ * in the documentation and/or other materials provided with the
+ * distribution.
+ *
+ *    * Neither the name of Google Inc. nor the names of its
+ * contributors may be used to endorse or promote products derived from
+ * this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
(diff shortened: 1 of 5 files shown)


oauth2_http/java/com/google/auth/oauth2/UserAuthorizer.java
```diff
@@ -181,6 +185,10 @@ public URL getAuthorizationUrl(String userId, String state, URI baseUri) {
       url.put("login_hint", userId);
     }
     url.put("include_granted_scopes", true);
+    if (pkce != null) {
```
wangyutongg, line 188: Do we need to check both `pkce.getCodeChallenge()` and `pkce.getCodeChallengeMethod()` not null for the case of user passing in their own PKCEProvider

oauth2_http/java/com/google/auth/oauth2/UserAuthorizer.java
```diff
@@ -248,6 +256,11 @@ public UserCredentials getCredentialsFromCode(String code, URI baseUri) throws I
     tokenData.put("client_secret", clientId.getClientSecret());
     tokenData.put("redirect_uri", resolvedCallbackUri);
     tokenData.put("grant_type", "authorization_code");
+
+    if (pkce != null) {
```
wangyutongg, line 260: Same. Check `pkce.getCodeVerifier()` not null

oauth2_http/javatests/com/google/auth/oauth2/UserAuthorizerTest.java
```diff
@@ -148,6 +149,7 @@ public void getAuthorizationUrl() throws IOException {
             .setScopes(DUMMY_SCOPES)
             .setCallbackUri(CALLBACK_URI)
             .setUserAuthUri(AUTH_URI)
+            .setPKCEProvider(pkce)
```
wangyutongg, line 152: Can we add a test case of without calling `setPKCEProvider()`, which should be the majority of existing customers' usage? We might not be able to check the code_challenge value, but can validate it empty or not.

oauth2_http/java/com/google/auth/oauth2/UserAuthorizer.java
clundin25: If those methods return NULL should the program continue to operate or throw an exception? 

Seems erroneous to have provided a PKCEProvider that is returning NULL.

oauth2_http/javatests/com/google/auth/oauth2/UserAuthorizerTest.java
clundin25: The rest of the test cases are using PKCE because the builder will build the default PKCE provider if one isn't injected.

This test case is different because I want to verify against a well known PKCE object.

oauth2_http/java/com/google/auth/oauth2/UserAuthorizer.java
wangyutongg: If a user provides custom PKCEProvider, I think we should validate it and throw an exception. Because we know if they pass in a PKCEProvider, they intend to use PKCE in fetch token call. Silently using plain or skipping PKCE in case of erroneous PKCEProvider will give them false impression of successful PKCE call. On another thought, do we need to give users the capability to pass in their own PKCEProvider? Maybe we can follow the Python implementation to always use DefaultPKCEProvider.

clundin25: I think it is still useful in case someone needs to disable it.

I've updated the code to include a null check in the builder

wangyutongg (approved):
LGTM

oauth2_http/java/com/google/auth/oauth2/DefaultPKCEProvider.java
```diff
@@ -0,0 +1,103 @@
...
+import java.security.NoSuchAlgorithmException;
+import java.security.SecureRandom;
+import java.util.Base64;
+
+public class DefaultPKCEProvider implements PKCEProvider {
+  private String codeVerifier;
+  private CodeChallenge codeChallenge;
+  private static final int MAX_CODE_VERIFIER_LENGTH = 127;
+
+  private class CodeChallenge {
```
TimurSadykov, line 44: nit: Helpers should go below

oauth2_http/javatests/com/google/auth/oauth2/UserAuthorizerTest.java
```diff
@@ -471,4 +475,33 @@ public void revokeAuthorization_revokesAndClears() throws IOException {
     UserCredentials credentials2 = authorizer.getCredentials(USER_ID);
     assertNull(credentials2);
   }
+
+  @Test(expected = IllegalArgumentException.class)
```
TimurSadykov, line 479: This does not test all the properties, only one.

oauth2_http/java/com/google/auth/oauth2/DefaultPKCEProvider.java
```diff
@@ -0,0 +1,103 @@
...
+ */
+
+package com.google.auth.oauth2;
+
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.security.SecureRandom;
+import java.util.Base64;
+
+public class DefaultPKCEProvider implements PKCEProvider {
```
TimurSadykov, line 39: missing javadoc: what is a default PKCE provider ) with a link to a related google docs

oauth2_http/java/com/google/auth/oauth2/DefaultPKCEProvider.java
TimurSadykov: and missing javadoc

TimurSadykov (commented):
Some comments

PR #1150, opened 2023-02-07
fix: Removed url pattern validation for google urls in external account credential configurations

aeitzman:
Removed pattern checking for external accounts to support TPC urls. Also added readme change to explain risk to users.


lsirac (approved):

TimurSadykov (approved):
LGTM

PR #1154, opened 2023-02-12
fix: create and reuse self signed jwt creds for better performance

arithmetic1728:
GAPIC clients all use self signed jwt (with scopes) now. Currently we are creating a new jwt credential each time, and this causes perf issue, similar to https://github.com/googleapis/google-cloud-java/issues/3149.

In this PR we create one jwt cred and reuse it each time.
Internal bug:  b/268603529

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
@@ -145,12 +145,6 @@ public class ServiceAccountCredentials extends GoogleCredentials
     this.lifetime = builder.lifetime;
     this.useJwtAccessWithScope = builder.useJwtAccessWithScope;
     this.defaultRetriesEnabled = builder.defaultRetriesEnabled;
-
-    // Create a jwt credential for self signed jwt with scopes, and reuse it to improve the
-    // performance. For example see https://github.com/googleapis/google-cloud-java/issues/3149.
-    if (!createScopedRequired() && this.useJwtAccessWithScope) {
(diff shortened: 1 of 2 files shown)


oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -935,8 +948,8 @@ public Map<String, List<String>> getRequestMetadata(URI uri) throws IOException
     // Otherwise, use self signed JWT with uri as the audience.
     JwtCredentials jwtCredentials;
     if (!createScopedRequired() && useJwtAccessWithScope) {
-      // Create JWT credentials with the scopes.
-      jwtCredentials = createSelfSignedJwtCredentials(null);
+      // Reuse selfSignedJwtCredentialsWithScope to improve the performance.
+      jwtCredentials = selfSignedJwtCredentialsWithScope;
```
wangyutongg, line 952: If the `selfSignedJwtCredentialsWithScope` expired (credential created 60m+ ago), the API call with the JWT will fail. How does the user trigger recreating a new JWT? Do they have to create a new credential? If that the case, we might want to check `exp` of the JWT here, and create a new one and store if the JWT expired or close to expiration.

arithmetic1728: The `selfSignedJwtCredentialsWithScope` is a JwtCredential object. At line 957 when its `getRequestMetadata`  is called, and it refreshes its token automatically:

https://github.com/googleapis/google-auth-library-java/blob/main/oauth2_http/java/com/google/auth/oauth2/JwtCredentials.java#L161-L164

TimurSadykov (approved):
LGTM

PR #1162, opened 2023-02-17
ci: configure the protected branch

TimurSadykov:
Configures CI for branch


suztomo (approved):
Build failure is strange. Can you check the errors there?

TimurSadykov:
> Build failure is strange. Can you check the errors there?

That test is unstable and later was disabled

PR #1164, opened 2023-02-17
fix: create and reuse self signed jwt creds for better performance 

TimurSadykov:
* fix: create and reuse self signed jwt creds for better performance
* only create jwt cred when needed


suztomo (approved):

suztomo:
You may want to comment (or in description) the original pull request that was merged to main branch.

PR #1172, opened 2023-02-22
fix: improve errors and warnings related to ADC

TimurSadykov:
This change is part of the bigger work [go/simplified-local-adc](http://goto.google.com/simplified-local-adc)

Here we improve couple errors: 

1. When user does not have credential setup, we do not recommend setting up service account, rather we redirect to a new ADC setup page. 
2. We show warning about user credentials only if production environment is detected


wangyutongg (approved):
LGTM

wangyutongg:
Nit: The second item in the description needs to be removed if we only fix the missing credentials error in this PR

TimurSadykov:
> Nit: The second item in the description needs to be removed if we only fix the missing credentials error in this PR

Updated the PR description, this PR covers both items

TimurSadykov:
> Nit: The second item in the description needs to be removed if we only fix the missing credentials error in this PR

PTAL latest version - left the second part as we decided to keep it

oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java
```diff
@@ -74,12 +72,13 @@ class DefaultCredentialsProvider {
...
       "764086051850-6qr4p6gpi6hn506pt8ejuq83di341hur.apps.googleusercontent.com";
   static final String CLOUDSDK_CREDENTIALS_WARNING =
-      "Your application has authenticated using end user credentials from Google "
-          + "Cloud SDK. We recommend that most server applications use service accounts "
-          + "instead. If your application continues to use end user credentials from Cloud "
-          + "SDK, you might receive a \"quota exceeded\" or \"API not enabled\" error. For "
-          + "more information about service accounts, see "
-          + "https://cloud.google.com/docs/authentication/.";
+      "You are authenticating using user credentials. "
+          + "For production, we recommend using service account credentials.";
```
wangyutongg, line 76: WARNING: You are authenticating using user credentials. For production, we recommend using service account credentials.

To learn more about service account credentials, see http://cloud.google.com/docs/authentication/external/set-up-adc-on-cloud

oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java
```diff
@@ -74,12 +72,13 @@ class DefaultCredentialsProvider {
...
+
+  static final String CLOUDSDK_MISSING_CREDENTIALS =
+      "Your default credentials were not found. To set up Application Default Credentials, "
+          + "see https://cloud.google.com/docs/authentication/external/set-up-adc "
+          + "for more information.";
```
wangyutongg, line 81: ERROR: Your credentials were not found. To set up Application Default Credentials for your environment, see https://cloud.google.com/docs/authentication/external/set-up-adc

oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java
TimurSadykov: Thanks, did not notice the update

oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java
TimurSadykov: Thanks, did not notice the update from the previous doc

PR #1079, opened 2022-11-04
fix: Validate url domain for aws metadata urls

sai-sunder-s:
Updating AWS credential source validation as per new updates in [AIP](https://github.com/aip-dev/google.aip.dev/pull/959). Make sure the host of url, region_url and imdsv2 session token url belong to AWS metadata server.

Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [ ] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/google-auth-library-java/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [ ] Ensure the tests and linter pass
- [ ] Code coverage does not decrease (if any source code was changed)
- [ ] Appropriate docs were updated (if necessary)

Fixes #<issue_number_goes_here> ☕️

If you write sample code, please follow the [samples format](
https://github.com/GoogleCloudPlatform/java-docs-samples/blob/main/SAMPLE_FORMAT.md).

PR #1177, opened 2023-03-02
fix: removes AWS credential source validation.

lsirac:
Removes the changes introduced by https://github.com/googleapis/google-auth-library-java/pull/1079.


TimurSadykov (approved):
LGTM, maybe also mention it in readme? Like previous removed validations

TimurSadykov:
Let's not merge until integration is fixed

PR #1179, opened 2023-03-09
fix: separating out nested classes to avoid import issue

TimurSadykov:


TimurSadykov:
no longer needed until next major bump

PR #1182, opened 2023-03-14
feat: option to clear cached application default credentials

### LINKED ISSUE #1181: `DefaultCredentialsProvider` should have an option to clear cached Credentials
Hi, we wish to integrate [Cloud Tools for Eclipse](https://github.com/GoogleCloudPlatform/google-cloud-eclipse) with this library in order to obtain the credentials from the `gcloud CLI`'s well-known location/env-var. Our use case is to switch accounts externally via the gcloud CLI (i.e. `login, revoke, login`), but this will not work if the second `login` is with a different account thanks to a [cache hit](https://github.com/googleapis/google-auth-library-java/blob/77a99c9cfab3c1ce2db50c92e89fc292efaeb3ab/oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java#L118-L124) in the `DefaultCredentialsProvider` class. We need an option to clear the cache manually.
Thanks

diegomarquezp:
fixes #1181 
Adds a public method in `GoogleCredentials` to clear the cached credentials of `DefaultCredentialsProvider`

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java
+++ b/oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java
@@ -32,6 +32,8 @@
 package com.google.auth.oauth2;
 
 import com.google.auth.http.HttpTransportFactory;
+import com.google.common.annotations.VisibleForTesting;
+
 import java.io.File;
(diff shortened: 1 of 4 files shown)


TimurSadykov:
i guess we don't need this anymore

PR #1188, opened 2023-03-29
feat: expose method to manually obtain ADC from gcloud CLI well-known…

diegomarquezp:
Second attempt to fix https://github.com/googleapis/google-auth-library-java/issues/1181

This time using an approach validated by @TimurSadykov 

It exposes a method in `GoogleCredentials` to manually obtain ADC from the well-known file ([docs](https://cloud.google.com/docs/authentication/application-default-credentials#personal))


oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java
```diff
@@ -282,6 +259,43 @@ private boolean runningOnAppEngine() {
         cause);
   }
 
+  final GoogleCredentials getFromGcloudCliWellKnownFile(HttpTransportFactory transportFactory)
+      throws IOException {
+    return tryGetFromWellKnownCredentialsFile(transportFactory);
+  }
+
+  @VisibleForTesting
+  final GoogleCredentials tryGetFromWellKnownCredentialsFile(HttpTransportFactory transportFactory)
```
TimurSadykov, line 268: IIRC, we agreed on exposing the logic that returns the path, not the actual credential. This logic has some ADC-specific stuff, like printing warnings about user credentials, that you don't need.

TimurSadykov: Seems like the getWellKnownCredentialsFile is what you need, or new method that encapsulates just the Path part of the File creation.

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -101,6 +101,24 @@ public static GoogleCredentials getApplicationDefault() throws IOException {
...
+   * <ol>
+   *   <li>{@code gcloud auth application-default login} for user account credentials.
+   *   <li>{@code gcloud auth application-default login --impersonate-service-account} for
+   *       impersonated service account credentials.
+   * </ol>
+   *
+   * @return the credentials, or null if not found
+   * @throws IOException if the credentials cannot be created in the current environment.
+   */
+  public static GoogleCredentials getFromGcloudCliWellKnownFile() throws IOException {
```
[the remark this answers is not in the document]
TimurSadykov: We don't want to add public method here unless it has a wide use case.

TimurSadykov (changes requested):
Sorry for late comments, did not submit earlier (

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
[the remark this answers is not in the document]
diegomarquezp: @TimurSadykov I modified the PR to expose the path only, using `DefaultCredentialProvider.getWellKnownCredentialsFile().getAbsolutePath()`. Our use case requires accessing a public method of this library

oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java
diegomarquezp: I exposed a method to wrap `getWellKnownCredentialsFile`'s path.

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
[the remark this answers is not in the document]
diegomarquezp: @TimurSadykov If not a public method in `GoogleCredentials`, where would it make most sense to expose the path method?

oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java
TimurSadykov: Yep, that is better

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -101,6 +101,21 @@ public static GoogleCredentials getApplicationDefault() throws IOException {
...
+   * =======
+   * Returns the gcloud CLI's well-known location file. Corresponds to one of the elements of {@code
+   * >>>>>>> 608179607cce89b7d8c3271ffd0d5a078a457255
+   * getApplicationDefault()}, exposed for convenience.
+   *
+   * @return the credentials, or null if not found
+   * @throws IOException if the credentials cannot be created in the current environment.
+   */
+  public static String getWellKnownCredentialsPath() {
+    return defaultCredentialsProvider.getWellKnownCredentialsPath();
```
TimurSadykov, line 116: This is sketchy. Because here we rely on the fact that default credentials provider is static and always the same. This could change and if we decide to change it - this method will makes things more complicated. 

How about you move the logic of the getWellKnownCredentialsPath into a new public  Utils class, like GoogleAuthUtils. (all the existing utils classes are internal and therefore private) We already have some other public static methods that could be moved to the new Utils class as well.

oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java
TimurSadykov: You probably want to revert this change (as no longer needed) or cover it with tests... either works

TimurSadykov (changes requested):
Suggested an alternative

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
diegomarquezp: I created a `GoogleAuthUtils` with a `getPath` method that would use the default provider instance to resolve the path. It also has internal methods to use a test provider

oauth2_http/java/com/google/auth/oauth2/GoogleAuthUtils.java
```diff
@@ -0,0 +1,77 @@
+/*
+ * Copyright 2015, Google Inc. All rights reserved.
```
TimurSadykov, line 2: nit: 2023

TimurSadykov: please double check if contents are the same in more recent examples. Should be the same

oauth2_http/java/com/google/auth/oauth2/GoogleAuthUtils.java
```diff
@@ -0,0 +1,77 @@
...
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package com.google.auth.oauth2;
+
+import java.io.File;
+
+public class GoogleAuthUtils {
```
TimurSadykov, line 36: Please add java doc for the class.. like: 

"This public class provides shared utilities for for common OAuth2 utils or ADC like calculating well-known file path"

oauth2_http/java/com/google/auth/oauth2/GoogleAuthUtils.java
```diff
@@ -0,0 +1,77 @@
...
+
+  /**
+   * Gets the gcloud CLI path
```
TimurSadykov, line 39: This reads' like we get the path of the CLI itself, maybe use same line like below: "Gets a path to the well-known Application Default Credentials file location"

oauth2_http/java/com/google/auth/oauth2/GoogleAuthUtils.java
```diff
@@ -0,0 +1,77 @@
...
+   * @return the path to the well-known gcloud CLI credentials location
+   */
+  public static final String getWellKnownCredentialsPath() {
+    return getWellKnownCredentialsFile(DefaultCredentialsProvider.DEFAULT).getAbsolutePath();
+  }
+
+  /**
+   * Testing version of getWellKnownCredentialsPath() that uses a custom provider
+   *
+   * @return the path to the well-known gcloud CLI credentials location
```
TimurSadykov, line 50: Here we get the ADC path, which is written by gcloud most of the time, but "gcloud CLI credential" is a thing on its own and stored in local db. Please change to smth like "well-known Application Default Credentials file location"

oauth2_http/java/com/google/auth/oauth2/GoogleAuthUtils.java
```diff
@@ -0,0 +1,77 @@
...
+  static final String getWellKnownCredentialsPath(DefaultCredentialsProvider provider) {
+    return getWellKnownCredentialsFile(provider).getAbsolutePath();
+  }
+
+  /**
+   * Platform-independent logic to obtain the gcloud CLI credentials file
+   *
+   * @param provider the provider used to resolve env and system properties (exposed for testing
+   *     purposes)
+   * @return the gcloud CLI credentials file
```
TimurSadykov, line 61: same as above

oauth2_http/java/com/google/auth/oauth2/GoogleAuthUtils.java
```diff
@@ -0,0 +1,77 @@
...
+   * Testing version of getWellKnownCredentialsPath() that uses a custom provider
+   *
+   * @return the path to the well-known gcloud CLI credentials location
+   */
+  static final String getWellKnownCredentialsPath(DefaultCredentialsProvider provider) {
+    return getWellKnownCredentialsFile(provider).getAbsolutePath();
+  }
+
+  /**
+   * Platform-independent logic to obtain the gcloud CLI credentials file
```
TimurSadykov, line 57: same as above about CLI creds

TimurSadykov (changes requested):
Overall looks good, just comments

oauth2_http/java/com/google/auth/oauth2/GoogleAuthUtils.java
diegomarquezp: Using [statement](https://github.com/googleapis/google-auth-library-java/blob/6c2be858c4024cb0f14ab927af5ff388825cb3b3/oauth2_http/java/com/google/auth/oauth2/GdchCredentials.java#L1-L30) from `GdchCredentials.java` (2022)

oauth2_http/java/com/google/auth/oauth2/GoogleAuthUtils.java
diegomarquezp: Fixed this and related comments by referring to this path as the ADC well-known file/path

oauth2_http/java/com/google/auth/oauth2/GoogleAuthUtils.java
diegomarquezp: done

oauth2_http/java/com/google/auth/oauth2/GoogleAuthUtils.java
diegomarquezp: done

oauth2_http/java/com/google/auth/oauth2/GoogleAuthUtils.java
diegomarquezp: Solved

oauth2_http/javatests/com/google/auth/oauth2/GoogleAuthUtilsTest.java
```diff
@@ -0,0 +1,27 @@
+package com.google.auth.oauth2;
```
TimurSadykov, line 1: please add the copyright header

TimurSadykov (changes requested):
one nit request

TimurSadykov (approved):

PR #1224, opened 2023-06-20
test: expose test-jar and mock classes in oauth2

diegomarquezp:
We need to expose some test utils to be used in https://github.com/googleapis/sdk-platform-java/pull/1642

Basically to allow `GdchCredentials` to be refreshed via fake transport


appengine/pom.xml
```diff
@@ -75,8 +75,10 @@
     <dependency>
       <groupId>com.google.auth</groupId>
       <artifactId>google-auth-library-oauth2-http</artifactId>
-      <type>test-jar</type>
+      <version>${project.version}</version>
```
blakeli0, line 78: Do we need the version? I think it is already configured in the parent pom?

oauth2_http/javatests/com/google/auth/oauth2/MockHttpTransportFactory.java
```diff
@@ -0,0 +1,15 @@
+package com.google.auth.oauth2;
+
+import com.google.api.client.http.HttpTransport;
+import com.google.api.client.testing.http.MockHttpTransport;
+import com.google.auth.http.HttpTransportFactory;
+
+class MockHttpTransportFactory implements HttpTransportFactory {
```
blakeli0, line 7: If you plan to use it in other repos, this needs to be a public class.

blakeli0 (approved):

blakeli0 (changes requested):
Didn't mean to approve the PR, please resolve the comments before merging.

appengine/pom.xml
diegomarquezp: I removed it, thanks for the catch.

oauth2_http/javatests/com/google/auth/oauth2/MockHttpTransportFactory.java
diegomarquezp: Seems like it was the only exposed class that needed a modifier. Changed

TimurSadykov (approved):
LGTM

blakeli0 (approved):

PR #1225, opened 2023-06-21
feat: Add custom parameters to token body during user cred refresh

sai-sunder-s:
Providing a way for passing additional parameters in token body of a token refresh request for user credentials.


sai-sunder-s:
Introduced a way to provide custom parameters only to auth endpoint.

PR #1232, opened 2023-06-28
feat: Byoid metrics framework

aeitzman:
Adding metrics for BYOID credentials.


BigTailWolf (approved):

oauth2_http/java/com/google/auth/oauth2/MetricsUtils.java
```diff
@@ -0,0 +1,63 @@
...
+class MetricsUtils {
+  static final String API_CLIENT_HEADER = "x-goog-api-client";
+  private static final String libVersion = getLibVersion();
+
+  static String getAuthAndLibVersion() {
+    return String.format("gl-java/%s auth/%s", System.getProperty("java.version"), libVersion);
+  }
+
+  private static String getLibVersion() {
+    // attempt to read the library's version from a properties file generated during the build
```
lsirac, line 47:
```suggestion
    // Attempt to read the library's version from a properties file generated during the build.
```

oauth2_http/java/com/google/auth/oauth2/MetricsUtils.java
```diff
@@ -0,0 +1,63 @@
...
+    // this value should be read and cached for later use
```
lsirac, line 48:
```suggestion
    // This value should be read and cached for later use.
```

oauth2_http/java/com/google/auth/oauth2/MetricsUtils.java
```diff
@@ -0,0 +1,63 @@
...
+    try (InputStream inputStream =
+        MetricsUtils.class.getResourceAsStream(
+            "/com/google/auth/oauth2/google-auth-library.properties")) {
+      if (inputStream != null) {
+        final Properties properties = new Properties();
+        properties.load(inputStream);
+        version = properties.getProperty("google-auth-library.version");
+      }
+    } catch (IOException e) {
+      // ignore
```
lsirac, line 59:
```suggestion
      // Ignore.
```

oauth2_http/java/com/google/auth/oauth2/MetricsUtils.java
```diff
@@ -0,0 +1,63 @@
...
+
+class MetricsUtils {
+  static final String API_CLIENT_HEADER = "x-goog-api-client";
+  private static final String libVersion = getLibVersion();
+
+  static String getAuthAndLibVersion() {
+    return String.format("gl-java/%s auth/%s", System.getProperty("java.version"), libVersion);
+  }
+
+  private static String getLibVersion() {
```
lsirac, line 46:
```suggestion
  private static String getLibraryVersion() {
```

oauth2_http/java/com/google/auth/oauth2/MetricsUtils.java
```diff
@@ -0,0 +1,63 @@
...
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Properties;
+
+class MetricsUtils {
+  static final String API_CLIENT_HEADER = "x-goog-api-client";
+  private static final String libVersion = getLibVersion();
+
+  static String getAuthAndLibVersion() {
```
lsirac, line 42: Shouldn't this be getAuthLibAndJavaVersion? I'd split them into two methods and call this getVersions() or something.

Also why not have a var for the java version like you do for libVersion?

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
```diff
@@ -589,6 +608,10 @@ public ServiceAccountImpersonationOptions getServiceAccountImpersonationOptions(
     return serviceAccountImpersonationOptions;
   }
 
+  protected String getMetricsSource() {
```
lsirac, line 611: getMetricsSource() is kind of confusing.

getCredentialSourcingMethodForMetrics?

getCredentialSource()?

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
```diff
@@ -663,12 +686,16 @@ static final class ServiceAccountImpersonationOptions implements java.io.Seriali
 
     private final int lifetime;
 
+    private final boolean customLifetimeProvided;
+
     ServiceAccountImpersonationOptions(Map<String, Object> optionsMap) {
       if (!optionsMap.containsKey(TOKEN_LIFETIME_SECONDS_KEY)) {
         lifetime = DEFAULT_TOKEN_LIFETIME_SECONDS;
+        customLifetimeProvided = false;
```
lsirac, line 694: this can be changed to:
```
customTokenLifetimeRequested = optionsMap.containsKey(TOKEN_LIFETIME_SECONDS_KEY);
if (!customTokenLifetimeRequested) {
  lifetime = DEFAULT..
}
```

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
```diff
@@ -663,12 +686,16 @@ static final class ServiceAccountImpersonationOptions implements java.io.Seriali
 
     private final int lifetime;
 
+    private final boolean customLifetimeProvided;
```
lsirac, line 689:
```suggestion
    private final boolean customTokenLifetimeRequested;
```

oauth2_http/java/com/google/auth/oauth2/ByoidMetricsOptions.java
```diff
@@ -0,0 +1,65 @@
...
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package com.google.auth.oauth2;
+
+class ByoidMetricsHandler implements java.io.Serializable {
```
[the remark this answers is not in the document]
aeitzman: Had it named options in a previous iteration, whoops

oauth2_http/java/com/google/auth/oauth2/ByoidMetricsOptions.java
```diff
@@ -0,0 +1,65 @@
...
+  private static final String SOURCE_KEY = "source";
+  private static final String IMPERSONATION_KEY = "sa-impersonation";
+  private static final String CONFIG_LIFETIME_KEY = "config-lifetime";
+
+  private final boolean configLifetime;
+  private final boolean saImpersonation;
+  private String source;
+
+  ByoidMetricsHandler(boolean saImpersonation, boolean configLifetime, String source) {
```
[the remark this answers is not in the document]
aeitzman: Had it as a seperate class because we need some object to pass between ExternalAccountCredentials objects due to how we handle the impersonated credentials. Could just have it be a dictionary like in python and have getMetricsHeader be in ExternalAccountCredentials but I thought it made more sense in Java to have it be its own class/object because that seems closer to how things are done in this lib.

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
```diff
@@ -896,6 +925,17 @@ public Builder setUniverseDomain(String universeDomain) {
       return this;
     }
 
+    /**
+     * Sets the optional metrics handler.
+     *
+     * @param metricsHandler the ByoidMetricsHandler object to set.
+     * @return this {@code Builder} object
+     */
+    Builder setMetricsHandler(ByoidMetricsHandler metricsHandler) {
```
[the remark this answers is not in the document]
aeitzman: Yeah, need this for the tests.

oauth2_http/java/com/google/auth/oauth2/MetricsUtils.java
aeitzman: Change to authLib version to match the other method, is there a reason we need Library fully spelled out?

oauth2_http/java/com/google/auth/oauth2/MetricsUtils.java
aeitzman: Why should they be split into two methods? When we are constructing header values we will always want to be adding both the language and auth library versions at the same time. Also the language version is just a single method call anyway, so creating another method for that would just end up being a single line. I had the lib version in a static var to avoid calling the method over and over again because it does a file read, which the java version check does not. Added it as a var though now for consistency.

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
aeitzman: change to getCredentialSourceType()

oauth2_http/java/com/google/auth/oauth2/MetricsUtils.java
lsirac: I'm thinking getting the auth library version and java version might be things we want to do in the future, and we may not necessarily want them together, formatted like this. go/shield-java-coding#methods-should-do-one-thing. I am nit picking here

oauth2_http/java/com/google/auth/oauth2/MetricsUtils.java
lsirac: Personal preference but apparently its a thing! go/java-practices/naming#general-advice 

```
Avoid acronyms, abbreviations, and in-team jargon.
```

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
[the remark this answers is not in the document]
lsirac: why?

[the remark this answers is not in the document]
aeitzman: We do a weird thing with the tests where we create a credential at the start and then convert it to a builder, and add options. When we convert to a builder it takes the original options and sets them in the new credential (This is done to support the impersonation issue). To get the actual expected header I had to add this builder option to test the metrics header we actually want. The builder option is also package-private so won't show up to people who are just using the library to make credentials.

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
```diff
@@ -192,6 +193,16 @@ public String retrieveSubjectToken() throws IOException {
...
   }
 
+  @Override
+  protected String getMetricsSource() {
+    if (((IdentityPoolCredentialSource) this.getCredentialSource()).credentialSourceType
+        == IdentityPoolCredentialSourceType.FILE) {
+      return "file";
+    } else {
+      return "url";
+    }
```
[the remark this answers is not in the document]
aeitzman: Do you mean a ternary? I personally don't prefer them condition you are checking is long like this.

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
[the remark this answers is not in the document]
aeitzman: removed the builder and fixed tests

oauth2_http/java/com/google/auth/oauth2/ByoidMetricsHandler.java
```diff
@@ -0,0 +1,77 @@
...
+
+  /**
+   * Gets the BYOID metrics header value for the x-goog-api-client header.
+   *
+   * @return the header value.
+   */
+  String getByoidMetricsHeader() {
+    return String.format(
+        "%s %s %s/%s %s/%s %s/%s",
+        MetricsUtils.API_CLIENT_HEADER,
```
arithmetic1728, line 68: I think we don't need `MetricsUtils.API_CLIENT_HEADER` because this method returns the header value.

lsirac (approved):

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
```diff
@@ -90,6 +91,7 @@ abstract static class CredentialSource implements java.io.Serializable {
   private final CredentialSource credentialSource;
   private final Collection<String> scopes;
   private final ServiceAccountImpersonationOptions serviceAccountImpersonationOptions;
+  private ExternalAccountMetricsHandler metricsHandler;
```
TimurSadykov, line 94: is this handler really specific to external accounts creds or other creds will have similar thing with different config vars later? 

If the latter we might need to move this to GoogleCredentials later.

aeitzman: This handler is specific to external account creds (the header value it's building is only attached to the STS token exchange request for external creds) Other credentials will need header values for x-goog-api-client, but wouldn't be creating or using them in the same way.

TimurSadykov (approved):
LGTM

PR #1239, opened 2023-07-16
chore: Update library definitions in README to the latest version

Maxr1998:
I recently switched to this library for a project, and since I just copied the info from the readme I used an ancient version at first. That was especially annoying since many APIs that would've been helpful for me during migration didn't exist back in version 1.3.0 (like creating ServiceAccountCredentials from a PKCS8 string without having to pass JSON).

I'd love to prevent other users from making the same mistake as myself in the future through this README update.


TimurSadykov:
Thanks for updating this )

TimurSadykov (approved):
LGTM

PR #1282, opened 2023-09-12
feat: multi universe support, adding universe_domain field

TimurSadykov:
This adds a multi-universe support to all the credentials.
Also adds reading a universe_domain field from ServiceAccount and modifies it for ExternalAccounts

A related change is to manage (builder, toString/hashCode/equals) common fields: universeDomain and quotaProjectId in the base GoogleCredentials class so we don't manage them in all the Credentials like before.

### DIFF
--- a/credentials/java/com/google/auth/Credentials.java
+++ b/credentials/java/com/google/auth/Credentials.java
@@ -59,9 +59,13 @@ public abstract class Credentials implements Serializable {
   /**
    * Returns the universe domain for the credential.
    *
-   * @return the default Google universe domain googleapis.com
+   * @return a universe domain value in the format some-domain.xyz. By default, returns the Google
+   * universe domain googleapis.com.
+   * @throws IOException extending classes might have to do remote calls to determine the universe
+   * domain. The exception should implement {@link Retryable} and {@code isRetryable()} will return
+   * true if the operation may be retried.
    */
-  public String getUniverseDomain() {
+  public String getUniverseDomain() throws IOException {
     return GOOGLE_DEFAULT_UNIVERSE;
   }
 
--- a/credentials/java/com/google/auth/Credentials.java
+++ b/credentials/java/com/google/auth/Credentials.java
@@ -57,12 +57,13 @@ public abstract class Credentials implements Serializable {
   public abstract String getAuthenticationType();
 
   /**
-   * Returns the universe domain for the credential.
+   * Gets the universe domain for the credential in a blocking manner, refreshing tokens if
+   * required.
    *
    * @return a universe domain value in the format some-domain.xyz. By default, returns the Google
    *     universe domain googleapis.com.
    * @throws IOException extending classes might have to do remote calls to determine the universe
-   *     domain. The exception should implement {@link Retryable} and {@code isRetryable()} will
+   *     domain. The exception must implement {@link Retryable} and {@code isRetryable()} will
    *     return true if the operation may be retried.
    */
   public String getUniverseDomain() throws IOException {
--- a/credentials/java/com/google/auth/Credentials.java
+++ b/credentials/java/com/google/auth/Credentials.java
@@ -43,6 +43,8 @@ public abstract class Credentials implements Serializable {
 
   private static final long serialVersionUID = 808575179767517313L;
 
+  public static final String GOOGLE_DEFAULT_UNIVERSE = "googleapis.com";
+
   /**
    * A constant string name describing the authentication technology.
    *
@@ -53,6 +55,15 @@ public abstract class Credentials implements Serializable {
    * @return The type of authentication used.
    */
   public abstract String getAuthenticationType();
+
+  /**
+   * Returns the universe domain for the credential
+   *
+   * @return Return a default Google universe domain googleapis.com
+   */
+  public String getUniverseDomain() {
+    return GOOGLE_DEFAULT_UNIVERSE;
+  }
 
   /**
    * Get the current request metadata, refreshing tokens if required.
--- a/credentials/java/com/google/auth/Credentials.java
+++ b/credentials/java/com/google/auth/Credentials.java
@@ -57,9 +57,9 @@ public abstract class Credentials implements Serializable {
   public abstract String getAuthenticationType();
 
   /**
-   * Returns the universe domain for the credential
+   * Returns the universe domain for the credential.
    *
-   * @return Return a default Google universe domain googleapis.com
+   * @return a default Google universe domain googleapis.com
    */
   public String getUniverseDomain() {
     return GOOGLE_DEFAULT_UNIVERSE;
--- a/credentials/java/com/google/auth/Credentials.java
+++ b/credentials/java/com/google/auth/Credentials.java
@@ -59,7 +59,7 @@ public abstract class Credentials implements Serializable {
   /**
    * Returns the universe domain for the credential.
    *
-   * @return a default Google universe domain googleapis.com
+   * @return the default Google universe domain googleapis.com
    */
   public String getUniverseDomain() {
     return GOOGLE_DEFAULT_UNIVERSE;
--- a/credentials/java/com/google/auth/Credentials.java
+++ b/credentials/java/com/google/auth/Credentials.java
@@ -60,10 +60,10 @@ public abstract class Credentials implements Serializable {
    * Returns the universe domain for the credential.
    *
    * @return a universe domain value in the format some-domain.xyz. By default, returns the Google
-   * universe domain googleapis.com.
+   *     universe domain googleapis.com.
    * @throws IOException extending classes might have to do remote calls to determine the universe
-   * domain. The exception should implement {@link Retryable} and {@code isRetryable()} will return
-   * true if the operation may be retried.
+   *     domain. The exception should implement {@link Retryable} and {@code isRetryable()} will
+   *     return true if the operation may be retried.
    */
   public String getUniverseDomain() throws IOException {
     return GOOGLE_DEFAULT_UNIVERSE;
--- a/oauth2_http/java/com/google/auth/oauth2/ExternalAccountAuthorizedUserCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/ExternalAccountAuthorizedUserCredentials.java
@@ -245,6 +245,7 @@ public class ExternalAccountAuthorizedUserCredentials extends GoogleCredentials 
   public int hashCode() {
     return Objects.hash(
         super.hashCode(),
+        getAccessToken(),
         clientId,
         clientSecret,
         refreshToken,
@@ -281,6 +282,7 @@ public class ExternalAccountAuthorizedUserCredentials extends GoogleCredentials 
     ExternalAccountAuthorizedUserCredentials credentials =
         (ExternalAccountAuthorizedUserCredentials) obj;
     return super.equals(credentials)
+        && Objects.equals(this.getAccessToken(), credentials.getAccessToken())
         && Objects.equals(this.clientId, credentials.clientId)
         && Objects.equals(this.clientSecret, credentials.clientSecret)
         && Objects.equals(this.refreshToken, credentials.refreshToken)
--- a/oauth2_http/java/com/google/auth/oauth2/ExternalAccountAuthorizedUserCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/ExternalAccountAuthorizedUserCredentials.java
@@ -47,6 +47,7 @@ import com.google.api.client.util.Preconditions;
 import com.google.auth.http.HttpTransportFactory;
 import com.google.common.base.MoreObjects;
 import com.google.common.io.BaseEncoding;
+import com.google.errorprone.annotations.CanIgnoreReturnValue;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.ObjectInputStream;
@@ -401,6 +402,7 @@ public class ExternalAccountAuthorizedUserCredentials extends GoogleCredentials 
      * @param transportFactory the {@code HttpTransportFactory} to set
      * @return this {@code Builder} object
      */
+    @CanIgnoreReturnValue
     public Builder setHttpTransportFactory(HttpTransportFactory transportFactory) {
       this.transportFactory = transportFactory;
       return this;
@@ -413,6 +415,7 @@ public class ExternalAccountAuthorizedUserCredentials extends GoogleCredentials 
      * @param audience the audience to set
      * @return this {@code Builder} object
      */
+    @CanIgnoreReturnValue
     public Builder setAudience(String audience) {
       this.audience = audience;
       return this;
@@ -424,6 +427,7 @@ public class ExternalAccountAuthorizedUserCredentials extends GoogleCredentials 
      * @param tokenUrl the token exchange url to set
      * @return this {@code Builder} object
      */
+    @CanIgnoreReturnValue
     public Builder setTokenUrl(String tokenUrl) {
       this.tokenUrl = tokenUrl;
       return this;
@@ -435,6 +439,7 @@ public class ExternalAccountAuthorizedUserCredentials extends GoogleCredentials 
      * @param tokenInfoUrl the token info url to set
      * @return this {@code Builder} object
      */
+    @CanIgnoreReturnValue
     public Builder setTokenInfoUrl(String tokenInfoUrl) {
       this.tokenInfoUrl = tokenInfoUrl;
       return this;
@@ -446,6 +451,7 @@ public class ExternalAccountAuthorizedUserCredentials extends GoogleCredentials 
      * @param revokeUrl the revoke url to set
      * @return this {@code Builder} object
      */
+    @CanIgnoreReturnValue
     public Builder setRevokeUrl(String revokeUrl) {
       this.revokeUrl = revokeUrl;
       return this;
@@ -457,6 +463,7 @@ public class ExternalAccountAuthorizedUserCredentials extends GoogleCredentials 
      * @param refreshToken the refresh token
      * @return this {@code Builder} object
      */
+    @CanIgnoreReturnValue
     public Builder setRefreshToken(String refreshToken) {
       this.refreshToken = refreshToken;
       return this;
@@ -468,6 +475,7 @@ public class ExternalAccountAuthorizedUserCredentials extends GoogleCredentials 
      * @param clientId the client ID
      * @return this {@code Builder} object
      */
+    @CanIgnoreReturnValue
     public Builder setClientId(String clientId) {
       this.clientId = clientId;
       return this;
@@ -479,6 +487,7 @@ public class ExternalAccountAuthorizedUserCredentials extends GoogleCredentials 
      * @param clientSecret the client secret
      * @return this {@code Builder} object
      */
+    @CanIgnoreReturnValue
     public Builder setClientSecret(String clientSecret) {
       this.clientSecret = clientSecret;
       return this;
@@ -490,6 +499,7 @@ public class ExternalAccountAuthorizedUserCredentials extends GoogleCredentials 
      * @param quotaProjectId the quota and billing project id to set
      * @return this {@code Builder} object
      */
+    @CanIgnoreReturnValue
     public Builder setQuotaProjectId(String quotaProjectId) {
       super.setQuotaProjectId(quotaProjectId);
       return this;
@@ -501,6 +511,7 @@ public class ExternalAccountAuthorizedUserCredentials extends GoogleCredentials 
      * @param accessToken the access token
      * @return this {@code Builder} object
      */
+    @CanIgnoreReturnValue
     public Builder setAccessToken(AccessToken accessToken) {
       super.setAccessToken(accessToken);
       return this;
--- a/oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
@@ -726,7 +726,9 @@ public abstract class ExternalAccountCredentials extends GoogleCredentials {
     @Nullable protected Collection<String> scopes;
     @Nullable protected String workforcePoolUserProject;
     @Nullable protected ServiceAccountImpersonationOptions serviceAccountImpersonationOptions;
-    @Nullable protected String universeDomain;
+    @Nullable
+    @Deprecated
+    protected String universeDomain;
     @Nullable protected ExternalAccountMetricsHandler metricsHandler;
 
     protected Builder() {}
@@ -920,8 +922,10 @@ public abstract class ExternalAccountCredentials extends GoogleCredentials {
      * @return this {@code Builder} object
      */
     @CanIgnoreReturnValue
+    @Override
     public Builder setUniverseDomain(String universeDomain) {
       this.universeDomain = universeDomain;
+      super.setUniverseDomain(universeDomain);
       return this;
     }
 
--- a/oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
@@ -726,6 +726,7 @@ public abstract class ExternalAccountCredentials extends GoogleCredentials {
     @Nullable protected Collection<String> scopes;
     @Nullable protected String workforcePoolUserProject;
     @Nullable protected ServiceAccountImpersonationOptions serviceAccountImpersonationOptions;
+    @Nullable protected String universeDomain;
     @Nullable protected ExternalAccountMetricsHandler metricsHandler;
 
     protected Builder() {}
@@ -745,6 +746,7 @@ public abstract class ExternalAccountCredentials extends GoogleCredentials {
       this.environmentProvider = credentials.environmentProvider;
       this.workforcePoolUserProject = credentials.workforcePoolUserProject;
       this.serviceAccountImpersonationOptions = credentials.serviceAccountImpersonationOptions;
+      this.universeDomain = credentials.getUniverseDomain();
       this.metricsHandler = credentials.metricsHandler;
     }
 
--- a/oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
@@ -726,9 +726,7 @@ public abstract class ExternalAccountCredentials extends GoogleCredentials {
     @Nullable protected Collection<String> scopes;
     @Nullable protected String workforcePoolUserProject;
     @Nullable protected ServiceAccountImpersonationOptions serviceAccountImpersonationOptions;
-    @Nullable
-    @Deprecated
-    protected String universeDomain;
+    @Nullable @Deprecated protected String universeDomain;
     @Nullable protected ExternalAccountMetricsHandler metricsHandler;
 
     protected Builder() {}
--- a/oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
@@ -36,6 +36,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import com.google.api.client.http.HttpHeaders;
 import com.google.api.client.json.GenericJson;
 import com.google.api.client.json.JsonObjectParser;
+import com.google.auth.Credentials;
 import com.google.auth.RequestMetadataCallback;
 import com.google.auth.http.HttpTransportFactory;
 import com.google.common.base.MoreObjects;
@@ -747,8 +748,14 @@ public abstract class ExternalAccountCredentials extends GoogleCredentials {
       this.environmentProvider = credentials.environmentProvider;
       this.workforcePoolUserProject = credentials.workforcePoolUserProject;
       this.serviceAccountImpersonationOptions = credentials.serviceAccountImpersonationOptions;
-      this.universeDomain = credentials.getUniverseDomain();
       this.metricsHandler = credentials.metricsHandler;
+      try {
+        universeDomain = credentials.getUniverseDomain();
+      }
+      catch (IOException ex) {
+        throw new RuntimeException("Unexpected exception while getting universe domain", ex);
+      }
+      this.universeDomain = universeDomain;
     }
 
     /**
--- a/oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
@@ -394,7 +394,7 @@ public abstract class ExternalAccountCredentials extends GoogleCredentials {
    */
   @SuppressWarnings("unchecked")
   static ExternalAccountCredentials fromJson(
-      Map<String, Object> json, HttpTransportFactory transportFactory) throws IOException {
+      Map<String, Object> json, HttpTransportFactory transportFactory) {
     checkNotNull(json);
     checkNotNull(transportFactory);
 
@@ -411,9 +411,9 @@ public abstract class ExternalAccountCredentials extends GoogleCredentials {
     String clientSecret = (String) json.get("client_secret");
     String quotaProjectId = (String) json.get("quota_project_id");
     String userProject = (String) json.get("workforce_pool_user_project");
+    String universeDomain = (String) json.get("universe_domain");
     Map<String, Object> impersonationOptionsMap =
         (Map<String, Object>) json.get("service_account_impersonation");
-    String universeDomain = (String) json.get("universe_domain");
 
     if (impersonationOptionsMap == null) {
       impersonationOptionsMap = new HashMap<String, Object>();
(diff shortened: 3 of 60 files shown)


oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -185,14 +188,55 @@ public static GoogleCredentials fromStream(
...
+   *
+   * @param json a map from the JSON representing the credentials.
+   * @return the credentials defined by the JSON.
+   * @throws IOException if the credential cannot be created from the JSON.
+   */
+  @VisibleForTesting
+  static GoogleCredentials fromJson(Map<String, Object> json) throws IOException {
+    Builder credentialsBuilder = GoogleCredentials.newBuilder();
+
+    // parse fields that are common for all the types of GoogleCredentials
```
lsirac, line 220:
```suggestion
    // Parse fields that are common for all the types of GoogleCredentials.
```

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -185,14 +188,55 @@ public static GoogleCredentials fromStream(
...
+    }
+
+    return credentialsBuilder.build();
+  }
+
+  /**
+   * Creates a credential with the provided universe domain.
+   *
+   * @param universeDomain the universe domain to set on the credential. Empty or null value will
+   *     result in the default universe domain 'googleapis.com'
```
lsirac, line 235:
```suggestion
   *     result in the default universe domain 'googleapis.com'.
```

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -185,14 +188,55 @@ public static GoogleCredentials fromStream(
...
+   * @return credential with the provided universe domain
```
lsirac, line 236:
```suggestion
   * @return credential with the provided universe domain.
```

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -205,6 +249,20 @@ public GoogleCredentials createWithQuotaProject(String quotaProject) {
     return this.toBuilder().setQuotaProjectId(quotaProject).build();
   }
 
+  /**
+   * Returns the universe domain for the credential
```
lsirac, line 253:
```suggestion
   * Returns the universe domain for the credential.
```

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -205,6 +249,20 @@ public GoogleCredentials createWithQuotaProject(String quotaProject) {
...
+   *
+   * @return An explicit universe domain if it was explicitly provided, {@code
+   *     GoogleAuthUtils.GOOGLE_DEFAULT_UNIVERSE} otherwise.
+   */
+  public String getUniverseDomain() {
+    if (this.universeDomain == null || universeDomain.trim().isEmpty()) {
```
lsirac, line 259: Why not just set universeDomain to the default value if none is provided, and then we can avoid this altogether and just return it directly?

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -236,9 +294,19 @@ protected GoogleCredentials() {
     this(new Builder());
   }
 
+  /**
+   * Constructor with explicit access token and quotaProjectId
```
lsirac, line 298:
```suggestion
   * Constructor with explicit access token and quotaProjectId.
```

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -185,14 +188,55 @@ public static GoogleCredentials fromStream(
...
+   */
+  @VisibleForTesting
+  static GoogleCredentials fromJson(Map<String, Object> json) throws IOException {
+    Builder credentialsBuilder = GoogleCredentials.newBuilder();
+
+    // parse fields that are common for all the types of GoogleCredentials
+    String universeDomain = (String) json.get("universe_domain");
+    if (universeDomain != null
+        && universeDomain.trim().isEmpty()
+        && !universeDomain.equalsIgnoreCase(GoogleAuthUtils.GOOGLE_DEFAULT_UNIVERSE)) {
```
lsirac, line 224: What's the point of this check?

lsirac (commented):
I didn't comment on all of them, but lets be consistent with our java docs (capitalization, periods, etc)

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
TimurSadykov: this was removed

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
TimurSadykov: Periods are too often omitted for single-sentence @params and @returns in this library. I don't thinks its too bad as the don't look like normal sentences when starts with keywords. Let's not enforce it until we clean up significant part of the existing code with a separate PR if we decide to do so. 

For the rest of the docs - Lets have correct sentences.

TimurSadykov:
> I didn't comment on all of them, but lets be consistent with our java docs (capitalization, periods, etc)

Agree for the javadocs body, except single sentence @params and other keywords.. its is mostly lowcaps, no period now.

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
TimurSadykov: outdated, it was an unfinished draft

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
TimurSadykov: To implement the alternative you will have to implement checks and test them in all creds now and in the future. Plus take into account edge cases like deserealization of the old credential versions...

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
```diff
@@ -415,10 +411,11 @@ static ExternalAccountCredentials fromJson(
     String clientSecret = (String) json.get("client_secret");
     String quotaProjectId = (String) json.get("quota_project_id");
     String userProject = (String) json.get("workforce_pool_user_project");
-    String universeDomain = (String) json.get("universe_domain");
     Map<String, Object> impersonationOptionsMap =
         (Map<String, Object>) json.get("service_account_impersonation");
 
+    GoogleCredentials baseCredential = GoogleCredentials.fromJson(json);
```
aeitzman, line 417: Is there no way to make this a super() call or something? Seems awkward to build a seperate credential here just to get the universe domain.

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -251,19 +321,64 @@ public GoogleCredentials(AccessToken accessToken) {
...
+        .add("universeDomain", getUniverseDomain());
+  }
+
+  @Override
+  public String toString() {
+    return toStringHelper().toString();
+  }
+
+  @Override
+  public boolean equals(Object obj) {
```
aeitzman, line 370: nit: we should either include access token here too or not add it the authorized user credentials I think.

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -463,26 +465,15 @@ public static ServiceAccountCredentials fromStream(InputStream credentialsStream
...
-    String fileType = (String) fileContents.get("type");
-    if (fileType == null) {
-      throw new IOException("Error reading credentials from stream, 'type' field not specified.");
-    }
-    if (SERVICE_ACCOUNT_FILE_TYPE.equals(fileType)) {
-      return fromJson(fileContents, transportFactory);
+    ServiceAccountCredentials credential =
+        (ServiceAccountCredentials)
+            GoogleCredentials.fromStream(credentialsStream, transportFactory);
+    if (credential == null) {
```
aeitzman, line 471: nit: could this ever actually be hit? Wouldn't the fromStream either return a credential (which will throw an error on the cast if its not serviceAccountCredentials) or throw an error?

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -491,6 +482,12 @@ public boolean createScopedRequired() {
     return scopes.isEmpty() && defaultScopes.isEmpty();
   }
 
+  /** Returns true if credential is configured domain wide delegation */
+  @VisibleForTesting
+  boolean isConfiguredForDomainWideDelegation() {
```
aeitzman, line 487: nit: could this just be named hasServiceAccountUser or something like that instead? Makes sense to me to have individual functions that just do what they say, and then if we end up with more checks later for Domain wide delegation we could have a method that combines the smaller methods.

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -1098,6 +1127,11 @@ public Builder setDefaultRetriesEnabled(boolean defaultRetriesEnabled) {
       return this;
     }
 
+    public Builder setUniverseDomain(String universeDomain) {
+      this.universeDomain = universeDomain;
```
aeitzman, line 1131: super.setUniverseDomain(universeDomain)?

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -912,9 +909,12 @@ JwtCredentials createSelfSignedJwtCredentials(final URI uri) {
   @Override
   public void getRequestMetadata(
       final URI uri, Executor executor, final RequestMetadataCallback callback) {
-    if (useJwtAccessWithScope) {
-      // This will call getRequestMetadata(URI uri), which handles self signed JWT logic.
-      // Self signed JWT doesn't use network, so here we do a blocking call to improve
+    // For default universe Self-signed JWT could be explicitly disabled with
+    // {@code ServiceAccountCredentials.useJwtAccessWithScope} flag.
+    // If universe is non-default, it only supports self-signed JWT, and it is always allowed.
+    if (this.useJwtAccessWithScope || !isDefaultUniverseDomain()) {
```
aeitzman, line 915: Can you add documentation on the builder setters for the jwtAccessWithScope flag that it gets ignored if a non-default universe is used?

oauth2_http/java/com/google/auth/oauth2/ExternalAccountAuthorizedUserCredentials.java
```diff
@@ -245,6 +245,7 @@ public static Builder newBuilder() {
   public int hashCode() {
     return Objects.hash(
         super.hashCode(),
+        getAccessToken(),
```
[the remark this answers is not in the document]
TimurSadykov: I thought about that, but I think no... We inevitably modify the hashCode with every new field. It is related to TPC in a way universe_domain is created in base class and it triggers this change as well

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
TimurSadykov: We don't want to do that. I did include it to Credentials that historically has it. Its just two. But in general they don't want to include access token

oauth2_http/java/com/google/auth/oauth2/ExternalAccountAuthorizedUserCredentials.java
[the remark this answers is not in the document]
TimurSadykov: To be clear, here i'm keeping the original behavior, due to a change in super implementation - I have to call getAccessToken directly. The sequence of fields in the formula changes, but the behavior stays the same.

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -932,17 +932,45 @@ public Map<String, List<String>> getRequestMetadata(URI uri) throws IOException
...
+  }
+
+  /** Provide the access JWT for scopes if provided, for uri as aud otherwise */
+  @VisibleForTesting
+  private Map<String, List<String>> getRequestMetadataWithSelfSignedJwt(URI uri)
+      throws IOException {
+    // If scopes are provided and self-signed JWT can be used, use self-signed JWT with scopes.
+    // Otherwise, use self-signed JWT with uri as the audience.
     JwtCredentials jwtCredentials;
-    if (!createScopedRequired() && useJwtAccessWithScope) {
```
arithmetic1728, line 945: is removing `useJwtAccessWithScope` backward compatible?

TimurSadykov: It's a diff making it look confusing ) 

Those lines were moved into private method. You can see that for GDU there is same check (line 945) and for non-gdu it is not needed. 

So no breaking change for GDU

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
TimurSadykov: Hopefully it makes sense )

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
TimurSadykov: Done )

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
TimurSadykov: It will be null if the returning type would be different from ServiceAccountCredentials

TimurSadykov: I will add a test for that if there is none yet

credentials/java/com/google/auth/Credentials.java
```diff
@@ -43,6 +43,8 @@ public abstract class Credentials implements Serializable {
 
   private static final long serialVersionUID = 808575179767517313L;
 
+  public static final String GOOGLE_DEFAULT_UNIVERSE = "googleapis.com";
```
lsirac, line 46: should this be private?

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -251,19 +321,64 @@ public GoogleCredentials(AccessToken accessToken) {
     this(accessToken, null);
   }
 
+  /**
+   * Constructor with using builder All the fields comes explicitly from builder.
```
lsirac, line 325: I think we can word this better (also "All" is incorrectly capitalized).

How about something like:

```
This constructor relies on a {@link GoogleCredential.Builder} to 
provide all the necessary field values for initialization
```

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -251,19 +321,64 @@ public GoogleCredentials(AccessToken accessToken) {
...
   protected GoogleCredentials(Builder builder) {
-    this(builder.getAccessToken(), builder.getQuotaProjectId());
+    super(builder.getAccessToken());
+    this.quotaProjectId = builder.getQuotaProjectId();
+    this.universeDomain = builder.getUniverseDomain();
   }
 
   /**
-   * Constructor with explicit access token and refresh times
+   * Constructor with explicit access token and refresh times.
```
lsirac, line 336:
```suggestion
   * Constructor with explicit access token and refresh margins.
```

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -251,19 +321,64 @@ public GoogleCredentials(AccessToken accessToken) {
...
+   *
+   * <p>Deprecated, please use the builder constructor whenever possible.
```
lsirac, line 338: Let's link it here (and same everywhere else where this applies)

credentials/java/com/google/auth/Credentials.java
TimurSadykov: It's a different package than everything else, so we want to have it public to compare if current value is different etc

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -186,18 +193,46 @@ public static GoogleCredentials fromStream(
...
+    if (universeDomain != null && !universeDomain.trim().isEmpty()) {
+      credentialsBuilder.setUniverseDomain(universeDomain);
+    }
+
+    return credentialsBuilder.build();
   }
 
   /**
-   * Creates a credential with the provided quota project.
+   * Creates a credential with the provided quota project. Expected format is: some-domain.xyz
```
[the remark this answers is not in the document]
TimurSadykov: oh funny ) it somehow left from a previous iteration where I had a createWithUniverseDomain helper.

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
TimurSadykov: Thanks for catching, seems like stopped in the middle of editing ) 

Suggestion sounds good

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
TimurSadykov: Agree, done

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -1086,6 +1116,10 @@ public Builder setLifetime(int lifetime) {
       return this;
     }
 
+    /**
+     * Sets the useJwtAccessWithScope flag. This flag will be ignored if universeDomain field is
+     * different from {@link Credentials.GOOGLE_DEFAULT_UNIVERSE}.
+     */
```
lqiu96, line 1122: I believe we have some logic regarding with using this flag in GAX: https://github.com/googleapis/sdk-platform-java/blob/c0e03cdce18e55f46e25f5aa71608416516cf4f9/gax-java/gax/src/main/java/com/google/api/gax/core/GoogleCredentialsProvider.java#L98-L102. qq, does this return a valid ServiceAccountCredentials that can be used or will it throw an exception somewhere later when the user tries to use the Credentials?

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -186,14 +193,42 @@ public static GoogleCredentials fromStream(
...
+   */
+  @VisibleForTesting
+  static GoogleCredentials fromJson(Map<String, Object> json) throws IOException {
+    Builder credentialsBuilder = GoogleCredentials.newBuilder();
+
+    // Parse fields that are common for all the types of GoogleCredentials.
+    String universeDomain = (String) json.get("universe_domain");
+    if (universeDomain != null && !universeDomain.trim().isEmpty()) {
+      credentialsBuilder.setUniverseDomain(universeDomain);
+    }
```
lqiu96, line 229: qq, does this need to return GDU if the map doesn't contain `universe_domain`?

TimurSadykov: Builder handles that, null and empty check is redundant here as well

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
TimurSadykov: It returns a valid ServiceAccountCredentials, its just this flag is ignored internally if universe is not default.

What the GAX code does is setting the value to true for GAPICs. Effectively enabling SelfSignedJwt. It will work as before in GDU. In other universes SelfSignedJWT is the only option no need to use the flag. But setting it will not cause any exception )

credentials/java/com/google/auth/Credentials.java
```diff
@@ -54,6 +56,19 @@ public abstract class Credentials implements Serializable {
    */
   public abstract String getAuthenticationType();
 
+  /**
+   * Returns the universe domain for the credential.
+   *
+   * @return a universe domain value in the format some-domain.xyz. By default, returns the Google
+   *     universe domain googleapis.com.
+   * @throws IOException extending classes might have to do remote calls to determine the universe
+   *     domain. The exception should implement {@link Retryable} and {@code isRetryable()} will
```
lqiu96, line 65:
```suggestion
   *     domain. The exception implements {@link Retryable} and {@code isRetryable()} will
```

nit: Correct me if I'm wrong but for `getUniverseDomain()` the IOException will always be a Retryable interface, right? Credentials without a remote call shouldn't throw an IOException.

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
lqiu96: Understood, thanks!

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
westarle: Using GoogleCredentials as a concrete holder of JSON data is surprising.  It would be better to provide something like GoogleCredentialJson as a simple data class to wrap the contents of a data file, but it ought to fully specify the contents of the file parsed.

Lacking a full specification of the format shared by all credentials, it is better to inline reading of all values in each concrete credential.

Thanks!

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
```diff
@@ -928,8 +926,10 @@ public Builder setServiceAccountImpersonationOptions(Map<String, Object> options
      * @return this {@code Builder} object
      */
     @CanIgnoreReturnValue
+    @Override
     public Builder setUniverseDomain(String universeDomain) {
       this.universeDomain = universeDomain;
+      super.setUniverseDomain(universeDomain);
```
westarle, line 932: nit: call super first.

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -186,14 +196,40 @@ public static GoogleCredentials fromStream(
...
-    if ("impersonated_service_account".equals(fileType)) {
+    if (ImpersonatedCredentials.IMPERSONATED_CREDENTIALS_FILE_TYPE.equals(fileType)) {
       return ImpersonatedCredentials.fromJson(fileContents, transportFactory);
     }
     throw new IOException(
         String.format(
             "Error reading credentials from stream, 'type' value '%s' not recognized."
-                + " Expecting '%s' or '%s'.",
-            fileType, USER_FILE_TYPE, SERVICE_ACCOUNT_FILE_TYPE));
+                + " Valid values are '%s', '%s', '%s', '%s', '%s', 's'.",
```
westarle, line 205: Last param missing %s;

I really like this idea -- please make sure it is tested!

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -206,6 +242,27 @@ public GoogleCredentials createWithQuotaProject(String quotaProject) {
...
 
+  /**
+   * Gets the universe domain for the credential.
+   *
+   * @return An explicit universe domain if it was explicitly provided, invokes the super
+   *     implementation otherwise
+   */
+  @Override
+  public String getUniverseDomain() throws IOException {
+    return this.universeDomain;
```
westarle, line 253: nit: this. not necessary.

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -71,7 +77,10 @@ public class GoogleCredentials extends OAuth2Credentials implements QuotaProject
    * @return the credentials instance
    */
   public static GoogleCredentials create(AccessToken accessToken) {
-    return GoogleCredentials.newBuilder().setAccessToken(accessToken).build();
+    return GoogleCredentials.newBuilder()
```
westarle, line 80: deprecate this constructor in favor of the builder?

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -251,19 +319,66 @@ public GoogleCredentials(AccessToken accessToken) {
...
+  /**
+   * Constructor that relies on a {@link GoogleCredential.Builder} to provide all the necessary
+   * field values for initialization.
+   *
+   * @param builder an instance of a builder
+   */
   protected GoogleCredentials(Builder builder) {
-    this(builder.getAccessToken(), builder.getQuotaProjectId());
+    super(builder.getAccessToken());
+    this.quotaProjectId = builder.getQuotaProjectId();
```
westarle, line 330: nit: `this.` is not needed here and elsewhere.

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -251,19 +319,66 @@ public GoogleCredentials(AccessToken accessToken) {
...
+    return toStringHelper().toString();
+  }
+
+  @Override
+  public boolean equals(Object obj) {
+    if (!(obj instanceof GoogleCredentials)) {
+      return false;
+    }
+    GoogleCredentials other = (GoogleCredentials) obj;
+    return Objects.equals(this.getQuotaProjectId(), other.getQuotaProjectId())
```
westarle, line 375: better to use quotaProjectId directly.

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -366,10 +489,23 @@ public Builder setQuotaProjectId(String quotaProjectId) {
...
+      this.universeDomain = universeDomain;
+      return this;
+    }
+
     public String getQuotaProjectId() {
       return this.quotaProjectId;
     }
 
+    public String getUniverseDomain() {
+      if (this.universeDomain == null || this.universeDomain.trim().isEmpty()) {
```
westarle, line 502: nit: I feel like it's more correct to do the defaulting when calling "build"; this way I can tell if a builder has a universe domain set or not.

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -1098,6 +1132,11 @@ public Builder setDefaultRetriesEnabled(boolean defaultRetriesEnabled) {
       return this;
     }
 
+    public Builder setUniverseDomain(String universeDomain) {
```
westarle, line 1135: Shouldn't this be provided by the superclass setUniverseDomain already?

westarle (commented):
feel free to ignore nits; noticing that use of explicit "this." is pretty common and comes down to personal preference :)

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
TimurSadykov: Agree, thanks!

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
TimurSadykov: The base will return GoogleCredentials and user will have to cast it back to ServiceAccount

credentials/java/com/google/auth/Credentials.java
```diff
@@ -54,6 +56,19 @@ public abstract class Credentials implements Serializable {
    */
   public abstract String getAuthenticationType();
 
+  /**
+   * Returns the universe domain for the credential.
```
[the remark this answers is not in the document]
TimurSadykov: Added, thanks!

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
TimurSadykov: Done

credentials/java/com/google/auth/Credentials.java
TimurSadykov: For credentials we control - yes. Otherwise - we cannot guarantee. We can change it for "must implement".

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
TimurSadykov: Oh nice catch :)

lqiu96 (approved):

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
```diff
@@ -398,7 +394,7 @@ public static ExternalAccountCredentials fromStream(
    */
   @SuppressWarnings("unchecked")
   static ExternalAccountCredentials fromJson(
-      Map<String, Object> json, HttpTransportFactory transportFactory) {
+      Map<String, Object> json, HttpTransportFactory transportFactory) throws IOException {
```
westarle, line 397: Is this throw still needed?

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
```diff
@@ -415,9 +411,9 @@ static ExternalAccountCredentials fromJson(
     String clientSecret = (String) json.get("client_secret");
     String quotaProjectId = (String) json.get("quota_project_id");
     String userProject = (String) json.get("workforce_pool_user_project");
-    String universeDomain = (String) json.get("universe_domain");
     Map<String, Object> impersonationOptionsMap =
         (Map<String, Object>) json.get("service_account_impersonation");
+    String universeDomain = (String) json.get("universe_domain");
```
westarle, line 416: nit: move this back under userProject to avoid change noise?

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
```diff
@@ -754,8 +745,12 @@ protected Builder(ExternalAccountCredentials credentials) {
       this.environmentProvider = credentials.environmentProvider;
       this.workforcePoolUserProject = credentials.workforcePoolUserProject;
       this.serviceAccountImpersonationOptions = credentials.serviceAccountImpersonationOptions;
-      this.universeDomain = credentials.universeDomain;
       this.metricsHandler = credentials.metricsHandler;
+      try {
```
westarle, line 749: even if we don't remove the universeDomain field, can we have the builder copy the field without the try/catch/RTE -- what breaks?  I don't expect we test the RTE case.

westarle (approved):
Just left a couple more nits to address, thank you.

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
TimurSadykov: actually you are right... This could not be effectively used yet anyways, it was shipped as public by mistake. The rest of the universeDomain stuff in the credential was not public yet. It would be a zombie field in the builder until we remove it.  I will remove the try-catch and document that the field is not used.

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -315,6 +295,7 @@ protected GoogleCredentials(AccessToken accessToken, String quotaProjectId) {
    *
    * @param accessToken initial or temporary access token
    */
+  @Deprecated
```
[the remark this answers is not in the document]
TimurSadykov: I think we want to keep this one and add abother overload with universe_domain. It has existing usage and per requirements, we don't want to push code changes with this feature. Let's sync offline if any questions

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
TimurSadykov: yep, thanks!

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
TimurSadykov: Look like not anymore. IDE does not gray it out though...

credentials/java/com/google/auth/Credentials.java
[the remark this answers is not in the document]
lsirac: Not sure I understand why we mention token refreshes or that is it blocking? Seems reasonable that all calls are blocking unless the API specifies otherwise.

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -237,33 +274,97 @@ protected GoogleCredentials() {
...
+   * field values for initialization.
+   *
+   * @param builder an instance of a builder
+   */
   protected GoogleCredentials(Builder builder) {
-    this(builder.getAccessToken(), builder.getQuotaProjectId());
+    super(builder.getAccessToken());
+    this.quotaProjectId = builder.getQuotaProjectId();
+
+    if (builder.universeDomain == null || builder.universeDomain.trim().isEmpty()) {
```
lsirac, line 313: Can be simplified:

```
this.universeDomain = builder.getUniverseDomain();
if (this.universeDomain == null || this.universeDomain.trim().isEmpty()) {
  this.universeDomain = Credentials.GOOGLE_DEFAULT_UNIVERSE;
}

lsirac (approved):

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
TimurSadykov: ah nope... finals can be assigned once and only once ) could have worked in c# though ))

PR #1284, opened 2023-09-13
fix: refactor compute and cloudshell credentials to pass quota project to base class

sai-sunder-s:
Currently quota project that is set in the builder is dropped when build() is called for cloudshell and gce creds because the constructors are not rolling up the builders upto the base class.

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/CloudShellCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/CloudShellCredentials.java
@@ -140,6 +140,12 @@ public class CloudShellCredentials extends GoogleCredentials {
       return this;
     }
 
+    @CanIgnoreReturnValue
+    public Builder setQuotaProjectId(String quotaProjectId) {
+      super.quotaProjectId = quotaProjectId;
+      return this;
+    }
+
     public int getAuthPort() {
       return authPort;
     }
--- a/oauth2_http/java/com/google/auth/oauth2/CloudShellCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/CloudShellCredentials.java
@@ -67,8 +67,9 @@ public class CloudShellCredentials extends GoogleCredentials {
     return CloudShellCredentials.newBuilder().setAuthPort(authPort).build();
   }
 
-  private CloudShellCredentials(int authPort) {
-    this.authPort = authPort;
+  private CloudShellCredentials(Builder builder) {
+    super(builder);
+    this.authPort = builder.getAuthPort();
   }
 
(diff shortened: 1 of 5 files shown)


oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
```diff
@@ -637,8 +637,8 @@ public void run() {
   public static class Builder {
 
     private AccessToken accessToken;
-    private Duration refreshMargin = DEFAULT_REFRESH_MARGIN;
-    private Duration expirationMargin = DEFAULT_EXPIRATION_MARGIN;
+    protected Duration refreshMargin = DEFAULT_REFRESH_MARGIN;
```
TimurSadykov, line 640: why this?

TimurSadykov (changes requested):
Overall looks good, except margin visibility

oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
sai-sunder-s: removed

oauth2_http/java/com/google/auth/oauth2/CloudShellCredentials.java
```diff
@@ -145,7 +146,7 @@ public int getAuthPort() {
 
     @Override
     public CloudShellCredentials build() {
-      return new CloudShellCredentials(authPort);
+      return new CloudShellCredentials(this);
```
TimurSadykov, line 149: Can you please add a test  for this, similar to toBuilder_build in ComputeCredentialsTests?

oauth2_http/java/com/google/auth/oauth2/CloudShellCredentials.java
```diff
@@ -67,8 +67,9 @@ public static CloudShellCredentials create(int authPort) {
     return CloudShellCredentials.newBuilder().setAuthPort(authPort).build();
   }
 
-  private CloudShellCredentials(int authPort) {
-    this.authPort = authPort;
+  private CloudShellCredentials(Builder builder) {
+    super(builder);
+    this.authPort = builder.getAuthPort();
```
TimurSadykov, line 72: Could you please test this as well? While it is partially covered in CredentialsProvider - we want to have coverage here as well. CredentialsProvider might someday go away or instantiate credentials differently.

TimurSadykov (changes requested):
please add more tests

TimurSadykov (approved):
LGTM, thanks!

PR #1300, opened 2023-10-06
fix: serialVersionUID fix for Serializable Classes

rng70-or:
##### Motivation:

In file: OAuth2Credentials.java and ExternalAccountMetricsHandler.java, two classes are serializable but they do not contain any serialVersionUID field. The compiler generates one by default in such scenarios, but the generated id is dependent on compiler implementation and may cause unwanted problems during deserialization.

##### The Role of serialVersionUID:

The primary role of serialVersionUID is to provide version control during deserialization. When deserialize an object, the JVM checks whether the serialVersionUID of the serialized data matches the serialVersionUID of the class in the current classpath. If they match, the deserialization proceeds without issues. However, if they do not match, programmers encounter InvalidClassException.

As, serialVersionUID servers the purpose of version control of class during serialization-deserialization, without a serialVersionUID, we risk breaking backward compatibility when making changes to classes, which can lead to unexpected issues and errors during deserialization.

##### Sponsorship and Support:

This work is done by the security researchers from OpenRefactory and is supported by the [Open Source Security Foundation (OpenSSF)](https://openssf.org/): [Project Alpha-Omega](https://alpha-omega.dev/). Alpha-Omega is a project partnering with open source software project maintainers to systematically find new, as-yet-undiscovered vulnerabilities in open source code - and get them fixed – to improve global software supply chain security.

The bug is found by running the Intelligent Code Repair (iCR) tool by OpenRefactory and then manually triaging the results.


lqiu96:
To me, this seems like a reasonable ask. I will raise an issue to track the serialVersionUID value across the auth library in the monorepo (https://github.com/googleapis/google-cloud-java) and any dicussions can occur on that issue tracker. I will try look to review this PR in a bit (I'll try to get around to this, but I may have other priorities).

Closing this so we can track it at the proper spot over at the monorepo. https://github.com/googleapis/google-cloud-java/issues/12720

PR #1324, opened 2023-11-05
fix: Remove -Xlint:unchecked, suppress all existing violations, add @CanIgnoreReturnValue

### LINKED ISSUE #1323: unchecked casts are bad practice, but allowed globally in pom.xml
The pom.xml adds `-Xlint:unchecked` globally. We should remove this, and deal with existing violations individually.

sinelaw:
Fixes #1323  ☕️

This will prevent further violations from being added without explicit suppressions.

A future step would be to review all existing violations and fix instead of suppressing them.

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java
@@ -49,6 +49,7 @@ import com.google.auth.http.HttpTransportFactory;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.MoreObjects;
 import com.google.common.collect.ImmutableMap;
+import com.google.errorprone.annotations.CanIgnoreReturnValue;
 import java.io.IOException;
 import java.io.ObjectInputStream;
 import java.text.DateFormat;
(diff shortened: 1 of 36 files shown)


oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java
```diff
@@ -418,7 +419,7 @@ public boolean createScopedRequired() {
   @Override
   public GoogleCredentials createScoped(Collection<String> scopes) {
     return toBuilder()
-        .setScopes(new ArrayList(scopes))
+        .setScopes(new ArrayList<>(scopes))
```
TimurSadykov, line 422: why this?

sinelaw: To avoid using raw types (ArrayList vs. generic ArrayList<>)

TimurSadykov:
@sinelaw thanks for the contribution, it makes sense, please see a single nit comment

sinelaw:
Added another commit - PTAL

TimurSadykov (approved):
LGTM

PR #1334, opened 2023-11-21
fix: handle error-prone warnings

sinelaw:
[ErrorProne](https://errorprone.info/) flags several issues in the codebase, this pull request fixes some of them:

- Missing `@Override` annotations where needed
- Using of deprecated or incompatibel APIs for Base64 encoding

There's also a warning by maven due to conflicting / duplicate dependencies in pom.xml, this is also fixed.

### DIFF
--- a/appengine/java/com/google/auth/appengine/AppEngineCredentials.java
+++ b/appengine/java/com/google/auth/appengine/AppEngineCredentials.java
@@ -139,6 +139,7 @@ public class AppEngineCredentials extends GoogleCredentials implements ServiceAc
     return new Builder();
   }
 
+  @Override
   public Builder toBuilder() {
     return new Builder(this);
   }
@@ -175,6 +176,7 @@ public class AppEngineCredentials extends GoogleCredentials implements ServiceAc
       return appIdentityService;
     }
 
+    @Override
     public AppEngineCredentials build() {
       return new AppEngineCredentials(scopes, appIdentityService);
     }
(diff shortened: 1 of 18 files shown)


ejona86:
The Base64 thing is about Android compatibility.

sinelaw:
> The Base64 thing is about Android compatibility.

Indeed, but it also triggers error prone on https://errorprone.info/bugpattern/AndroidJdkLibsChecker

ejona86:
> > The Base64 thing is about Android compatibility.
> 
> Indeed, but it also triggers error prone on https://errorprone.info/bugpattern/AndroidJdkLibsChecker

Yet the proper solution in many projects is to turn off the check. API desugaring supports this on Android. For a library, it's mostly a question if you require your users to use API desugaring.

You can avoid the API or use it. The point is that it is a decision that is not made "because Error Prone complained." Error Prone is well beyond the "only tell you about highly likely bugs" that it set out to be. You have to tweak and tune it to your usages and be willing to add ignores (as has been repeated to me multiple times when I complained that a suggestion was encouraging security vulnerabilities or was wrong). And some of this is auto-configured inside Google; sometimes an warning means you have a configuration mistake, not a code mistake.

sinelaw:
Removed the Base64 commit from this PR, so we can discuss it separately.

TimurSadykov (approved):
LGTM, thanks!

lsirac (approved):

PR #1336, opened 2023-11-27
feat: Adds support for user defined subject token suppliers in AWSCredentials and IdentityPoolCredentials

aeitzman:
see go/programmable-auth-design.
Also adds quality of life improvements for building credentials directly via the builder methods, and a subject token type enum.


oauth2_http/java/com/google/auth/oauth2/AwsCredentials.java
```diff
@@ -132,6 +156,9 @@ public GoogleCredentials createScoped(Collection<String> newScopes) {
 
   @Override
   String getCredentialSourceType() {
+    if (this.awsSecurityCredentialsSupplier != null) {
+      return "programmatic";
```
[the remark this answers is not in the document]
aeitzman: We can already tell that its AWS vs identity pool from the token type requested (AWSV4 vs JWT/Id_token/SAML2. will change to constants though

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
```diff
@@ -280,8 +294,6 @@ protected ExternalAccountCredentials(ExternalAccountCredentials.Builder builder)
         builder.metricsHandler == null
             ? new ExternalAccountMetricsHandler(this)
             : builder.metricsHandler;
-
-    this.impersonatedCredentials = buildImpersonatedCredentials();
```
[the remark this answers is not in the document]
aeitzman: Sure, probably makes sense to just to keep it the same. I didn't make that change initially since the issue we have with the impersonated credentials is only possible in the builder constructor

lsirac:
We should also update the README here

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
```diff
@@ -53,6 +54,9 @@
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Map;
+import java.util.function.Supplier;
+import javax.annotation.Nullable;
 
 /**
  * Url-sourced and file-sourced external account credentials.
```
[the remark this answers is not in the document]
aeitzman: added supplier info in javadoc for both identity pool and aws

aeitzman:
> We should also update the README here + add integration tests

added integration tests, will add README update in a separate PR

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
```diff
@@ -492,10 +503,10 @@ private static boolean isAwsCredential(Map<String, Object> credentialSource) {
...
-    // Internal override takes priority.
-    if (impersonatedCredentialsOverride != null) {
-      return impersonatedCredentialsOverride.refreshAccessToken();
-    } else if (impersonatedCredentials != null) {
+    if (this.isImpersonationEnabled() && this.impersonatedCredentials == null) {
+      this.impersonatedCredentials = this.buildImpersonatedCredentials();
+    }
+    if (impersonatedCredentials != null) {
       return impersonatedCredentials.refreshAccessToken();
     }
```
TimurSadykov, line 511: This can look easier to understand and simpler if combined. Or does it mean that impersonatedCredentials can be provided with isImpersonationEnabled == false?

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
```diff
@@ -492,10 +503,10 @@ private static boolean isAwsCredential(Map<String, Object> credentialSource) {
   protected AccessToken exchangeExternalCredentialForAccessToken(
       StsTokenExchangeRequest stsTokenExchangeRequest) throws IOException {
     // Handle service account impersonation if necessary.
-    // Internal override takes priority.
-    if (impersonatedCredentialsOverride != null) {
-      return impersonatedCredentialsOverride.refreshAccessToken();
-    } else if (impersonatedCredentials != null) {
+    if (this.isImpersonationEnabled() && this.impersonatedCredentials == null) {
```
TimurSadykov, line 506: Why we can we have a impersonationUrl but impersonated credentials not ready? Can we build them whenever url is set?

TimurSadykov (commented):
Couple first comments, still looking

TimurSadykov:
> _ No description provided. _

@aeitzman please add description: reference to an internal doc. And a brief description, like your QoL changes. I know the context, but if other team members or Cloud Java folks will look - it is unclear. Also in the future it would be hard to uncover context.

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
aeitzman: Not sure exactly what you are implying with combined? you mean adding the this.impersonatedCredentials check  into the isImpersonationEnabled method? I can do that, would just rename the method shouldBuildImpersatedCredentials I guess. The idea of the check here is to only call the build method if we haven't already instantiated the impersonated credentials if that wasn't clear.

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
```diff
@@ -104,11 +123,7 @@ abstract static class CredentialSource implements java.io.Serializable {
 
   protected transient HttpTransportFactory transportFactory;
 
-  @Nullable protected final ImpersonatedCredentials impersonatedCredentials;
-
-  // Internal override for impersonated credentials. This is done to keep
-  // impersonatedCredentials final.
-  @Nullable private ImpersonatedCredentials impersonatedCredentialsOverride;
+  @Nullable protected ImpersonatedCredentials impersonatedCredentials;
```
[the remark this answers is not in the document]
aeitzman: The issue with having impersonated credentials final is that the ImpersonatedCredentials object is built by passing the current credential as the SourceCredential. If we have it final, the Impersonated Credential needs to get created in the parent constructor before the child constructor runs, which means the credential object that gets passed as the source credential is not fully instantiated. As a series of steps, the exact problem is:

1. AwsCredentials constructor is called via the builder
2. constructor super() is called, going to the External account credentials constructor
3. in the parent constructor, buildImpersonatedCredentials is called
4. the ImpersonatedCredentials object is built, but the source credential passed doesn't have any of the AwsCredentials specific builder variables set (AwsSecurityCredentialsSupplier for example) because the rest of the AwsCredentials constructor logic has not run yet, we are still in the parent constructor.

Since the source credential passed to the impersonated credential isn't actually correct here we have 1 of two options if we want any child class specific builder setters
1: Set the impersonatedCredentialOverride in the child constructor and just use that, ignoring the "final" impersonatedCredentials. This is what we do for executableCredentials, but the issue is that its kind of a weird pattern that we just have the unused and incorrect impersonatedCredentials object attached to the credential, and that it means we can't do any input validation on the child credential.
2: Make impersonatedCredentials not final, and just set it when we use it the first time when retrieving an access token. This lets us get rid of the override, and makes it so the impersonatedCredential is never built with the incorrect source credential.

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
aeitzman: see my reply on your first comment, building the impersonated credential in the parent constructor has some issues.

oauth2_http/java/com/google/auth/oauth2/PluggableAuthCredentials.java
```diff
@@ -192,7 +192,7 @@ public PluggableAuthCredentials createScoped(Collection<String> newScopes) {
 
   @Override
   String getCredentialSourceType() {
-    return "executable";
+    return PLUGGABLE_AUTH_METRICS_HEADER_VALUE;
```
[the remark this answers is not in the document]
aeitzman: that's what this method is used for, can rename the method if you feel strongly since its package private so wouldn't be a breaking change.

oauth2_http/java/com/google/auth/oauth2/ExternalAccountMetricsHandler.java
```diff
@@ -43,7 +43,7 @@ class ExternalAccountMetricsHandler implements java.io.Serializable {
 
   private final boolean configLifetime;
   private final boolean saImpersonation;
-  private String credentialSourceType;
+  private ExternalAccountCredentials credentials;
```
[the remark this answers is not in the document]
aeitzman: This needed to change so I could keep the metrics handler final in External account credentials. Since the metric source could change due to things in child credential specific builder and we create the metrics handler in the parent class constructor.

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
```diff
@@ -53,22 +54,44 @@
...
 public class IdentityPoolCredentials extends ExternalAccountCredentials {
 
+  static final String FILE_METRICS_HEADER_VALUE = "file";
+  static final String URL_METRICS_HEADER_VALUE = "url";
   private static final long serialVersionUID = 2471046175477275881L;
 
-  private final IdentityPoolCredentialSource identityPoolCredentialSource;
+  @Nullable private final IdentityPoolCredentialSource identityPoolCredentialSource;
+
+  @Nullable private final Supplier<String> subjectTokenSupplier;
```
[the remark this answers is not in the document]
aeitzman: no, since its just a different way to retrieve the subject token. The only place we currently do any additional logic on the subject token after retrieving it is the AwsCredentials (because we take the AWS security credentials and turn it into a signed key). All other current use cases are covered by the logic added to identity pool credentials.

oauth2_http/java/com/google/auth/oauth2/PluggableAuthCredentials.java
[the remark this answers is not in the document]
TimurSadykov: Why not the PLUGGABLE_AUTH_CREDENTIAL_SOURCE_TYPE?

lsirac (approved):

oauth2_http/java/com/google/auth/oauth2/AwsCredentials.java
```diff
@@ -66,17 +71,50 @@ public class AwsCredentials extends ExternalAccountCredentials {
...
+  @Nullable private final Supplier<AwsSecurityCredentials> awsSecurityCredentialsSupplier;
+  @Nullable private final String regionalCredentialVerificationUrlOverride;
+  @Nullable private final String region;
 
   /** Internal constructor. See {@link AwsCredentials.Builder}. */
   AwsCredentials(Builder builder) {
     super(builder);
-    this.awsCredentialSource = (AwsCredentialSource) builder.credentialSource;
+    // Check that one and only one of supplier or credential source are provided.
+    if (builder.awsSecurityCredentialsSupplier != null && builder.credentialSource != null) {
```
TimurSadykov, line 94: This and the rest of the if-else changes indicate, that you can split it. 

It seems to be cleaner to have a separate AwsProgrammaticCredential that extends this one. It can add its own source type, but hide setting the other one. And it can override whatever methods you change here. 

It is easier to understand, could this be an option?

TimurSadykov: Similarly to IdentityPool.

lsirac: Hey Timur, we appreciate your suggestions but we are not going to add new classes. The design has been approved and it is too late to make changes.

lsirac: Discussed offline, we've agreed to go with @westarle's suggestions. Thanks guys.

lsirac (changes requested):

oauth2_http/java/com/google/auth/oauth2/InternalAwsSecurityCredentialsProvider.java
```diff
@@ -0,0 +1,262 @@
...
+      EnvironmentProvider environmentProvider,
+      HttpTransportFactory transportFactory,
+      String region) {
+    this.environmentProvider = environmentProvider;
+    this.awsCredentialSource = awsCredentialSource;
+    this.transportFactory = transportFactory;
+    this.region = region;
+  }
+
+  public AwsSecurityCredentials getCredentials() throws IOException {
```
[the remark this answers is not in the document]
aeitzman: IDE was making me have this public since I guess all methods from the interface are required to be public, assuming this is why we had the mismatch for the credential sources as well. Will fix it by making the parent abstract instead of an interface

oauth2_http/java/com/google/auth/oauth2/IdentityPoolSubjectTokenProvider.java
```diff
@@ -0,0 +1,70 @@
...
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.Serializable;
+import java.nio.charset.StandardCharsets;
+import java.util.function.Supplier;
+
+abstract class IdentityPoolSubjectTokenProvider implements Serializable {
+  abstract String getSubjectToken() throws IOException;
+
+  abstract boolean isUserSupplied();
```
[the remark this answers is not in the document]
aeitzman: Need it to figure out what metric header value to use (programmatic vs file/URL) but thinking about it more, it makes more sense for the provider to just return the metrics header value instead of a bool here

oauth2_http/java/com/google/auth/oauth2/InternalAwsSecurityCredentialsProvider.java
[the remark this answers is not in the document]
aeitzman: Changed to abstract class, and made all package-private

oauth2_http/java/com/google/auth/oauth2/IdentityPoolSubjectTokenProvider.java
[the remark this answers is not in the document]
aeitzman: Changed to metrics header, and added a separate bool in the actual credentials classes to keep track of whether we should be passing the supplier in the builder - I think that makes it more clear what's happening and obvious when making changes in the future.

oauth2_http/java/com/google/auth/oauth2/InternalAwsSecurityCredentialsProvider.java
```diff
@@ -0,0 +1,262 @@
...
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableList;
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Supplier;
+import javax.annotation.Nullable;
+
+class InternalAwsSecurityCredentialsProvider implements AwsSecurityCredentialsProvider {
```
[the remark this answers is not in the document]
aeitzman: the naming of having the different stuff at the beginning makes more sense to me tbh, just seems inconsistent otherwise (I.e. we don't call it CredentialsAws). could put them in a folder I guess too which would make the file structure more readable

oauth2_http/java/com/google/auth/oauth2/AwsSecurityCredentialsProvider.java
```diff
@@ -35,13 +35,39 @@
...
+   */
+  abstract String getMetricsHeaderValue();
 
-  Supplier<AwsSecurityCredentials> getSupplier();
+  /**
+   * Gets the Aws security credential supplier.
+   *
+   * @return the Supplier used to retrieve the AWS security credentials.
+   */
+  abstract Supplier<AwsSecurityCredentials> getSupplier();
```
lsirac, line 72: Why do we expose this? getCredentials() should be enough, right?

oauth2_http/java/com/google/auth/oauth2/AwsCredentials.java
```diff
@@ -103,13 +107,15 @@ public class AwsCredentials extends ExternalAccountCredentials {
       this.awsSecurityCredentialsProvider =
           new ProgrammaticAwsSecurityCredentialsProvider(
               builder.awsSecurityCredentialsSupplier, this.awsRegion);
+      this.builtWithSupplier = true;
```
lsirac, line 110: Hmm, we shouldn't have to track this. Why not handle this in the builder so we can set the provider directly and not have to handle this logic here?

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
```diff
@@ -52,6 +52,9 @@ public class IdentityPoolCredentials extends ExternalAccountCredentials {
   private static final long serialVersionUID = 2471046175477275881L;
 
   private final IdentityPoolSubjectTokenProvider subjectTokenProvider;
+  // Boolean to tell if a Supplier was provided to build the credential. This is needed to tell
+  // whether the supplier should be passed when creating a new builder from this credential.
+  private final boolean builtWithSupplier;
```
lsirac, line 57: We don't need track this -- instead of starting with builder.subjectTokenSupplier != null, move that to the else statement, and then rely on IdentityPoolCredentialSourceType.FILE and IdentityPoolCredentialSourceType.URL.

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
```diff
@@ -136,7 +139,7 @@ public static class Builder extends ExternalAccountCredentials.Builder {
 
     Builder(IdentityPoolCredentials credentials) {
       super(credentials);
-      if (credentials.subjectTokenProvider.isUserSupplied()) {
+      if (credentials.builtWithSupplier) {
```
lsirac, line 142: If the above change is made, we don't need this here (also, probably don't need to expose supplier at all).

oauth2_http/java/com/google/auth/oauth2/IdentityPoolSubjectTokenProvider.java
```diff
@@ -43,11 +43,32 @@
...
+ * access tokens.
+ */
 abstract class IdentityPoolSubjectTokenProvider implements Serializable {
+
+  /**
+   * Gets a subject token that can be exchanged for a GCP access token.
+   *
+   * @return a valid subject token.
+   * @throws IOException
+   */
```
lsirac, line 57: Should we just call this getCredentials()?

lsirac (changes requested):

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
```diff
@@ -488,10 +473,10 @@ private static boolean isAwsCredential(Map<String, Object> credentialSource) {
...
       StsTokenExchangeRequest stsTokenExchangeRequest) throws IOException {
     // Handle service account impersonation if necessary.
-    // Internal override takes priority.
-    if (impersonatedCredentialsOverride != null) {
-      return impersonatedCredentialsOverride.refreshAccessToken();
-    } else if (impersonatedCredentials != null) {
+    if (this.shouldBuildImpersonatedCredential()) {
+      this.impersonatedCredentials = this.buildImpersonatedCredentials();
+    }
+    if (impersonatedCredentials != null) {
```
westarle, line 479: Nit: consistency for `this.impersonatedCredentials` or just `impersonatedCredentials` here.

oauth2_http/java/com/google/auth/oauth2/IdentityPoolSubjectTokenProvider.java
```diff
@@ -0,0 +1,91 @@
...
+import java.io.InputStreamReader;
+import java.io.Serializable;
+import java.nio.charset.StandardCharsets;
+import java.util.function.Supplier;
+
+/**
+ * Provider for retrieving subject tokens for {@Link IdentityPoolCredentials} to exchange for GCP
+ * access tokens.
+ */
+abstract class IdentityPoolSubjectTokenProvider implements Serializable {
```
westarle, line 50: nit: Interface?

oauth2_http/java/com/google/auth/oauth2/IdentityPoolSubjectTokenProvider.java
```diff
@@ -0,0 +1,91 @@
...
+   * @return the metrics header value.
+   */
+  abstract String getMetricsHeaderValue();
+
+  /**
+   * Gets the supplier used to retrieve the subject token.
+   *
+   * @return the supplier used for getSubjectToken.
+   */
+  abstract Supplier<String> getSupplier();
```
westarle, line 72: Maybe similar to @lsirac comment: we can probably remove getSupplier() and use getToken() directly (it would call the supplier in the programmatic case);

That said, it might be better if customers provided a `SubjectTokenProvider` directly.  The (internal) API allows us to cheat and deal with `IOException` in a way we don't allow via `Supplier<String>`;  I see both problems of us eating `IOException` and turning them into RTEs as well as us catching RTEs and turning them into not-retryable `IOException` -- even though we can anticipate that there will be network, IPC or File errors present based on all of our own implementations.

So...my suggestion is to ditch "Supplier" and offer "Provider" to our customers.  Now everyone has the same API `String getSubjectToken() throws IOException`; they can implement Retryable on their exceptions.  You might need to find another way to support the metrics header value if you're worried about clients overriding it, but overall the code should be simpler.

oauth2_http/java/com/google/auth/oauth2/InternalAwsSecurityCredentialsProvider.java
```diff
@@ -0,0 +1,282 @@
...
+        return false;
+      }
+    }
+    return true;
+  }
+
+  @VisibleForTesting
+  boolean shouldUseMetadataServer() {
+    return (!canRetrieveRegionFromEnvironment()
+        || !canRetrieveSecurityCredentialsFromEnvironment());
```
westarle, line 216: This can happen in the future, but:

now that you have an indirection to choose token providers, this logic can happen when installing a token provider and therefore break this class into one for environment and one for MDS.

oauth2_http/java/com/google/auth/oauth2/IdentityPoolSubjectTokenProvider.java
```diff
@@ -0,0 +1,91 @@
...
+
+  protected static String parseToken(
```
westarle, line 74: Can this live with FileIDPoolCredential?

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
```diff
@@ -88,76 +103,17 @@ public AccessToken refreshAccessToken() throws IOException {
 
   @Override
   public String retrieveSubjectToken() throws IOException {
-    if (identityPoolCredentialSource.credentialSourceType
-        == IdentityPoolCredentialSource.IdentityPoolCredentialSourceType.FILE) {
-      return retrieveSubjectTokenFromCredentialFile();
-    }
-    return getSubjectTokenFromMetadataServer();
+    return this.subjectTokenProvider.getSubjectToken();
```
westarle, line 106: Here's an example of us cheating in our own implementation, offering `Retryable` `IOExceptions` for our own implementations but losing our customers RTEs into a generic `IOException`.

oauth2_http/java/com/google/auth/oauth2/IdentityPoolSubjectTokenProvider.java
aeitzman: Made this change - I wanted to keep the supplier so we could have a functional interface for users so they could use a lambda instead of creating a whole class, but I agree that its creating more problems than its worth. Refactored everything to be public. I named the objects IdentityPoolSubjectTokenSupplier and AwsSecurityCredentialSupplier instead of provider just because I call the similar function/object supplier in other PR's for other languages (still in flight, not merged) but don't feel super strongly about that either way.

oauth2_http/java/com/google/auth/oauth2/InternalAwsSecurityCredentialsProvider.java
aeitzman: Unfortunately I think that would be a breaking change because the currently implementation checks the environment variables when get security credentials is called instead of in the constructor, so technically its possible for a user to create the credential and set environment variables after.

oauth2_http/java/com/google/auth/oauth2/IdentityPoolSubjectTokenProvider.java
aeitzman: Both file and URL need this, moved to file provider for now since it can't live in the interface and I don't think its worth creating a new class to hold 1 utility method.

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
aeitzman: I believe this is resolved by the refactor to have the supplier/provider be public to users but feel free to reopen if I'm not understanding correctly

oauth2_http/java/com/google/auth/oauth2/AwsCredentials.java
aeitzman: Figured out a better way to handle in the builder, we can just check whether the builder has a credential source, and only set the supplier if it doesn't

westarle (commented):
This is looking pretty good now, can you please comment on the coverage of the change in unit and functional tests?

aeitzman:
> This is looking pretty good now, can you please comment on the coverage of the change in unit and functional tests?

Added unit tests for the builders using the new supplier objects, as well as tests for the new default values. Also added tests for the credentials built with the suppliers (happy path with/without service account impersonation and error paths). Did not add tests for the internal suppliers since those paths are covered by existing testing. Also added integration tests for AWS and IdentityPool subject token supplier built credentials.

oauth2_http/java/com/google/auth/oauth2/IdentityPoolSubjectTokenSupplier.java
```diff
@@ -31,50 +31,21 @@
...
-  @Override
-  String getMetricsHeaderValue() {
-    return PROGRAMMATIC_METRICS_HEADER_VALUE;
-  }
-
-  @Override
-  Supplier<String> getSupplier() {
-    return this.subjectTokenSupplier;
-  }
+  String getSubjectToken() throws IOException;
```
lsirac, line 50: Why not getCredentials() for a more consist API surface?

lsirac (approved):
LGTM thanks Alex 💯

oauth2_http/java/com/google/auth/oauth2/IdentityPoolSubjectTokenSupplier.java
westarle: my 2c: if it returns a credentials object yes, otherwise I prefer this descriptive name for the string.

westarle (approved):
If @lsirac or another reviewer has covered the Java details for bugs etc. this looks good to me.

PR #1345, opened 2023-12-18
docs: Update README.md to link to Cloud authentication documentation rather than AIPs

piaxc:
The AIPs are not a good experience for customers, they are more of an internal resource, at least for auth. The Cloud authentication documentation includes a section about Application Default Credentials now, so that's a better place to send users.


westarle (approved):

TimurSadykov (approved):
LGTM

lsirac (approved):

PR #1346, opened 2023-12-26
feat: adds universe domain support for compute credentials

TimurSadykov:
This adds universe domain support to GCE credentials.

In case explicit universe_domain was provided - it takes priority
Otherwise - GCE credential tries to obtain a universe domain from Metadata service

If request fails with 404 or request returns an empty string - universe domain is set to GDU
If any other error - throws a retryable exception

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
@@ -205,8 +205,8 @@ public class ComputeEngineCredentials extends GoogleCredentials
    *
    * <p>Returns an explicit universe domain if it was provided during credential initialization.
    *
-   * <p>Returns the {@link Credentials#GOOGLE_DEFAULT_UNIVERSE} if universe domain endpoint is
-   * not found (404) or returns an empty string.
+   * <p>Returns the {@link Credentials#GOOGLE_DEFAULT_UNIVERSE} if universe domain endpoint is not
+   * found (404) or returns an empty string.
    *
    * <p>Otherwise, returns universe domain from GCE metadata service.
    *
--- a/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
@@ -221,12 +221,14 @@ public class ComputeEngineCredentials extends GoogleCredentials
       return super.getUniverseDomain();
     }
 
-    if (universeDomainFromMetadata != null) {
+    synchronized (this) {
+      if (universeDomainFromMetadata != null) {
+        return universeDomainFromMetadata;
+      }
+
+      universeDomainFromMetadata = getUniverseDomainFromMetadata();
       return universeDomainFromMetadata;
     }
-
-    universeDomainFromMetadata = getUniverseDomainFromMetadata();
-    return universeDomainFromMetadata;
   }
 
   private String getUniverseDomainFromMetadata() throws IOException {
--- a/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
@@ -657,6 +657,12 @@ public class ComputeEngineCredentials extends GoogleCredentials
       return this;
     }
 
+    @CanIgnoreReturnValue
+    public Builder setQuotaProjectId(String quotaProjectId) {
+      super.quotaProjectId = quotaProjectId;
+      return this;
+    }
+
     public HttpTransportFactory getHttpTransportFactory() {
       return transportFactory;
     }
--- a/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
@@ -156,7 +156,7 @@ public class ComputeEngineCredentials extends GoogleCredentials
   @Override
   public GoogleCredentials createScoped(Collection<String> newScopes) {
     ComputeEngineCredentials.Builder builder =
-        ComputeEngineCredentials.newBuilder()
+        this.toBuilder()
             .setHttpTransportFactory(transportFactory)
             .setScopes(newScopes);
     return new ComputeEngineCredentials(builder);
@@ -219,7 +219,7 @@ public class ComputeEngineCredentials extends GoogleCredentials
    */
   @Override
   public String getUniverseDomain() throws IOException {
-    if (!isDefaultUniverseDomain()) {
+    if (isExplicitUniverseDomain()) {
       return super.getUniverseDomain();
     }
 
@@ -252,6 +252,8 @@ public class ComputeEngineCredentials extends GoogleCredentials
       throw new GoogleAuthException(true, cause);
     }
     String responseString = response.parseAsString();
+
+    /* Earlier versions of MDS that supports universe_domain return empty string instead of GDU. */
     if (responseString.isEmpty()) {
       return Credentials.GOOGLE_DEFAULT_UNIVERSE;
     }
@@ -521,6 +523,9 @@ public class ComputeEngineCredentials extends GoogleCredentials
     if (!(obj instanceof ComputeEngineCredentials)) {
       return false;
     }
+    if (!super.equals(obj)) {
+      return false;
+    }
     ComputeEngineCredentials other = (ComputeEngineCredentials) obj;
     return Objects.equals(this.transportFactoryClassName, other.transportFactoryClassName)
         && Objects.equals(this.scopes, other.scopes)
@@ -625,6 +630,7 @@ public class ComputeEngineCredentials extends GoogleCredentials
     }
 
     protected Builder(ComputeEngineCredentials credentials) {
+      super(credentials);
       this.transportFactory = credentials.transportFactory;
       this.scopes = credentials.scopes;
     }
--- a/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
@@ -41,6 +41,8 @@ import com.google.api.client.http.HttpResponseException;
 import com.google.api.client.http.HttpStatusCodes;
 import com.google.api.client.json.JsonObjectParser;
 import com.google.api.client.util.GenericData;
+import com.google.auth.Credentials;
+import com.google.auth.Retryable;
 import com.google.auth.ServiceAccountSigner;
 import com.google.auth.http.HttpTransportFactory;
 import com.google.common.annotations.VisibleForTesting;
@@ -79,12 +81,12 @@ public class ComputeEngineCredentials extends GoogleCredentials
     implements ServiceAccountSigner, IdTokenProvider {
 
   // Decrease timing margins on GCE.
-  // This is needed because GCE VMs maintain their own OAuth cache that expires T-5mins, attempting
+  // This is needed because GCE VMs maintain their own OAuth cache that expires T-4 mins, attempting
   // to refresh a token before then, will yield the same stale token. To enable pre-emptive
   // refreshes, the margins must be shortened. This shouldn't cause problems since the clock skew
   // on the VM and metadata proxy should be non-existent.
   static final Duration COMPUTE_EXPIRATION_MARGIN = Duration.ofMinutes(3);
-  static final Duration COMPUTE_REFRESH_MARGIN = Duration.ofMinutes(4);
+  static final Duration COMPUTE_REFRESH_MARGIN = Duration.ofMinutes(3).plusSeconds(45);
 
   private static final Logger LOGGER = Logger.getLogger(ComputeEngineCredentials.class.getName());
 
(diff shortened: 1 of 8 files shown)


einobie (approved):
Ok

oauth2_http/javatests/com/google/auth/oauth2/MockMetadataServerTransport.java
```diff
@@ -83,134 +85,152 @@ public void setIdToken(String idToken) {
   @Override
   public LowLevelHttpRequest buildRequest(String method, String url) throws IOException {
     if (url.equals(ComputeEngineCredentials.getTokenServerEncodedUrl())) {
+      return getMockRequestForTokenEndpoint(url);
```
TimurSadykov, line 88: This diff might be tricky to read. What I did is I factored out all the if-statements contents in a separate private methods. Some are big statements that made this method hard to read.

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -79,12 +81,12 @@ public class ComputeEngineCredentials extends GoogleCredentials
...
 
   // Decrease timing margins on GCE.
-  // This is needed because GCE VMs maintain their own OAuth cache that expires T-5mins, attempting
+  // This is needed because GCE VMs maintain their own OAuth cache that expires T-4 mins, attempting
   // to refresh a token before then, will yield the same stale token. To enable pre-emptive
   // refreshes, the margins must be shortened. This shouldn't cause problems since the clock skew
   // on the VM and metadata proxy should be non-existent.
   static final Duration COMPUTE_EXPIRATION_MARGIN = Duration.ofMinutes(3);
-  static final Duration COMPUTE_REFRESH_MARGIN = Duration.ofMinutes(4);
+  static final Duration COMPUTE_REFRESH_MARGIN = Duration.ofMinutes(3).plusSeconds(45);
```
TimurSadykov, line 89: This is unrelated to TPC but included since it is simply an update and related to GCE credential: b/311414224

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -172,7 +180,8 @@ public GoogleCredentials createScoped(
    * @return new ComputeEngineCredentials
    */
   public static ComputeEngineCredentials create() {
-    return new ComputeEngineCredentials(null, null, null);
+    ComputeEngineCredentials.Builder builder = ComputeEngineCredentials.newBuilder();
```
westarle, line 183: nit: no need to name the builder here (just pass it into the constructor)

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
westarle: Nit: In general, we should break out any unnecessary refactoring or bugfixes to separate CLs in case we need rollbacks, etc.  This also makes reviews and fixes more predictable.

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -192,6 +201,61 @@ String createTokenUrlWithScopes() {
     return tokenUrl.toString();
   }
 
+  /**
+   * Gets the universe domain from the GCE metadata server.
+   *
+   * <p>Returns an explicit universe domain if it was provided during credential initialization.
+   *
+   * <p>Returns the {@link Credentials#GOOGLE_DEFAULT_UNIVERSE} if universe domain endpoint is
```
westarle, line 209: "if universe domain endpoint is unavailable" --> "if universe domain endpoint is not found (404)"

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -192,6 +201,61 @@ String createTokenUrlWithScopes() {
...
+   * <p>Any above value is cached for the credential lifetime.
+   *
+   * @throws IOException if a call to GCE metadata service was unsuccessful. Check if exception
+   *     implements the {@link Retryable} and {@code isRetryable()} will return true if the
+   *     operation may be retried.
+   * @return string representing a universe domain in the format some-domain.xyz
+   */
+  @Override
+  public String getUniverseDomain() throws IOException {
+    if (!isDefaultUniverseDomain()) {
```
westarle, line 223: I think I'd rather break the contract and have the universe also fetched from MDS if the creds are fetched from MDS, for consistency.  WDYT?

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -192,6 +201,61 @@ String createTokenUrlWithScopes() {
...
+      return super.getUniverseDomain();
+    }
+
+    synchronized (this) {
+      if (universeDomainFromMetadata != null) {
+        return universeDomainFromMetadata;
+      }
+
+      universeDomainFromMetadata = getUniverseDomainFromMetadata();
```
westarle, line 232: We should avoid holding the mutex while making a network call.  Since the response should always provide the same universe, we could:

```
synch(this) {
  if (this.uDFM != null) { 
    return this.uDFM;
  }
}

String uDFM = getUDFM();

synchronized (this) {
  this.uDFM = uDFM;
}
return uDFM;

File: oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java [hunk 7]
```diff
@@ -192,6 +201,61 @@ String createTokenUrlWithScopes() {
...
+          new IOException(
+              String.format(
+                  "Unexpected Error code %s trying to get universe domain"
+                      + " from Compute Engine metadata for the default service account: %s",
+                  statusCode, response.parseAsString()));
+      throw new GoogleAuthException(true, cause);
+    }
+    String responseString = response.parseAsString();
+    if (responseString.isEmpty()) {
+      return Credentials.GOOGLE_DEFAULT_UNIVERSE;
```
westarle, line 254: When would this happen?

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
TimurSadykov: thanks, makes sense )

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
TimurSadykov: There was an intermediate design of MDS where endpoint returns empty string for default universe. It was quickly abandoned, but made it into the wild already :) Now we expect either 404 or a non-empty value, but adding this just in case the older build persists somewhere.

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
TimurSadykov: i was hesitant as well :) I think one key scenario for explicit credential is testing, and it might be relevant for GCE credential as well.

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
TimurSadykov: Agree, noted )

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
TimurSadykov: https://docs.google.com/document/d/1Fk4XRywIKfnVzKUn3I2HffZWrmxhtnnBXWi0-lU0D2M/edit?resourcekey=0-CDsW8urip79pyXK4MZuETw&disco=AAABAOEGSPE

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -192,6 +201,61 @@ String createTokenUrlWithScopes() {
...
+   * Gets the universe domain from the GCE metadata server.
+   *
+   * <p>Returns an explicit universe domain if it was provided during credential initialization.
+   *
+   * <p>Returns the {@link Credentials#GOOGLE_DEFAULT_UNIVERSE} if universe domain endpoint is
+   * unavailable or returns an empty string.
+   *
+   * <p>Otherwise, returns universe domain from GCE metadata service.
+   *
+   * <p>Any above value is cached for the credential lifetime.
```
lqiu96, line 214: qq, What is the default credential lifetime? Say if the lifetime is like 1 hour (as an example), would that mean that a subsequent `getUniverseDomain()` call in an hour would make another network call since the universe domain value is no longer cached?

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -192,6 +201,61 @@ String createTokenUrlWithScopes() {
...
+    }
+    if (statusCode != HttpStatusCodes.STATUS_CODE_OK) {
+      IOException cause =
+          new IOException(
+              String.format(
+                  "Unexpected Error code %s trying to get universe domain"
+                      + " from Compute Engine metadata for the default service account: %s",
+                  statusCode, response.parseAsString()));
+      throw new GoogleAuthException(true, cause);
+    }
```
lqiu96, line 251: Just to confirm: Any non-successful status code besides 404 Not Found is always retryable? Are there any possible non-successful status codes that we should not retry?

TimurSadykov: It could be the case, but so far unknown. We know that current behavior to retry everything - so it is safe to continue to do so. We have an item to investigate MDS retryable errors, but it might happen early in 24.

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
TimurSadykov: Agree, updated

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
TimurSadykov: Updated, thanks!

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
TimurSadykov: There are two key entities: credential and token

Credential lifetime is the object lifetime. Effectively unlimited. 
Token lifetime is 1 hour. 

The universe domain is a property of the credential. We fetch it once and cache for object's lifetime. Every token for the credential expected to target that universe.

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
TimurSadykov: Per Slava, we want to maintain the contract here

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
westarle: Thanks, maybe put a comment here explaining.

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
westarle: If the rule is to return the universe domain set by the user for testing, we should do that consistently whether or not it is the 'default'.  To keep the contract here you would need to return GDU if that was set in the construction of the object.  Otherwise, I can't force a GDU value in my tests.

To do that, you might do something like update the builder for this class to set a "user set universe domain" flag when users call `setUniverseDomain()` and consult that bit here.

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
lqiu96: Understood. Thanks!

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
lqiu96: Understood, thanks! nit: Perhaps update to `The Universe Domain value is cached...`?

lqiu96 (approved):

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
TimurSadykov: Agree )

arithmetic1728 (approved):

lsirac (approved):

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
TimurSadykov: Oh you are right... somehow I fixed on testing the custom universe domain case.

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
TimurSadykov: Done

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
TimurSadykov: Added the isExplicitUniverseDomain

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -233,6 +246,11 @@ public String getUniverseDomain() throws IOException {
     return this.universeDomain;
   }
 
+  @VisibleForTesting
```
westarle, line 249: Please put detailed documentation here to indicate the purpose of this field and expected use by subclasses.

westarle (approved):

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
TimurSadykov: SG

PR #1352, opened 2024-01-10
feat: Modify the refresh window to match go/async-token-refresh. Serverless tokens are cached until 4 minutes before expiration, so 4 minutes is the ideal refresh window.

clundin25:

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
@@ -68,7 +68,7 @@ import javax.annotation.Nullable;
 public class OAuth2Credentials extends Credentials {
 
   private static final long serialVersionUID = 4556936364828217687L;
-  static final Duration DEFAULT_EXPIRATION_MARGIN = Duration.ofMinutes(5);
+  static final Duration DEFAULT_EXPIRATION_MARGIN = Duration.ofMinutes(4);
   static final Duration DEFAULT_REFRESH_MARGIN = Duration.ofMinutes(6);
   private static final ImmutableMap<String, List<String>> EMPTY_EXTRA_HEADERS = ImmutableMap.of();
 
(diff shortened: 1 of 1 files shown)


arithmetic1728 (approved):

oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java
```diff
@@ -68,7 +68,7 @@
 public class OAuth2Credentials extends Credentials {
 
   private static final long serialVersionUID = 4556936364828217687L;
-  static final Duration DEFAULT_EXPIRATION_MARGIN = Duration.ofMinutes(5);
+  static final Duration DEFAULT_EXPIRATION_MARGIN = Duration.ofMinutes(3).plusSeconds(45);
   static final Duration DEFAULT_REFRESH_MARGIN = Duration.ofMinutes(6);
```
TimurSadykov, line 72: I think you want to change both? This to 4 mins or slightly below end expiration to below 3min. 

Right now it will start refreshing at 6 mins and get insta cached response for next 2 mins.

TimurSadykov (changes requested):
Left comment on changing both values

clundin25:
> Left comment on changing both values

Good point. I have revised the change

TimurSadykov (approved):
LGTM, thanks!

TimurSadykov:
@clundin25 Hey sorry for the late comment... but it looks like those margins aren't covered with tests, could you please add those? There are margin related tests in ComputeCredentialTests

clundin25:
@TimurSadykov It seems that it is currently just testing the getters. Is there an additional test you had in mind https://github.com/googleapis/google-auth-library-java/blob/main/oauth2_http/javatests/com/google/auth/oauth2/ComputeEngineCredentialsTest.java#L200-L205?

TimurSadykov:
@clundin25 it is a bit more than a getter validation. The test you highlighted validates that compute credentials has their specific margins set. As i understand similar tests are missing for other credentials that supposed to use defaults. It could be that custom margins set up for some credentials and its a bug.

PR #1364, opened 2024-02-05
fix: add missing copyright header

aeitzman:
Adds missing copyright header on external account supplier context class.


TimurSadykov (approved):
lgtm

PR #1368, opened 2024-02-08
chore: enabling ut coverege check

TimurSadykov:
Fixing the current coverage so it does not go lower. 

Setting current limit for the main oauth2 package. 
Configuring SonarCloud build for visualization (more tuning required). Right now shows 0s, after merge should show real numbers.


oauth2_http/javatests/com/google/auth/oauth2/GoogleCredentialsTest.java
```diff
@@ -474,333 +469,6 @@ public void fromStream_identityPoolCredentials_providesToken() throws IOExceptio
     TestUtils.assertContainsBearerToken(metadata, transportFactory.transport.getAccessToken());
   }
 
-  @Test
```
TimurSadykov, line 477: removing to validate

sai-sunder-s (approved):

lqiu96:
I think the changes LGTM, thanks @TimurSadykov. 

CC: @suztomo if you could also take a quick look at the configs here. Thanks!

PR #1374, opened 2024-03-07
feat: Support Workload Identity Federation on AWS ECS/Fargate

### LINKED ISSUE #957: Workload Identity Federation should support ECS Fargate container credentials
**Is your feature request related to a problem? Please describe.**
I am trying to use Workload Identity Federation on AWS ECS Fargate. It seems that `AwsCredentials.getAwsSecurityCredentials` ([code](https://github.com/googleapis/google-auth-library-java/blob/acc1ce3603435f1c0cf23b8606af71b05e566f2f/oauth2_http/java/com/google/auth/oauth2/AwsCredentials.java#L319-L325
)) only supports EC2 metadata service and does not support [ECS Fargate container credential URL](https://docs.aws.amazon.com/sdkref/latest/guide/feature-container-credentials.html).

**Describe the solution you'd like**
Google auth SDK should support [ECS Fargate container credential URL](https://docs.aws.amazon.com/sdkref/latest/guide/feature-container-credentials.html) through environment variable `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI`.

**Describe alternatives you've considered**
The alternative to use Workload Identity Federation on ECS Fargate is implementing the [REST authentication](https://cloud.google.com/iam/docs/using-workload-identity-federation#rest) myself.

jurgispods:

Fixes #957 ☕️

This PR adds functionality to detect if the code is running in an ECS container. In that case, the private container metadata endpoint is used to obtain temporary credentials. 

This makes it possible to use Workload Identity Federation on ECS without using a workaround like manually obtaining temporary session credentials and exporting them as env vars (see [here](https://stackoverflow.com/a/72588646/2429404)).

I was able to successfully test this PR by running an ECS task which connects to Bigquery via Workload Identity Federation.

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/InternalAwsSecurityCredentialsSupplier.java
+++ b/oauth2_http/java/com/google/auth/oauth2/InternalAwsSecurityCredentialsSupplier.java
@@ -109,14 +109,8 @@ class InternalAwsSecurityCredentialsSupplier implements AwsSecurityCredentialsSu
           "Unable to determine the AWS IAM role name. The credential source does not contain the"
               + " url field.");
     }
-    String roleName = retrieveResource(awsCredentialSource.url, "IAM role", metadataRequestHeaders);
 
-    // Retrieve the AWS security credentials by calling the endpoint specified by the credential
-    // source.
-    String awsCredentials =
-        retrieveResource(
-            awsCredentialSource.url + "/" + roleName, "credentials", metadataRequestHeaders);
-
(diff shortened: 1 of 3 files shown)


lsirac:
Hi @jurgispods, we support [custom credential suppliers](https://cloud.google.com/iam/docs/workload-identity-federation-with-kubernetes#eks). This is the approach we recommend for now.

lsirac:
Yes, we may add it in the future but it would be part of a larger overhaul. I am going to close this for now. If we do add support for this I will tag you.

PR #1383, opened 2024-04-09
fix: makes default token url universe aware

aeitzman:
Makes the default sts token url logic in external account credentials universe aware.

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
@@ -73,7 +73,7 @@ public abstract class ExternalAccountCredentials extends GoogleCredentials {
   static final String EXTERNAL_ACCOUNT_FILE_TYPE = "external_account";
   static final String EXECUTABLE_SOURCE_KEY = "executable";
 
-  static final String DEFAULT_TOKEN_URL = "https://sts.googleapis.com/v1/token";
+  static final String DEFAULT_TOKEN_URL = "https://sts.{UNIVERSE_DOMAIN}/v1/token";
   static final String PROGRAMMATIC_METRICS_HEADER_VALUE = "programmatic";
 
   private final String transportFactoryClassName;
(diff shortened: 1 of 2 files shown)


oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
```diff
@@ -235,7 +235,19 @@ protected ExternalAccountCredentials(ExternalAccountCredentials.Builder builder)
...
+      try {
+        this.tokenUrl = DEFAULT_TOKEN_URL.replace("{UNIVERSE_DOMAIN}", this.getUniverseDomain());
+      } catch (IOException e) {
+        // Throwing an IOException would be a breaking change, so wrap it here.
+        throw new IllegalStateException(
+            "Error occurred when attempting to retrieve universe domain.", e);
+      }
+    } else {
+      this.tokenUrl = builder.tokenUrl;
+    }
```
lsirac, line 249:
```suggestion
    this.tokenUrl = builder.tokenUrl;
    if (this.tokenUrl == null) {
      try {
        this.tokenUrl = DEFAULT_TOKEN_URL.replace("{UNIVERSE_DOMAIN}", this.getUniverseDomain());
      } catch (IOException e) {
        // Throwing an IOException would be a breaking change, so wrap it here.
        // This should not happen for this credential type.
        throw new IllegalStateException(
            "Error occurred when attempting to retrieve universe domain.", e);
      }
    }
```

lsirac (approved):

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
aeitzman: had to add the else back in since tokenUrl is final so can only be set once.

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
```diff
@@ -235,7 +235,20 @@ protected ExternalAccountCredentials(ExternalAccountCredentials.Builder builder)
...
     this.clientId = builder.clientId;
     this.clientSecret = builder.clientSecret;
-    this.tokenUrl = builder.tokenUrl == null ? DEFAULT_TOKEN_URL : builder.tokenUrl;
+
+    if (builder.tokenUrl == null) {
+      try {
+        this.tokenUrl = DEFAULT_TOKEN_URL.replace("{UNIVERSE_DOMAIN}", this.getUniverseDomain());
+      } catch (IOException e) {
+        // Throwing an IOException would be a breaking change, so wrap it here.
+        // This should not happen for this credential type.
```
westarle, line 244: Is there a place we can read this value for this credential type?

Can we override getUniverseDomain() and strip the exception if it is always available?

aeitzman: Yeah we could override it, I think we would still have the dead code in the overridden function, but I may be misunderstanding what you mean by strip the exception, do you just mean doing this? - 

```
  @Override
  public String getUniverseDomain() {
    try {
      return super.getUniverseDomain();
    } catch (IOException e) {
      throw new RuntimeException(e);
    }
  }
```

The other easier option would be to make universeDomain protected instead of private in GoogleCredentials so the external account credential could just read it directly instead of calling the parent getUniverseDomain() function.

westarle (approved):

westarle:
Coverage test is complaining about new getUniverseDomain, where the 2nd branch isn't exercisable by tests.  We can ignore for this PR.

PR #1400, opened 2024-05-09
feat: Support querying S2A Addresses from MDS

rmehta19:
Add utility to get S2A address from MDS MTLS autoconfiguration endpoint. 

This utility will be used when creating mTLS channel using S2A Java Client, which takes S2A Address as input to create S2AChannelCredentials.

Parallel change in go: https://github.com/googleapis/google-api-go-client/pull/1874
S2A Java client: https://github.com/grpc/grpc-java/pull/11113 

Resolving b/376258193 means that S2A.java is no longer experimental


rmehta19:
cc: @xmenxk

oauth2_http/java/com/google/auth/oauth2/S2A.java
```diff
@@ -0,0 +1,107 @@
...
+import javax.annotation.concurrent.ThreadSafe;
+
+/**
+ * Utilities to fetch the S2A (Secure Session Agent) address from the mTLS configuration.
+ *
+ * <p>mTLS configuration is queried from the MDS MTLS Autoconfiguration endpoint.
+ */
+@ThreadSafe
+public final class S2A {
+  public static final String DEFAULT_METADATA_SERVER_URL = "http://169.254.169.254";
```
xmenxk, line 22: It seems like getting metadata server address is already implemented here: https://github.com/googleapis/google-auth-library-java/blob/dde6876154d6e7ff43dbbf7b75c5205eadd679a5/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java#L467

should we reuse that definition?

oauth2_http/java/com/google/auth/oauth2/S2A.java
```diff
@@ -0,0 +1,107 @@
...
+      HttpResponse response = request.execute();
+
+      if (!response.isSuccessStatusCode()) {
+        return MtlsConfig.createBuilder().build();
+      }
+
+      InputStream content = response.getContent();
+      if (content == null) {
+        return MtlsConfig.createBuilder().build();
+      }
```
xmenxk, line 89: is it possible to reuse the code below  for querying mds endpoint?
https://github.com/googleapis/google-auth-library-java/blob/dde6876154d6e7ff43dbbf7b75c5205eadd679a5/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java#L338

oauth2_http/java/com/google/auth/oauth2/S2A.java
rmehta19: Done. Used ComputeEngineCredentials.getMeadataServerUrl()

oauth2_http/java/com/google/auth/oauth2/S2A.java
rmehta19: We could use this function to get the MDS HttpResponse, but it would only replace a few lines (~7) in this function: 

creating the HttpRequest and executing it to get the response, replaced with

```response = getMetadataResponse(url)```.

However, in order to do this, we would also have to:
- ignore the ```ComputeEngineCredentials``` specific errors thrown by the function, because we only care if we are able to successfully create a response (aligning with Go implementation, return empty S2A Address if any error). This technically works, although I am not sure it is best practice.
- ```getMetadataResponse(String url)``` is not static, so we would have to create an instance of ```ComputeEngineCredentials``` to use it.

WDYT?

File: oauth2_http/java/com/google/auth/oauth2/S2A.java [hunk 3]
```diff
@@ -0,0 +1,107 @@
...
+      String url = getMdsMtlsEndpoint();
+      GenericUrl genericUrl = new GenericUrl(url);
+      HttpRequest request =
+          transportFactory.create().createRequestFactory().buildGetRequest(genericUrl);
+      JsonObjectParser parser = new JsonObjectParser(OAuth2Utils.JSON_FACTORY);
+      request.setParser(parser);
+      request.getHeaders().set(METADATA_FLAVOR, GOOGLE);
+      request.setThrowExceptionOnExecuteError(false);
+      HttpResponse response = request.execute();
```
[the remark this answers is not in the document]
rmehta19: I added retry logic, similar to https://github.com/googleapis/google-auth-library-java/blob/dde6876154d6e7ff43dbbf7b75c5205eadd679a5/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java#L435

rmehta19:
@westarle, friendly ping :). Please review when you get a chance.

oauth2_http/java/com/google/auth/oauth2/MtlsConfig.java
```diff
@@ -0,0 +1,58 @@
+package com.google.auth.oauth2;
+
+import com.google.errorprone.annotations.CanIgnoreReturnValue;
+
+/** Holds an mTLS configuration (consists of address of S2A) retrieved from the Metadata Server. */
+public final class MtlsConfig {
```
clundin25, line 6: I think this should be re-branded to `S2AConfig`. `MtlsConfig` is much more generic than the settings in this class.

oauth2_http/java/com/google/auth/oauth2/S2A.java
```diff
@@ -0,0 +1,112 @@
...
+ */
+@ThreadSafe
+public final class S2A {
+  public static final String MTLS_CONFIG_ENDPOINT =
+      "/computeMetadata/v1/instance/platform-security/auto-mtls-configuration";
+
+  public static final String METADATA_FLAVOR = "Metadata-Flavor";
+  public static final String GOOGLE = "Google";
+  private static final int MAX_MDS_PING_TRIES = 3;
+  private static final String PARSE_ERROR_S2A = "Error parsing Mtls Auto Config response.";
```
clundin25, line 28: Is there any existing code for these constants?

oauth2_http/java/com/google/auth/oauth2/S2A.java
```diff
@@ -0,0 +1,112 @@
...
+import javax.annotation.concurrent.ThreadSafe;
+
+/**
+ * Utilities to fetch the S2A (Secure Session Agent) address from the mTLS configuration.
+ *
+ * <p>mTLS configuration is queried from the MDS MTLS Autoconfiguration endpoint.
+ */
+@ThreadSafe
+public final class S2A {
+  public static final String MTLS_CONFIG_ENDPOINT =
```
clundin25, line 22:
```suggestion
  public static final String S2A_CONFIG_ENDPOINT_POSTFIX =
```

oauth2_http/java/com/google/auth/oauth2/S2A.java
clundin25, line 28: I think this error message should be expanded to explain what went wrong and how it can be triaged.

oauth2_http/java/com/google/auth/oauth2/S2A.java
```diff
@@ -0,0 +1,112 @@
...
+  public void setHttpTransportFactory(HttpTransportFactory tf) {
+    this.transportFactory = tf;
+  }
+
+  /** @return the mTLS S2A Address from the mTLS config. */
+  public synchronized String getMtlsS2AAddress() {
+    if (config == null) {
+      config = getMdsMtlsConfig();
+    }
+    return config.getMtlsS2AAddress();
```
clundin25, line 45: Could a valid config be a requirement to instantiate this class?

oauth2_http/java/com/google/auth/oauth2/S2A.java
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  /**
+   * Queries the MDS mTLS Autoconfiguration endpoint and returns the {@link MtlsConfig}.
+   *
+   * <p>Returns {@link MtlsConfig} with empty addresses on error.
+   *
+   * @return the {@link MtlsConfig}.
+   */
+  private MtlsConfig getMdsMtlsConfig() {
```
clundin25, line 63:
```suggestion
  private MtlsConfig getS2AConfigFromMDS() {
```

oauth2_http/java/com/google/auth/oauth2/S2A.java
```diff
@@ -0,0 +1,112 @@
...
+    String url = getMdsMtlsEndpoint();
+    GenericUrl genericUrl = new GenericUrl(url);
+    
+    for (int i = 0; i < MAX_MDS_PING_TRIES; i++) {
+      try {
+        if (transportFactory == null) {
+          transportFactory =
+              Iterables.getFirst(
+                  ServiceLoader.load(HttpTransportFactory.class), OAuth2Utils.HTTP_TRANSPORT_FACTORY);
+        }
```
clundin25, line 76: Should this be outside of the loop?

oauth2_http/java/com/google/auth/oauth2/S2A.java
```diff
@@ -0,0 +1,112 @@
...
+        HttpRequest request =
+            transportFactory.create().createRequestFactory().buildGetRequest(genericUrl);
+        JsonObjectParser parser = new JsonObjectParser(OAuth2Utils.JSON_FACTORY);
```
clundin25, line 79: Should this be outside of the loop?

oauth2_http/java/com/google/auth/oauth2/S2A.java
```diff
@@ -0,0 +1,112 @@
...
+        mtlsS2AAddress = OAuth2Utils.validateString(responseData, "mtls_address", PARSE_ERROR_S2A);
+      } catch (IOException e) {
+        continue;
+      }
+      return MtlsConfig.createBuilder()
+          .setPlaintextS2AAddress(plaintextS2AAddress)
+          .setMtlsS2AAddress(mtlsS2AAddress)
+          .build();
+    }
+    return MtlsConfig.createBuilder().build();
```
clundin25, line 105: How will callers know if a valid config is returned?

oauth2_http/java/com/google/auth/oauth2/S2A.java
```diff
@@ -0,0 +1,107 @@
...
+public final class S2A {
+  public static final String DEFAULT_METADATA_SERVER_URL = "http://169.254.169.254";
+  public static final String MTLS_CONFIG_ENDPOINT =
+      "/computeMetadata/v1/instance/platform-security/auto-mtls-configuration";
+
+  public static final String METADATA_FLAVOR = "Metadata-Flavor";
+  public static final String GOOGLE = "Google";
+  private static final String PARSE_ERROR_S2A = "Error parsing Mtls Auto Config response.";
+
+  private MtlsConfig config;
```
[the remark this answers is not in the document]
rmehta19: We'll rename to S2AConfig as Carl suggested above.

oauth2_http/java/com/google/auth/oauth2/MtlsConfig.java
rmehta19: Done in https://github.com/googleapis/google-auth-library-java/pull/1400/commits/359fd432480bfd92440ee680e1459bfe0e782fb9

oauth2_http/java/com/google/auth/oauth2/S2A.java
rmehta19: Done in https://github.com/googleapis/google-auth-library-java/pull/1400/commits/bce602e84717d76c284e55e3cd6c8c69c99007e9

oauth2_http/java/com/google/auth/oauth2/S2A.java
rmehta19: Done in https://github.com/googleapis/google-auth-library-java/pull/1400/commits/bce602e84717d76c284e55e3cd6c8c69c99007e9

oauth2_http/java/com/google/auth/oauth2/S2A.java
rmehta19: Some are defined in [ComputeEngineCredentials](https://github.com/googleapis/google-auth-library-java/blob/main/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java#L110-L117).

oauth2_http/java/com/google/auth/oauth2/S2A.java
rmehta19: The behavior we would like is for the S2A config to be queried once from the MDS, all subsequent requests from the application for the S2A config should return the cached config. 
Re: Could a valid config be a requirement to instantiate this class
> I think we can do this and perhaps simplify the code. I moved the call to get the config to the constructor.

https://github.com/googleapis/google-auth-library-java/pull/1400/commits/32caef5066f64f69ede2f8e4f3592fc53286c8d8

oauth2_http/java/com/google/auth/oauth2/S2A.java
rmehta19: Done in https://github.com/googleapis/google-auth-library-java/pull/1400/commits/b82790ac44f2d45d2232cd362656dc112fe6390b

oauth2_http/java/com/google/auth/oauth2/S2A.java
rmehta19: Done in https://github.com/googleapis/google-auth-library-java/pull/1400/commits/05aa9cc3d33b57e2d06541f51011c6b850286719

oauth2_http/java/com/google/auth/oauth2/S2A.java
rmehta19: Done in https://github.com/googleapis/google-auth-library-java/pull/1400/commits/05aa9cc3d33b57e2d06541f51011c6b850286719

oauth2_http/java/com/google/auth/oauth2/S2A.java
rmehta19: If the plaintext and mTLS S2A addresses are able to be parsed from the MDS response, it is a valid config. The MDS MTLS Autoconfig endpoint always returns the address that S2A is running at, if available. It is the caller's responsibility to check if the address is empty before passing it on the S2A Client to create ChannelCredentials ([example](https://github.com/grpc/grpc-java/blob/master/s2a/src/test/java/io/grpc/s2a/handshaker/IntegrationTest.java#L197)). This is similar to what we do in Go: https://github.com/googleapis/google-api-go-client/blob/main/internal/s2a.go#L22-L31

oauth2_http/java/com/google/auth/oauth2/S2A.java
```diff
@@ -0,0 +1,107 @@
...
+              ServiceLoader.load(HttpTransportFactory.class), OAuth2Utils.HTTP_TRANSPORT_FACTORY);
+    }  
+
+    for (int i = 0; i < MAX_MDS_PING_TRIES; i++) {
+      try {
+        HttpRequest request =
+            transportFactory.create().createRequestFactory().buildGetRequest(genericUrl);
+        request.setParser(parser);
+        request.getHeaders().set(METADATA_FLAVOR, GOOGLE);
+        request.setThrowExceptionOnExecuteError(false);
```
clundin25, line 77: Can the request be re-used across loops?

Not familiar with Java, but I imagine it's re-usable / expensive to create on each loop.

oauth2_http/java/com/google/auth/oauth2/S2A.java
```diff
@@ -0,0 +1,107 @@
...
+  /**
+   * Queries the MDS mTLS Autoconfiguration endpoint and returns the {@link S2AConfig}.
+   *
+   * <p>Returns {@link S2AConfig} with empty addresses on error.
+   *
+   * @return the {@link S2AConfig}.
+   */
+  private S2AConfig getS2AConfigFromMDS() {
+    String plaintextS2AAddress = "";
+    String mtlsS2AAddress = "";
```
clundin25, line 60: nit: Why declare these at the top? Is it a code style choice?

oauth2_http/java/com/google/auth/oauth2/S2A.java
```diff
@@ -0,0 +1,107 @@
...
+        mtlsS2AAddress = OAuth2Utils.validateString(responseData, "mtls_address", PARSE_ERROR_S2A);
+      } catch (IOException e) {
+        continue;
+      }
+      return S2AConfig.createBuilder()
+          .setPlaintextAddress(plaintextS2AAddress)
+          .setMtlsAddress(mtlsS2AAddress)
+          .build();
+    }
+    return S2AConfig.createBuilder().build();
```
clundin25, line 100: So consumers will have to check if the config contains empty strings? Could you use a result type or other form of indication that you were unable to build a valid config?

oauth2_http/java/com/google/auth/oauth2/S2A.java
rmehta19: Done in https://github.com/googleapis/google-auth-library-java/pull/1400/commits/99b7257802dbef5512c43a5c19b80d8efa805566

oauth2_http/java/com/google/auth/oauth2/S2A.java
rmehta19: This was done to keep consistency with Go implementation: https://github.com/googleapis/google-api-go-client/blob/main/internal/s2a.go#L22

The rationale is that we do not want to throw an error from this API that the client has to handle. If any problem occurs querying the endpoint for the address, we return an empty address. The client can pass that empty address to the S2A Client libraries, which will throw an error (e.g. something like "Error: Empty S2A Address.").

oauth2_http/java/com/google/auth/oauth2/S2AConfig.java
```diff
@@ -0,0 +1,58 @@
...
+    private String mtlsAddress;
+
+    Builder() {
+      plaintextAddress = "";
+      mtlsAddress = "";
+    }
+
+    @CanIgnoreReturnValue
+    public Builder setPlaintextAddress(String plaintextAddress) {
+      this.plaintextAddress = plaintextAddress;
```
zhumin8, line 39: Does these addresses need any validation or format check when setting with builder?

zhumin8 (commented):
Looks reasonable on first pass. I will take another look after reading a bit more on the context.

oauth2_http/java/com/google/auth/oauth2/S2A.java
```diff
@@ -0,0 +1,110 @@
...
+      request =
+              transportFactory.create().createRequestFactory().buildGetRequest(genericUrl);
+      request.setParser(parser);
+      request.getHeaders().set(METADATA_FLAVOR, GOOGLE);
+      request.setThrowExceptionOnExecuteError(false);
+    } catch (IOException e) {
+      return S2AConfig.createBuilder().build();
+    }
+
+    for (int i = 0; i < MAX_MDS_PING_TRIES; i++) {
```
zhumin8, line 79: I am not super familiar with google-http-client, but can this retry logic be handled by maybe `HttpRequestRetryHandler`?

`ExponentialBackOff` used in retrys [here](https://github.com/googleapis/google-auth-library-java/blob/d4a3763799896a35e27b9f5c178f77364e41668d/oauth2_http/java/com/google/auth/oauth2/IamUtils.java#L130C24-L143) is not thread-safe.
cc.@lqiu96 who might be more familiar

oauth2_http/java/com/google/auth/oauth2/S2A.java
```diff
@@ -0,0 +1,110 @@
...
+ * <p>mTLS configuration is queried from the MDS MTLS Autoconfiguration endpoint.
+ */
+@ThreadSafe
+public final class S2A {
+  public static final String S2A_CONFIG_ENDPOINT_POSTFIX =
+      "/computeMetadata/v1/instance/platform-security/auto-mtls-configuration";
+
+  public static final String METADATA_FLAVOR = "Metadata-Flavor";
+  public static final String GOOGLE = "Google";
+  private static final int MAX_MDS_PING_TRIES = 3;
```
lqiu96, line 27: Can we use the default retry value here: https://github.com/googleapis/google-auth-library-java/blob/d4a3763799896a35e27b9f5c178f77364e41668d/oauth2_http/java/com/google/auth/oauth2/OAuth2Utils.java#L110?

oauth2_http/java/com/google/auth/oauth2/S2A.java
```diff
@@ -0,0 +1,110 @@
...
+ *
+ * <p>mTLS configuration is queried from the MDS MTLS Autoconfiguration endpoint.
+ */
+@ThreadSafe
+public final class S2A {
+  public static final String S2A_CONFIG_ENDPOINT_POSTFIX =
+      "/computeMetadata/v1/instance/platform-security/auto-mtls-configuration";
+
+  public static final String METADATA_FLAVOR = "Metadata-Flavor";
+  public static final String GOOGLE = "Google";
```
lqiu96, line 26: Do these constants need to be public? Can they be package-private? Or are they intended to be used by users?

oauth2_http/java/com/google/auth/oauth2/S2A.java
```diff
@@ -0,0 +1,110 @@
...
+  private S2AConfig config;
+
+  private transient HttpTransportFactory transportFactory;
+
+  public S2A() {}
+
+  public void setHttpTransportFactory(HttpTransportFactory tf) {
+    this.transportFactory = tf;
+    this.config = getS2AConfigFromMDS();
+  }
```
lqiu96, line 39: Is there a reason why the config needs to be computed when the HttpTransportFactory is set? To me, it seems a bit odd that setting a value in the setter ends up making a network call. 

What if a user (for whatever reason) ends up calling `setHttpTransportFactory` multiple times?

oauth2_http/java/com/google/auth/oauth2/S2A.java
```diff
@@ -0,0 +1,110 @@
...
+          .setMtlsAddress(mtlsS2AAddress)
+          .build();
+    }
+    return S2AConfig.createBuilder().build();
+  }
+
+  /** @return MDS mTLS autoconfig endpoint. */
+  private String getMdsMtlsEndpoint() {
+    return ComputeEngineCredentials.getMetadataServerUrl() + S2A_CONFIG_ENDPOINT_POSTFIX;
+  }
```
lqiu96, line 109: Can this just be a static constant?

oauth2_http/java/com/google/auth/oauth2/S2AConfig.java
```diff
@@ -0,0 +1,58 @@
+package com.google.auth.oauth2;
+
+import com.google.errorprone.annotations.CanIgnoreReturnValue;
+
+/** Holds an mTLS configuration (consists of address of S2A) retrieved from the Metadata Server. */
+public final class S2AConfig {
```
lqiu96, line 6: Does this need to be public?

oauth2_http/java/com/google/auth/oauth2/S2A.java
```diff
@@ -0,0 +1,110 @@
+package com.google.auth.oauth2;
```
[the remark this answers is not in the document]
rmehta19: Done in https://github.com/googleapis/google-auth-library-java/pull/1400/commits/c89b56ca809c02377a4ddd895416a6a1f240f49e

oauth2_http/java/com/google/auth/oauth2/S2AConfig.java
```diff
@@ -0,0 +1,58 @@
+package com.google.auth.oauth2;
```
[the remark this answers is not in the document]
rmehta19: Done in https://github.com/googleapis/google-auth-library-java/commit/c89b56ca809c02377a4ddd895416a6a1f240f49e

oauth2_http/java/com/google/auth/oauth2/S2AConfig.java
rmehta19: No validation/format check is necessary here, because we own the MDS endpoint that is being queried to get the address, and it is up to the client which consumes the address ([S2A client](https://github.com/grpc/grpc-java/tree/master/s2a)) to return error if there is a problem connecting to the S2A at that address.

oauth2_http/java/com/google/auth/oauth2/S2A.java
```diff
@@ -0,0 +1,110 @@
...
+  public static final String METADATA_FLAVOR = "Metadata-Flavor";
+  public static final String GOOGLE = "Google";
+  private static final int MAX_MDS_PING_TRIES = 3;
+  private static final String PARSE_ERROR_S2A = "Error parsing S2A Config from MDS JSON response.";
+
+  private S2AConfig config;
+
+  private transient HttpTransportFactory transportFactory;
+
+  public S2A() {}
```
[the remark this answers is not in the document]
rmehta19: Yes, that is why the constructor was included, but thank you for pointing that out Min, I've removed it in https://github.com/googleapis/google-auth-library-java/pull/1400/commits/544d9d1a05d5355b066e0fb1fc25b823e537b226

oauth2_http/java/com/google/auth/oauth2/S2A.java
rmehta19: Done in https://github.com/googleapis/google-auth-library-java/pull/1400/commits/c3ede1d32b1b4c3cd4c336f9c88a0eba399634e5

oauth2_http/java/com/google/auth/oauth2/S2A.java
rmehta19: They aren't intended for users, but they are used in the tests (see: MockMetadataServerTransport.java). In this case would you prefer we make these private and add static getters so tests can use them?

oauth2_http/java/com/google/auth/oauth2/S2A.java
rmehta19: Thanks for pointing this out Lawrence! It is wrong to be setting the config here. Another case where this would be a bug is if the caller doesn't call `setHttpTransportFactory` (if they are using the default instance), in which case the config would never get set. 

In https://github.com/googleapis/google-auth-library-java/pull/1400/commits/8238d50e14e86a529ce0c421079d74b84725a57a I have changed it to set the config in the constructor, and made the transportFactory an optional parameter, rather than having a separate setter. I think this is fine because then the caller can also be aware (without checking the setters) that they are not setting transport and will be using default transport. This also avoids having code like:
```
public String getMtlsS2AAddress() {
    if (config == null) {
      this.config = getS2AConfigFromMDS();
    }
    return config.getMtlsAddress();
  }
public String getPlaintextS2AAddress() {
    if (config == null) {
      this.config = getS2AConfigFromMDS();
    }
    return config.getPlaintextAddress();
  }
```

which introduces a race condition on setting the config.

oauth2_http/java/com/google/auth/oauth2/S2A.java
rmehta19: Done in https://github.com/googleapis/google-auth-library-java/pull/1400/commits/36ab0a99f69a6ad94758b0bde9ab9ea3d1123d57

oauth2_http/java/com/google/auth/oauth2/S2AConfig.java
rmehta19: It does not, thanks for pointing that out. Done in https://github.com/googleapis/google-auth-library-java/pull/1400/commits/36a0ac7cad28656f791a057bdcac70bb6cfd1e27

oauth2_http/java/com/google/auth/oauth2/S2A.java
rmehta19: Thanks for pointing this out Min! A few followup questions: 
- I'm not using `ExponentialBackOff` in the retries, is their still a thread-safety problem? The retry logic here is mirrored from https://github.com/googleapis/google-auth-library-java/blob/dde6876154d6e7ff43dbbf7b75c5205eadd679a5/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java#L435
- From quickly looking at https://cloud.google.com/java/docs/reference/google-http-client/latest/com.google.api.client.http, I don't see anything similar to `HttpRequestRetryHandler`. Should we instead use the `org.apache.http.impl.client` implementation so we can use `org.apache.http.impl.client.DefaultHttpRequestRetryHandler` or something similar?

rmehta19:
Thanks for the review @zhumin8 and @lqiu96 ! I've addressed all your comments, please leave any followup comments!

oauth2_http/java/com/google/auth/oauth2/S2A.java
lqiu96: Can these package-private? I think it should be accessible

oauth2_http/java/com/google/auth/oauth2/S2A.java
```diff
@@ -0,0 +1,133 @@
...
+  private static final String MDS_MTLS_ENDPOINT = ComputeEngineCredentials.getMetadataServerUrl() + S2A_CONFIG_ENDPOINT_POSTFIX;
+
+  private S2AConfig config;
+
+  private transient HttpTransportFactory transportFactory;
+
+  public S2A(Optional<HttpTransportFactory> transportFactory) {
+    this.transportFactory = transportFactory.get();
+    this.config = getS2AConfigFromMDS();
+  }
```
lqiu96, line 68: Can the constructor be done similar to ServiceAccount (https://github.com/googleapis/google-auth-library-java/blob/2761a39633a73bb93d36c097a854346091ee70fa/oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java#L124)?

The constructor takes in a builder. We can make the constructor package-private and we wouldn't need to add new method args to the constructor in the future.

oauth2_http/java/com/google/auth/oauth2/S2A.java
```diff
@@ -0,0 +1,133 @@
...
+  public String getPlaintextS2AAddress() {
+    return config.getPlaintextAddress();
+  }
+
+  /**
+   * Queries the MDS mTLS Autoconfiguration endpoint and returns the {@link S2AConfig}.
+   *
+   * <p>Returns {@link S2AConfig} with empty addresses on error.
+   *
+   * @return the {@link S2AConfig}.
```
lqiu96, line 85: Can you add a more descriptive javadocs that details when empty addresses occur and what users are expected to do with it? What if only plaintextAddress is empty or mtsAddress is empty? Is that different when both are empty?

oauth2_http/java/com/google/auth/oauth2/S2A.java
```diff
@@ -0,0 +1,133 @@
...
+        if (content == null) {
+          continue;
+        }
+        GenericData responseData = response.parseAs(GenericData.class);
+        plaintextS2AAddress =
+            OAuth2Utils.validateString(responseData, "plaintext_address", PARSE_ERROR_S2A);
+        mtlsS2AAddress = OAuth2Utils.validateString(responseData, "mtls_address", PARSE_ERROR_S2A);
+      } catch (IOException e) {
+        continue;
+      }
```
lqiu96, line 125: Can you add a comment in the code explaining why IOExceptions are caught here? Oauth2Utils seem to throw when the response data is null or not a string. Why is it fine to ignore those (i.e. s2a server is expected to always...)

oauth2_http/java/com/google/auth/oauth2/S2A.java
rmehta19: Thanks for pointing that out! Done in https://github.com/googleapis/google-auth-library-java/pull/1400/commits/ae545c82e217353bf086a856674ac7ae54e59e35

oauth2_http/java/com/google/auth/oauth2/S2A.java
rmehta19: Done in https://github.com/googleapis/google-auth-library-java/pull/1400/commits/47b3f2e4205fc69ab40759214a259aca7215c9db

oauth2_http/java/com/google/auth/oauth2/S2A.java
rmehta19: Done in https://github.com/googleapis/google-auth-library-java/pull/1400/commits/fb577a1b6f3ba7b34f6fa0318309deb7efb76db8

Also, what if only plaintextAddress is empty or mtsAddress is empty? Is that different when both are empty?
> The behavior we will add to the Client libraries (specifically, we are adding logic to decide to create channel using S2A in GAX layer) is: 
> try to fetch mtlsS2AAdress first (call `getMtlsS2AAddress()`. if that is empty or if there is a failure when trying to load the mTLS-to-MDS credentials, then fallback to fetch plaintextS2AAddress (call `getPlaintextS2AAddress()`) . If the plaintextS2AAdress is also empty, that means there is some error when trying to talk the MDS, or S2A is not available in that environment, thus we should not try and use S2A.

oauth2_http/java/com/google/auth/oauth2/S2A.java
```diff
@@ -143,6 +143,11 @@ private S2AConfig getS2AConfigFromMDS() {
       try {
         HttpResponse response = request.execute();
         if (!response.isSuccessStatusCode()) {
+          int statusCode = response.getStatusCode();
```
rmehta19, line 146: fyi: @lqiu96 @zhumin8 , after chatting with @xmenxk (authored the Go implementation: https://github.com/googleapis/google-cloud-go/blob/main/auth/internal/transport/s2a.go), he mentioned that we shouldn't retry if the endpoint doesn't exist. This logic exists in Go impl, so adding it here as well.

oauth2_http/java/com/google/auth/oauth2/S2A.java
rmehta19: Done in https://github.com/googleapis/google-auth-library-java/pull/1400/commits/0bbd3207a5cfe3f64e84eba510a50dc1fe755685

rmehta19:
Thanks for the second pass through @lqiu96! I've addressed the comments, please let me know of any followups / additional comments.

oauth2_http/java/com/google/auth/oauth2/S2A.java
zhumin8: I only mentioned `ExponentialBackOff` because that is used in one of the recent changes related to retry (https://github.com/googleapis/google-auth-library-java/pull/1452/files)
I suppose you can do similar, by using HttpRequest's built-in retry handlers and set  `request.setNumberOfRetries(OAuth2Utils.DEFAULT_NUMBER_OF_RETRIES);`?  If you do not need `ExponentialBackOff`, maybe something simple e.g.

```
      // Set the number of retries
      request.setNumberOfRetries(OAuth2Utils.DEFAULT_NUMBER_OF_RETRIES); 
      // Retry on specific status codes (you might want to adjust these)
      request.setUnsuccessfulResponseHandler(
          new HttpUnsuccessfulResponseHandler() {
            @Override
            public boolean handleResponse(
                HttpRequest request, HttpResponse response, boolean supportsRetry)
                throws IOException   {
              return   RETRYABLE_STATUS_CODES.contains(response.getStatusCode());
            }
          });
```
About thread-safety, noticed HttpRequest is also marked as not safe [here](https://github.com/googleapis/google-http-java-client/blob/4830ad788a62fe9cd4f64873b771e6ef8ef92193/google-http-client/src/main/java/com/google/api/client/http/HttpRequest.java#L44), likely because its member variable hold state information . But I don't see it a concern in this implementation, as `executeAsync()` is not used and `getS2AConfigFromMDS()` creates and executes a request each time.

oauth2_http/java/com/google/auth/oauth2/S2A.java
zhumin8: I lean toward switching to HttpRequest's built-in retry handlers, logic like to customize the retry behavior based on return code can go into `HttpUnsuccessfulResponseHandler`. See other comment for details.

oauth2_http/java/com/google/auth/oauth2/S2AConfig.java
zhumin8: Can you add this explanation as javadoc comment here or to `getS2AConfigFromMDS()` for future references?

oauth2_http/java/com/google/auth/oauth2/S2A.java
lqiu96: > we shouldn't retry if the endpoint doesn't exist

This means that we don't retry on 404 status code, right?

oauth2_http/java/com/google/auth/oauth2/S2A.java
lqiu96: Yeah I agree and I would prefer if the loops could be modified to use the unsuccessfulResponseHandler and exponential backoff (configured with a list of status codes that can be retried on).

oauth2_http/java/com/google/auth/oauth2/S2A.java
lqiu96:
```        
plaintextS2AAddress =
            OAuth2Utils.validateString(responseData, "plaintext_address", PARSE_ERROR_S2A);
        mtlsS2AAddress = OAuth2Utils.validateString(responseData, "mtls_address", PARSE_ERROR_S2A);
      } catch (IOException e) {
...
      }
```

If the response data is something like:
```
{
   plaintext_address: null,
   mtls_address: "localhost",
}
```

then I believe an IOException would be thrown and mtls_address wouldn't be set even though it's valid. Just want confirm the behavior.

rmehta19: Yes this behavior could happen where an IOException is thrown and mtls_address is not set even though it is able to be parsed (I don't think it's too likely since we own the MDS mTLS autoconfiguration endpoint, and the json format is tested). 

In https://github.com/googleapis/google-auth-library-java/pull/1400/commits/0e6f5cef7a289b2a875a4788a4eca2d17545c944 I've adjusted the code so that it catches and ignores any exception thrown when parsing the addresses. So in this case, if for whatever reason, parsing the plaintext_address address fails, we still try and parse the mtls_address address. I chatted with @xmenxk, and we agree that in this case we shouldn't really retry since if we get a successful response from the MDS, it is unlikely to change if we query it again.

oauth2_http/java/com/google/auth/oauth2/S2A.java
rmehta19: I am leaning toward switching to HttpRequest's built-in retry handlers, logic like to customize the retry behavior based on return code can go into HttpUnsuccessfulResponseHandler. See other comment for details.
> SG! I'll address your other comment.

This means that we don't retry only on 404 status code, right?
> Yes, we only want to retry on 404.

oauth2_http/java/com/google/auth/oauth2/S2A.java
```diff
@@ -0,0 +1,133 @@
+/*
+ * Copyright 2016, Google Inc. All rights reserved.
```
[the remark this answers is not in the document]
rmehta19: Done in https://github.com/googleapis/google-auth-library-java/pull/1400/commits/e786886b996e5530dc5959b1d4a42b0023b27aa3

oauth2_http/java/com/google/auth/oauth2/S2A.java
rmehta19: Thanks for the detail @zhumin8 ! In https://github.com/googleapis/google-auth-library-java/pull/1400/commits/12b248d8245162d63a1fe4863921b4cb595afa48 I have modified to set the number of retries on the request and use `ExponentialBackOff` and `HttpUnsuccessfulResponseHandler`. We only want to be retrying on 5xx error codes, so I have specified them as retryable error codes.

oauth2_http/java/com/google/auth/oauth2/S2AConfig.java
rmehta19: Done in https://github.com/googleapis/google-auth-library-java/pull/1400/commits/4d056380457e33f17fb758aec2b51a57849e751d

rmehta19:
Thanks for another review @zhumin8 and @lqiu96 ! I have addressed this round of comments, please let me know if there is anything else

oauth2_http/java/com/google/auth/oauth2/S2A.java
```diff
@@ -0,0 +1,196 @@
...
+
+      // Retry on 5xx status codes.
+      request.setUnsuccessfulResponseHandler(
+          new HttpBackOffUnsuccessfulResponseHandler(backoff)
+              .setBackOffRequired(
+                  response -> RETRYABLE_STATUS_CODES.contains(response.getStatusCode())));
+      request.setIOExceptionHandler(new HttpBackOffIOExceptionHandler(backoff));
+    } catch (IOException e) {
+      return S2AConfig.createBuilder().build();
+    }
```
lqiu96, line 163: `S2AConfig.createBuilder().build()` sets plainTextAddress and mtlsS2AAddress to `""` by default. Can we move this code in this try block into the one below (merging them)?

oauth2_http/java/com/google/auth/oauth2/S2A.java
```diff
@@ -0,0 +1,196 @@
...
+
+    String plaintextS2AAddress = "";
+    String mtlsS2AAddress = "";
+    try {
+      HttpResponse response = request.execute();
+      InputStream content = response.getContent();
+      if (content == null) {
+        return S2AConfig.createBuilder().build();
+      }
```
lqiu96, line 172: Why is this check needed? I don't think there is any logic below regarding content.

oauth2_http/javatests/com/google/auth/oauth2/MockMetadataServerTransport.java
```diff
@@ -62,6 +62,15 @@ public class MockMetadataServerTransport extends MockHttpTransport {
 
   private byte[] signature;
 
+  private String plaintextS2AAddressJsonKey;
+
+  private String plaintextS2AAddress;
+
+  private String mtlsS2AAddressJsonKey;
```
lqiu96, line 69: Do the tests need to make the two JsonKeys configurable? Can they just be referenced from the constants from the S2A class?

oauth2_http/java/com/google/auth/oauth2/S2A.java
rmehta19: This is true that there is any logic regarding content after this, I included this logic when writing this from looking at: https://github.com/googleapis/google-auth-library-java/blob/2761a39633a73bb93d36c097a854346091ee70fa/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java#L292, to avoid a NullPointerException on the `parseAs` call.

oauth2_http/javatests/com/google/auth/oauth2/MockMetadataServerTransport.java
lqiu96: Ah nvm, I jumped the gun. I see the tests below where the data is put in the wrong key

oauth2_http/javatests/com/google/auth/oauth2/MockMetadataServerTransport.java
```diff
@@ -272,6 +303,39 @@ public LowLevelHttpResponse execute() throws IOException {
...
+
+        String metadataRequestHeader = getFirstHeaderValue(S2A.METADATA_FLAVOR);
+        if (!S2A.GOOGLE.equals(metadataRequestHeader)) {
+          throw new IOException("Metadata request header not found");
+        }
+
+        // Create the JSON response
+        GenericJson content = new GenericJson();
+        content.setFactory(OAuth2Utils.JSON_FACTORY);
+        if (requestStatusCode < 400) {
```
lqiu96, line 319: nit: Can this be 200? Or are there other 2xx/ 3xx codes?

lqiu96 (commented):
Thanks for working with us to address the comments. I think this mostly looks good to me given the context. 

Can you also do two additional things:
1. Run `mvn fmt:format` on this PR?
2. Update the title to be `feat:  {title}`? Something that describes the functionality `feat: Support querying S2A Addresses from MDS` or whatever you think aptly describes it.

oauth2_http/java/com/google/auth/oauth2/S2A.java
rmehta19: Done in https://github.com/googleapis/google-auth-library-java/pull/1400/commits/ed681f58e377caef73ec3fc7d6696704d4bbb62f

oauth2_http/javatests/com/google/auth/oauth2/MockMetadataServerTransport.java
rmehta19: In `S2ATest`, we reference the constants in the `S2A` class. The reason I made the keys configurable was to test the logic added as part of this comment thread: https://github.com/googleapis/google-auth-library-java/pull/1400#discussion_r1815800047. 

For example, in `getS2AAddress_queryEndpointResponseInvalidPlaintextJsonKey_plaintextEmptyAddress`, I inject an invalid json key via `plaintextS2AAddressJsonKey`, to confirm that the `mtls_address` is still parsed, even though there's a problem parsing the `plaintext_address`.

oauth2_http/javatests/com/google/auth/oauth2/MockMetadataServerTransport.java
rmehta19: Yeah it can just be 200. We expect only 200 on successful responses. Done in https://github.com/googleapis/google-auth-library-java/pull/1400/commits/20825f70a2e6e21a03a4480267f62f706f7ea35f

oauth2_http/javatests/com/google/auth/oauth2/MockMetadataServerTransport.java
```diff
@@ -62,6 +62,15 @@ public class MockMetadataServerTransport extends MockHttpTransport {
 
   private byte[] signature;
 
+  private String plaintextS2AAddressJsonKey;
+
+  private String plaintextS2AAddress;
+
+  private String mtlsS2AAddressJsonKey;
+
+  private String mtlsS2AAddress;
```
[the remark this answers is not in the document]
rmehta19: Done in https://github.com/googleapis/google-auth-library-java/pull/1400/commits/594df7b06c0f18f407f03dc2f307f204e7b00a00

rmehta19:
@lqiu96 , I've addressed your latest round of comments -- thanks for the feedback!

Run mvn fmt:format on this PR?
> Done in https://github.com/googleapis/google-auth-library-java/pull/1400/commits/16fd964472cde6cde3cee7df8a22afab54eb9455

Update the title to be feat:  {title}? Something that describes the functionality feat: Support querying S2A Addresses from MDS or whatever you think aptly describes it.
> Done.

oauth2_http/java/com/google/auth/oauth2/S2A.java
lqiu96: Gotcha, thanks. Shouldn't hurt to have additional checks.

lqiu96 (commented):
LGTM. Added a few nits. Please do let @zhumin8 to review and give an approval as well

oauth2_http/javatests/com/google/auth/oauth2/S2ATest.java
```diff
@@ -0,0 +1,139 @@
...
+
+  @Test
+  public void getS2AAddress_validAddress() {
+    MockMetadataServerTransportFactory transportFactory = new MockMetadataServerTransportFactory();
+    transportFactory.transport.setS2AContentMap(
+        "plaintextS2AAddressJsonKey", S2A.S2A_PLAINTEXT_ADDRESS_JSON_KEY);
+    transportFactory.transport.setS2AContentMap("plaintextS2AAddress", S2A_PLAINTEXT_ADDRESS);
+    transportFactory.transport.setS2AContentMap(
+        "mtlsS2AAddressJsonKey", S2A.S2A_MTLS_ADDRESS_JSON_KEY);
+    transportFactory.transport.setS2AContentMap("mtlsS2AAddress", S2A_MTLS_ADDRESS);
```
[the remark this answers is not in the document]
rmehta19: This is more succinct -- thanks. Done in https://github.com/googleapis/google-auth-library-java/pull/1400/commits/1e6c05872c3512c7fd237936da40405d20854708

rmehta19:
> LGTM. Added a few nits. Please do let @zhumin8 to review and give an approval as well

Thanks for the review @lqiu96 ! @zhumin8 , please let me know if there is anything else to address, thanks!

oauth2_http/java/com/google/auth/oauth2/S2AConfig.java
```diff
@@ -0,0 +1,96 @@
...
+  private final String plaintextAddress;
+
+  // mtlsAddress is the mTLS address to reach the S2A.
+  private final String mtlsAddress;
+
+  public static Builder createBuilder() {
+    return new Builder();
+  }
+
+  public String getPlaintextAddress() {
```
westarle, line 47: I'd like Javadocs for all these public things, pointing to public docs for MDS.

rmehta19: We don't have this documented in public MDS docs (e.g. https://cloud.google.com/compute/docs/metadata/predefined-metadata-keys). We do have an AIP: https://google.aip.dev/auth/4115 which discusses this autconfig endpoint and how it fits in the mTLS via S2A + bound tokens story. WDYT about https://github.com/googleapis/google-auth-library-java/pull/1400/commits/934679c76ab3ad0243a07e656afe45c3b85a79dc?

oauth2_http/java/com/google/auth/oauth2/S2A.java
```diff
@@ -0,0 +1,194 @@
...
+
+      // Retry on 5xx status codes.
+      request.setUnsuccessfulResponseHandler(
+          new HttpBackOffUnsuccessfulResponseHandler(backoff)
+              .setBackOffRequired(
+                  response -> RETRYABLE_STATUS_CODES.contains(response.getStatusCode())));
+      request.setIOExceptionHandler(new HttpBackOffIOExceptionHandler(backoff));
+
+      HttpResponse response = request.execute();
+      InputStream content = response.getContent();
```
zhumin8, line 166: maybe do try/catch block around these to be more specific about the exceptions catching? Nested try/catch blocks are a bit hard to read.

zhumin8 (commented):
Added two minor comments below, otherwise LGTM.

oauth2_http/java/com/google/auth/oauth2/S2A.java
lqiu96: Ah, I think I was the one that suggested this, but am open to having the try-catch blocks be more specific or a different implementation.

I suggested just have one try-catch block since this code doesn't catch or return the error message back to the user. It just catches any IO error and returns an empty S2AConfig back. Unfortunately, this doesn't provide any useful feedback to the user. Given that this is how it behaves already in Go, I thought it would just be simpler to have a catch-all (one big try block) for this logic. 

Ideally, we would have more specific try-catch blocks which could tell the user exactly what went wrong (i.e. failed to build the request because x, or the request failed because of y).

I believe there are two spots where we potentially would need to add try-catches:
1. buildGetRequest()
2. request.execute()

oauth2_http/java/com/google/auth/oauth2/S2A.java
```diff
@@ -0,0 +1,194 @@
...
+      if (content == null) {
+        return S2AConfig.createBuilder().build();
+      }
+      GenericData responseData = response.parseAs(GenericData.class);
+      try {
+        plaintextS2AAddress =
+            OAuth2Utils.validateString(
+                responseData, S2A_PLAINTEXT_ADDRESS_JSON_KEY, PARSE_ERROR_S2A);
+      } catch (IOException ignore) {
+      }
```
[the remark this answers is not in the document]
rmehta19: I added a comment in https://github.com/googleapis/google-auth-library-java/pull/1400/commits/6644d50203d25affc95bca8d447bed04bfe0779f. In general, if there is any error in this function, we ignore, and populate empty addresses in the `S2AConfig`.

oauth2_http/java/com/google/auth/oauth2/S2A.java
rmehta19: There are a few places in this block that throw an `IOException` (the `parseAs` and `buildGetRequest` also throw `IOException`),  so I removed the current nested try/catch block structure and added two separate try/catch blocks around the lines that throw an exception. Because we are getting rid of the big block, I also moved around variable definitions closer to where they are used. 

https://github.com/googleapis/google-auth-library-java/pull/1400/commits/6644d50203d25affc95bca8d447bed04bfe0779f

rmehta19: @lqiu96 , I think my changes and your comment happened at same time. Reading through your comment, the changes I just push match your suggestion. I removed the big try/catch and just did 2 smaller ones around building the request and executing it. 

@zhumin8 , @lqiu96: I am fine with either option -- let me know if you prefer reverting to the 1 big/try catch block, or if you prefer the current state which has no nesting, but a few smaller try/catch blocks

lqiu96: I am fine with splitting it up to reduce nesting.

rmehta19:
@zhumin8 , @lqiu96, please leave any additional feedback, and let me know if anything else is blocking submission, thanks!

oauth2_http/java/com/google/auth/oauth2/S2AConfig.java
westarle: Please include the documentation for each public method here.  The AIP isn't a great landing place -- can you please look into improving the public MDS docs at least with a bug to the metadata service team?

rmehta19: Done in https://github.com/googleapis/google-auth-library-java/pull/1400/commits/8ca8d693514c60756bef34904e45d064d1208f5e

The AIP isn't a great landing place -- can you please look into improving the public MDS docs at least with a bug to the metadata service team?
> Agreed that public MDS docs will need to be updated to document this auto-mtls-configuration MDS endpoint that is being queried to learn where S2A is running. The endpoint is currently still experimental. I created a bug to track this internally, will share it with you in our meeting later today.

rmehta19:
kokoro integration test is failing, but looks like an unrelated failure in PluggableAuthHandlerTest.java getExecutableResponse_oidcResponse

cc: @zhumin8 @lqiu96

lqiu96 (approved):
I think the test may be flaky. Can you re-run it to confirm?

PR #1421, opened 2024-06-07
feat: adding support for authentication type on UserAuthorizer

BigTailWolf:

The change basically following the logic that NodeJS change: https://github.com/googleapis/google-auth-library-nodejs/pull/1814

The key point is telling the client how are the `UserAuthorizer` going to provide auth with token URI.
Our current way is to have `client_secret` sending as part of the post url parameter. The STS endpoint won't allow that and they are not accepting `client_secret` field. Instead, the STS is using basic auth which takes a base64 encoding of `client_id:client_secret`.

Here the change is to provide a parameter to `UserAuthorizer` which auth from [#RFC](https://datatracker.ietf.org/doc/html/rfc7591) we are using and set the `POST` (Current way) as default.
Then in the implementation, when sending the token request, we apply a basic auth header if the authentication type is set to `BASIC`.

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/UserAuthorizer.java
+++ b/oauth2_http/java/com/google/auth/oauth2/UserAuthorizer.java
@@ -332,9 +332,8 @@ public class UserAuthorizer {
     tokenRequest.setParser(new JsonObjectParser(OAuth2Utils.JSON_FACTORY));
 
     if (this.clientAuthenticationType == ClientAuthenticationType.CLIENT_SECRET_BASIC) {
-      BaseEncoding base64 = BaseEncoding.base64();
       String encodedCredentials =
-          base64.encode((clientId.getClientId() + ":" + clientId.getClientSecret()).getBytes());
+          BaseEncoding.base64().encode((clientId.getClientId() + ":" + clientId.getClientSecret()).getBytes());
       tokenRequest.getHeaders().setAuthorization("Basic " + encodedCredentials);
     }
 
--- a/oauth2_http/java/com/google/auth/oauth2/UserAuthorizer.java
+++ b/oauth2_http/java/com/google/auth/oauth2/UserAuthorizer.java
@@ -43,7 +43,6 @@ import com.google.api.client.util.Joiner;
 import com.google.api.client.util.Preconditions;
 import com.google.auth.http.HttpTransportFactory;
 import com.google.common.collect.ImmutableList;
-import com.google.common.io.BaseEncoding;
 import com.google.errorprone.annotations.CanIgnoreReturnValue;
 import java.io.IOException;
 import java.net.URI;
@@ -96,7 +95,7 @@ public class UserAuthorizer {
    * @param tokenServerUri URI of the end point that provides tokens
    * @param userAuthUri URI of the Web UI for user consent
    * @param pkce PKCE implementation
-   * @param clientAuthenticationType ClientAuthentication type as defined in RFC 7591
+   * @param clientAuthentication ClientAuthentication type as defined in RFC 7591
    *     basic/post/none
    */
   private UserAuthorizer(
@@ -305,6 +304,11 @@ public class UserAuthorizer {
   /**
    * Returns a UserCredentials instance by exchanging an OAuth2 authorization code for tokens.
    *
+   * <p>The library is going to exchange an access token depends on the token_uri we provide. If
+   * will be either oauth2 or sts. The two endpoints accepting different way of basic auth. STS want
+   * client id and client secret being base64 encoded and put into header. OAuth2 endpoint want the
+   * client secret being sent as a field in post.
+   *
    * @param code Code returned from OAuth2 consent prompt.
    * @param baseUri The URI to resolve the OAuth2 callback URI relative to.
    * @param additionalParameters Additional parameters to be added to the post body of token
@@ -322,10 +326,7 @@ public class UserAuthorizer {
     tokenData.put("client_id", clientId.getClientId());
     tokenData.put("redirect_uri", resolvedCallbackUri);
     tokenData.put("grant_type", "authorization_code");
-    // If the token request is sent to STS. STS will fail by design if the `client_secret` in post
-    // body.
-    // Sending client_secret when the client authentication type is post is to maintain existing
-    // behavior for existing code.
+
     if (this.clientAuthenticationType == ClientAuthenticationType.CLIENT_SECRET_POST) {
       tokenData.put("client_secret", clientId.getClientSecret());
     }
@@ -347,7 +348,7 @@ public class UserAuthorizer {
     tokenRequest.setParser(new JsonObjectParser(OAuth2Utils.JSON_FACTORY));
 
     if (this.clientAuthenticationType == ClientAuthenticationType.CLIENT_SECRET_BASIC) {
-      tokenRequest.getHeaders().setAuthorization(getBasicAuthString(clientId));
+      tokenRequest.getHeaders().setAuthorization(OAuth2Utils.getBasicAuthString(clientId));
     }
 
     HttpResponse tokenResponse = tokenRequest.execute();
@@ -506,19 +507,6 @@ public class UserAuthorizer {
       UserCredentials userCredentials = (UserCredentials) credentials;
       storeCredentials(userId, userCredentials);
     }
-  }
-
-  /**
-   * A helper function to get a base64 encoded basic auth.
-   *
-   * @param clientId A ClientId object which contains a string value and a secret
-   * @return The basic auth header
-   */
-  public static String getBasicAuthString(ClientId clientId) {
-    String encodedCredentials =
-        BaseEncoding.base64()
-            .encode((clientId.getClientId() + ":" + clientId.getClientSecret()).getBytes());
-    return "Basic " + encodedCredentials;
   }
 
   public static Builder newBuilder() {
--- a/oauth2_http/java/com/google/auth/oauth2/UserAuthorizer.java
+++ b/oauth2_http/java/com/google/auth/oauth2/UserAuthorizer.java
@@ -322,6 +322,10 @@ public class UserAuthorizer {
     tokenData.put("client_id", clientId.getClientId());
     tokenData.put("redirect_uri", resolvedCallbackUri);
     tokenData.put("grant_type", "authorization_code");
+    // If the token request is sent to STS. STS will fail by design if the `client_secret` in post
+    // body.
+    // Sending client_secret when the client authentication type is post is to maintain existing
+    // behavior for existing code.
     if (this.clientAuthenticationType == ClientAuthenticationType.CLIENT_SECRET_POST) {
       tokenData.put("client_secret", clientId.getClientSecret());
     }
@@ -343,10 +347,7 @@ public class UserAuthorizer {
     tokenRequest.setParser(new JsonObjectParser(OAuth2Utils.JSON_FACTORY));
 
     if (this.clientAuthenticationType == ClientAuthenticationType.CLIENT_SECRET_BASIC) {
-      String encodedCredentials =
-          BaseEncoding.base64()
-              .encode((clientId.getClientId() + ":" + clientId.getClientSecret()).getBytes());
-      tokenRequest.getHeaders().setAuthorization("Basic " + encodedCredentials);
+      tokenRequest.getHeaders().setAuthorization(getBasicAuthString(clientId));
     }
 
     HttpResponse tokenResponse = tokenRequest.execute();
@@ -505,6 +506,19 @@ public class UserAuthorizer {
       UserCredentials userCredentials = (UserCredentials) credentials;
       storeCredentials(userId, userCredentials);
     }
+  }
+
+  /**
+   * A helper function to get a base64 encoded basic auth.
+   *
+   * @param clientId A ClientId object which contains a string value and a secret
+   * @return The basic auth header
+   */
+  public static String getBasicAuthString(ClientId clientId) {
+    String encodedCredentials =
+        BaseEncoding.base64()
+            .encode((clientId.getClientId() + ":" + clientId.getClientSecret()).getBytes());
+    return "Basic " + encodedCredentials;
   }
 
   public static Builder newBuilder() {
--- a/oauth2_http/java/com/google/auth/oauth2/UserAuthorizer.java
+++ b/oauth2_http/java/com/google/auth/oauth2/UserAuthorizer.java
@@ -43,6 +43,7 @@ import com.google.api.client.util.Joiner;
 import com.google.api.client.util.Preconditions;
 import com.google.auth.http.HttpTransportFactory;
 import com.google.common.collect.ImmutableList;
+import com.google.common.io.BaseEncoding;
 import com.google.errorprone.annotations.CanIgnoreReturnValue;
 import java.io.IOException;
 import java.net.URI;
@@ -70,6 +71,7 @@ public class UserAuthorizer {
   private final URI tokenServerUri;
   private final URI userAuthUri;
   private final PKCEProvider pkce;
+  private final ClientAuthenticationType clientAuthenticationType;
 
   /**
    * Constructor with all parameters.
@@ -83,6 +85,8 @@ public class UserAuthorizer {
    * @param tokenServerUri URI of the end point that provides tokens
    * @param userAuthUri URI of the Web UI for user consent
    * @param pkce PKCE implementation
+   * @param clientAuthenticationType ClientAuthentication type in RFC7591. Choice from
+   *     basic/post/none
    */
   private UserAuthorizer(
       ClientId clientId,
@@ -92,7 +96,8 @@ public class UserAuthorizer {
       HttpTransportFactory transportFactory,
       URI tokenServerUri,
       URI userAuthUri,
-      PKCEProvider pkce) {
+      PKCEProvider pkce,
+      ClientAuthenticationType clientAuthentication) {
     this.clientId = Preconditions.checkNotNull(clientId);
     this.scopes = ImmutableList.copyOf(Preconditions.checkNotNull(scopes));
     this.callbackUri = (callbackUri == null) ? DEFAULT_CALLBACK_URI : callbackUri;
@@ -102,6 +107,10 @@ public class UserAuthorizer {
     this.userAuthUri = (userAuthUri == null) ? OAuth2Utils.USER_AUTH_URI : userAuthUri;
     this.tokenStore = (tokenStore == null) ? new MemoryTokensStorage() : tokenStore;
     this.pkce = pkce;
+    this.clientAuthenticationType =
+        (clientAuthentication == null)
+            ? ClientAuthenticationType.CLIENT_SECRET_POST
+            : clientAuthentication;
   }
 
   /**
@@ -159,6 +168,15 @@ public class UserAuthorizer {
    */
   public TokenStore getTokenStore() {
     return tokenStore;
+  }
+
+  /**
+   * Returns the client authentication type in RFC7591
+   *
+   * @return The ClientAuthenticationType
+   */
+  public ClientAuthenticationType getClientAuthenticationType() {
+    return clientAuthenticationType;
   }
 
   /**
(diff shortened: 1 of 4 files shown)


oauth2_http/java/com/google/auth/oauth2/UserAuthorizer.java
```diff
@@ -161,6 +170,15 @@ public TokenStore getTokenStore() {
     return tokenStore;
   }
 
+  /**
+   * Returns the client authentication type in RFC7591
+   *
+   * @return The ClientAuthenticationType
+   */
+  public ClientAuthenticationType getClientAuthenticationType() {
```
lsirac, line 178: Returns the client authentication type as defined in RFC 7591.

@return The {@link ClientAuthenticationType}.

oauth2_http/java/com/google/auth/oauth2/UserAuthorizer.java
```diff
@@ -291,9 +309,11 @@ public UserCredentials getCredentialsFromCode(
     GenericData tokenData = new GenericData();
     tokenData.put("code", code);
     tokenData.put("client_id", clientId.getClientId());
-    tokenData.put("client_secret", clientId.getClientSecret());
     tokenData.put("redirect_uri", resolvedCallbackUri);
     tokenData.put("grant_type", "authorization_code");
+    if (this.clientAuthenticationType == ClientAuthenticationType.CLIENT_SECRET_POST) {
```
lsirac, line 314: Can you group this in an if/else with the if statement below?

oauth2_http/java/com/google/auth/oauth2/UserAuthorizer.java
```diff
@@ -600,7 +638,14 @@ public UserAuthorizer build() {
           transportFactory,
           tokenServerUri,
           userAuthUri,
-          pkce);
+          pkce,
+          clientAuthenticationType);
     }
   }
+
+  public enum ClientAuthenticationType {
```
lsirac, line 646: Can you add javadoc here? Also I think this should go at the top but not entirely sure.

oauth2_http/java/com/google/auth/oauth2/UserAuthorizer.java
```diff
@@ -311,6 +331,13 @@ public UserCredentials getCredentialsFromCode(
         requestFactory.buildPostRequest(new GenericUrl(tokenServerUri), tokenContent);
     tokenRequest.setParser(new JsonObjectParser(OAuth2Utils.JSON_FACTORY));
 
+    if (this.clientAuthenticationType == ClientAuthenticationType.CLIENT_SECRET_BASIC) {
+      BaseEncoding base64 = BaseEncoding.base64();
+      String encodedCredentials =
+          base64.encode((clientId.getClientId() + ":" + clientId.getClientSecret()).getBytes());
```
lsirac, line 337:
```suggestion
      String encodedCredentials =
          BaseEncoding.base64().encode((clientId.getClientId() + ":" + clientId.getClientSecret()).getBytes());
```

oauth2_http/javatests/com/google/auth/oauth2/MockTokenServerTransport.java
```diff
@@ -206,7 +210,8 @@ public LowLevelHttpResponse execute() throws IOException {
             }
             String foundSecret = query.get("client_secret");
             String expectedSecret = clients.get(foundId);
-            if (foundSecret == null || !foundSecret.equals(expectedSecret)) {
+            if ((foundSecret == null || !foundSecret.equals(expectedSecret))
+                && clients.get("Authorization") == null) {
```
lsirac, line 214: I don't think we are validating the auth header anywhere.

lsirac (changes requested):

oauth2_http/java/com/google/auth/oauth2/UserAuthorizer.java
```diff
@@ -83,6 +85,8 @@ public class UserAuthorizer {
    * @param tokenServerUri URI of the end point that provides tokens
    * @param userAuthUri URI of the Web UI for user consent
    * @param pkce PKCE implementation
+   * @param clientAuthenticationType ClientAuthentication type in RFC7591. Choice from
```
lsirac, line 88:
```suggestion
   * @param clientAuthenticationType ClientAuthentication type as defined in RFC 7591
```

oauth2_http/java/com/google/auth/oauth2/UserAuthorizer.java
BigTailWolf: I'm afraid not here. In the Java code practice we are having a token request content built here and from this content to generate a request build. After the request build, we adding proper basic auth (if chosen) to the request.

oauth2_http/java/com/google/auth/oauth2/UserAuthorizer.java
BigTailWolf: I spot checked some code consistency in this repo. Seems the public enums goes top and has a javadoc and private enums at the bottom of the class.
I'll move to top and add javadoc.

oauth2_http/javatests/com/google/auth/oauth2/MockTokenServerTransport.java
BigTailWolf: This is the limitation of `MockTokenServerTransport` as it doesn't capture request header during it's `execute` method. So we have to use an `addHeader` method to simulate what has been added to the headers. And then in the test code call we cal this `addHeader` prior to make the http request.

lsirac: You can get the request header -- we do it in other tests I'm pretty sure

BigTailWolf:
> The library methods here return UserCredentials. Is this compatible with BYOID? We have ExternalAccountAuthorizedUserCredentials for BYOID.

`ExternalAccountAuthorizedUserCredentials` won't need any changes to have the capability of BYOID. Cloud Code can choose this one if they want to switch.
Per the sync with Cloud Code, they currently using `UserAuthorizer` for existing flow, what we do is just adding the capability to `UserAuthorizer` calling token endpoint with basic auth header.

oauth2_http/java/com/google/auth/oauth2/UserAuthorizer.java
```diff
@@ -56,6 +57,17 @@
 /** Handles an interactive 3-Legged-OAuth2 (3LO) user consent authorization. */
 public class UserAuthorizer {
 
+  /**
+   * Represents the client authentication types as specified in RFC 7591.
+   *
+   * <p>For more details, see <a href="https://tools.ietf.org/html/rfc7591">RFC 7591</a>.
```
clundin25, line 63: nit: Missing closing <p>

oauth2_http/java/com/google/auth/oauth2/UserAuthorizer.java
```diff
@@ -291,9 +320,11 @@ public UserCredentials getCredentialsFromCode(
     GenericData tokenData = new GenericData();
     tokenData.put("code", code);
     tokenData.put("client_id", clientId.getClientId());
-    tokenData.put("client_secret", clientId.getClientSecret());
     tokenData.put("redirect_uri", resolvedCallbackUri);
     tokenData.put("grant_type", "authorization_code");
+    if (this.clientAuthenticationType == ClientAuthenticationType.CLIENT_SECRET_POST) {
+      tokenData.put("client_secret", clientId.getClientSecret());
```
clundin25, line 326: Can you add a comment that explains the following

* STS will reject requests that contains this in the POST body.
* This is to maintain existing behavior for existing code.

oauth2_http/java/com/google/auth/oauth2/UserAuthorizer.java
```diff
@@ -311,6 +342,13 @@ public UserCredentials getCredentialsFromCode(
         requestFactory.buildPostRequest(new GenericUrl(tokenServerUri), tokenContent);
     tokenRequest.setParser(new JsonObjectParser(OAuth2Utils.JSON_FACTORY));
 
+    if (this.clientAuthenticationType == ClientAuthenticationType.CLIENT_SECRET_BASIC) {
+      String encodedCredentials =
+          BaseEncoding.base64()
+              .encode((clientId.getClientId() + ":" + clientId.getClientSecret()).getBytes());
```
clundin25, line 348: Can you pull this into a helper function that can be unit tested?

clundin25: Ideally it takes the client id and secret as params, so a unit test can hard code expect values.

oauth2_http/javatests/com/google/auth/oauth2/MockTokenServerTransport.java
clundin25, line 214: Can you pull this condition into a variable? It's getting hard to grok

oauth2_http/java/com/google/auth/oauth2/UserAuthorizer.java
BigTailWolf: That's not a missing. Like the last javadoc change. it won't pass lint if I have that closing `</p>`. And the auto format tool running is to just remove it.

oauth2_http/javatests/com/google/auth/oauth2/MockTokenServerTransport.java
```diff
@@ -97,6 +97,10 @@ public void addAuthorizationCode(
     }
   }
 
+  public void addHeader(String header, String value) {
+    clients.put(header, value);
+  }
```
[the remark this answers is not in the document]
BigTailWolf: I'll tune this method. Actually could be stored in additional parameters map.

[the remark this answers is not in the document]
BigTailWolf: In the mock, to simulate the basic auth behavior.

oauth2_http/javatests/com/google/auth/oauth2/MockTokenServerTransport.java
BigTailWolf: Do it through the way of in-house authorization way.

lsirac:
> > The library methods here return UserCredentials. Is this compatible with BYOID? We have ExternalAccountAuthorizedUserCredentials for BYOID.
> 
> `ExternalAccountAuthorizedUserCredentials` won't need any changes to have the capability of BYOID. Cloud Code can choose this one if they want to switch. Per the sync with Cloud Code, they currently using `UserAuthorizer` for existing flow, what we do is just adding the capability to `UserAuthorizer` calling token endpoint with basic auth header.

The method you've updated here returns UserCredentials. This is not compatible with the BYOID flow.

BigTailWolf:
> > > The library methods here return UserCredentials. Is this compatible with BYOID? We have ExternalAccountAuthorizedUserCredentials for BYOID.
> > 
> > 
> > `ExternalAccountAuthorizedUserCredentials` won't need any changes to have the capability of BYOID. Cloud Code can choose this one if they want to switch. Per the sync with Cloud Code, they currently using `UserAuthorizer` for existing flow, what we do is just adding the capability to `UserAuthorizer` calling token endpoint with basic auth header.
> 
> The method you've updated here returns UserCredentials. This is not compatible with the BYOID flow.

`UserCredentials` is holding `clientId`, `tokenServerUri`, `refreshToken`, `access_token` etc. That's all the user wants.
And during the authentication flow. Now with modification. We have the generate auth URL done inside the `getCredentialsFromCode` with the `addtionalParameters` properly set. That will give the auth url we need, then we make the http request to sts.googleapis.com/v1/oauthtoken and exchange the access token. The access token is going to be set in the property of `UserCredentials`. And this will be return to the user agent (Cloud Code in our case).

Maybe this is not the same flow with our existing BYOID flow. I believe the flow is still work as help the `UserAuthorizer` class able to get the access token with minimum changes.

oauth2_http/java/com/google/auth/oauth2/UserAuthorizer.java
```diff
@@ -83,6 +96,8 @@ public class UserAuthorizer {
    * @param tokenServerUri URI of the end point that provides tokens
    * @param userAuthUri URI of the Web UI for user consent
    * @param pkce PKCE implementation
+   * @param clientAuthenticationType ClientAuthentication type as defined in RFC 7591
+   *     basic/post/none
```
lqiu96, line 100:
```suggestion
   * @param clientAuthentication ClientAuthentication type as defined in RFC 7591
   *     basic/post/none
```

oauth2_http/java/com/google/auth/oauth2/UserAuthorizer.java
```diff
@@ -92,7 +107,8 @@ private UserAuthorizer(
       HttpTransportFactory transportFactory,
       URI tokenServerUri,
       URI userAuthUri,
-      PKCEProvider pkce) {
+      PKCEProvider pkce,
+      ClientAuthenticationType clientAuthentication) {
```
lqiu96, line 111: Since this is a private constructor, can we update this to take in the Builder class instead of additional parameters? 

```
private UserAuthorizer(Builder builder) {
  // Assign the fields.
}
```

oauth2_http/java/com/google/auth/oauth2/UserAuthorizer.java
```diff
@@ -102,6 +118,10 @@ private UserAuthorizer(
     this.userAuthUri = (userAuthUri == null) ? OAuth2Utils.USER_AUTH_URI : userAuthUri;
     this.tokenStore = (tokenStore == null) ? new MemoryTokensStorage() : tokenStore;
     this.pkce = pkce;
+    this.clientAuthenticationType =
+        (clientAuthentication == null)
+            ? ClientAuthenticationType.CLIENT_SECRET_POST
+            : clientAuthentication;
```
lqiu96, line 124: Just to confirm my understanding, the default ClientAuthenticationType would be `CLIENT_SECRET_POST`, right? This would be the type if a user doesn't explicitly specify it/ set it via the setter?

And `ClientAuthenticationType.NONE` has a different behavior and needs to be explicitly set if they want to use NONE.

oauth2_http/java/com/google/auth/oauth2/UserAuthorizer.java
```diff
@@ -291,9 +320,15 @@ public UserCredentials getCredentialsFromCode(
     GenericData tokenData = new GenericData();
     tokenData.put("code", code);
     tokenData.put("client_id", clientId.getClientId());
-    tokenData.put("client_secret", clientId.getClientSecret());
     tokenData.put("redirect_uri", resolvedCallbackUri);
     tokenData.put("grant_type", "authorization_code");
+    // If the token request is sent to STS. STS will fail by design if the `client_secret` in post
+    // body.
+    // Sending client_secret when the client authentication type is post is to maintain existing
+    // behavior for existing code.
```
lqiu96, line 328: Since this is a public method, can you add this behavior to the javadocs above? i.e. something along the lines of `No validation for uri with clientAuthorizationType and may result in failure` (but worded clearer)

oauth2_http/java/com/google/auth/oauth2/UserAuthorizer.java
```diff
@@ -470,6 +508,19 @@ public void onChanged(OAuth2Credentials credentials) throws IOException {
...
+   *
+   * @param clientId A ClientId object which contains a string value and a secret
+   * @return The basic auth header
+   */
+  public static String getBasicAuthString(ClientId clientId) {
+    String encodedCredentials =
+        BaseEncoding.base64()
+            .encode((clientId.getClientId() + ":" + clientId.getClientSecret()).getBytes());
+    return "Basic " + encodedCredentials;
+  }
```
lqiu96, line 522: nit: Can you make this either private scope or move it to a public util class (if there exists one)?

oauth2_http/java/com/google/auth/oauth2/UserAuthorizer.java
BigTailWolf: Correct. The `CLIENT_SECRET_POST` is what the current behavior following, so that should be the default.

And `NONE` shall be explicitly set. It throw errors naturally since on backend we need some sort of basic auth either having `client_secret` in post or having `authorization` in the headers.

oauth2_http/java/com/google/auth/oauth2/UserAuthorizer.java
BigTailWolf: Make sense. Moved to `OAuth2Utils`

oauth2_http/java/com/google/auth/oauth2/UserAuthorizer.java
BigTailWolf: No problem. Moved to javadoc

oauth2_http/java/com/google/auth/oauth2/UserAuthorizer.java
lqiu96: I believe you can change the builder's build() method so that it pass the builder as the sole parameter to the constructor:
```
    public UserAuthorizer build() {
      return new UserAuthorizer(this);
    }
```

I think this would be cleaner because future implementations can add additional params without having to explicitly new fields to the constructor:
```
  private UserAuthorizer(UserAuthorizer.Builder builder) {
    this.clientId = Preconditions.checkNotNull(builder.clientId);
    this.scopes = ImmutableList.copyOf(Preconditions.checkNotNull(builder.scopes));
    this.callbackUri = (builder.callbackUri == null) ? DEFAULT_CALLBACK_URI : builder.callbackUri;
    this.transportFactory =
        (builder.transportFactory == null) ? OAuth2Utils.HTTP_TRANSPORT_FACTORY : builder.transportFactory;
    this.tokenServerUri = (builder.tokenServerUri == null) ? OAuth2Utils.TOKEN_SERVER_URI : builder.tokenServerUri;
    this.userAuthUri = (builder.userAuthUri == null) ? OAuth2Utils.USER_AUTH_URI : builder.userAuthUri;
    this.tokenStore = (builder.tokenStore == null) ? new MemoryTokensStorage() : builder.tokenStore;
    this.pkce = builder.pkce;
    this.clientAuthenticationType =
        (builder.clientAuthenticationType == null)
            ? ClientAuthenticationType.CLIENT_SECRET_POST
            : builder.clientAuthenticationType;
  }
```

I think this can be refactored even more, so that the builder sets default values and it doesn't need to be validated in the constructor:
```
  public static class Builder {

    private ClientId clientId;
    private TokenStore tokenStore = new MemoryTokensStorage();
    private URI callbackUri = DEFAULT_CALLBACK_URI;
    private URI tokenServerUri = OAuth2Utils.TOKEN_SERVER_URI;
    private URI userAuthUri = OAuth2Utils.USER_AUTH_URI;
    private Collection<String> scopes;
    private HttpTransportFactory transportFactory = OAuth2Utils.HTTP_TRANSPORT_FACTORY;
    private PKCEProvider pkce;
    private ClientAuthenticationType clientAuthenticationType = ClientAuthenticationType.CLIENT_SECRET_POST;
...
```

and the constructor could be something like:
```
  private UserAuthorizer(UserAuthorizer.Builder builder) {
    this.clientId = Preconditions.checkNotNull(builder.clientId);
    this.scopes = ImmutableList.copyOf(Preconditions.checkNotNull(builder.scopes));
    this.callbackUri = builder.callbackUri;
    this.transportFactory = builder.transportFactory;
    this.tokenServerUri = builder.tokenServerUri;
    this.userAuthUri = builder.userAuthUri;
    this.tokenStore = builder.tokenStore;
    this.pkce = builder.pkce;
    this.clientAuthenticationType = builder.clientAuthenticationType;
  }
```

BigTailWolf: The refactor make sense to me. But that's sort of a bigger refactor than this change and irrelevant to the purpose of this change.
Can I do a separate one for that? Maybe create an issue and resolve it through the refactor change.

lqiu96: Sure, it's a private constructor and we can change implementation later. If you could create an issue in this repo and follow up on it sometime in the coming weeks, that would be great! Thanks!

PR #1428, opened 2024-07-03
fix: ComputeEngineCredentials.createScoped should invalidate existing AccessToken

### LINKED ISSUE #1387: ComputeEngineCredentials.createScoped copies existing AccessToken
After upgrading com.google.oauth-client from v1.21.0 to 1.23.0 we started getting 403 ACCESS_TOKEN_SCOPE_INSUFFICIENT errors from the PlayIntegrity.decodeIntegrityToken service. This was the case until the initial AccessToken expired after which it started working again.

We obtained the credentials used (through a HttpCredentialsAdapter) to access the PlayIntegrity service in this way: 

```java
var scopedCredentials = GoogleCredentials.getApplicationDefault().createScoped(PlayIntegrityScopes.all());
```
The cause of the changed behavior appears to be from this commit: 
https://github.com/googleapis/google-auth-library-java/commit/7e268611d2c2152e84702b1c67ca846902bbe2d5
which changes the ComputeEngineCredentials.createScope implementation to copy a lot more state from the original ComputeEngineCredentials object to the new, scoped ComputeEngineCredentials object. This includes copying the AccessToken from the original, which does not seem like the desired behaviour when the scopes are updated. 

We have added an immediate refresh after creating the scoped credentials to make sure we get a new access token with the correct scope. This has solved the issue for us, but I assume this may be a gotcha that affects more users.    

Workaround:
```java
var scopedCredentials = GoogleCredentials.getApplicationDefault().createScoped(PlayIntegrityScopes.all());
scopedCredentials.refresh();
```

zhumin8:
Fixes #1387 ☕️

As described in the original issue, this looks like a regression introduced in https://github.com/googleapis/google-auth-library-java/commit/7e268611d2c2152e84702b1c67ca846902bbe2d5 when migrating from deprecated constructor to use builder. 
[Access token](https://cloud.google.com/docs/authentication/token-types#access-contents) is scoped and should be invalidated when scope changes.

This PR include changes:
- ComputeEngineCredentials.createScoped() should invalidate existing AccessToken 
- ComputeEngineCredentials.createScoped(newScopes, newDefaultScopes) should respect universe domain settings.
- For testing, update to mock transport to support returning access token when url is scoped. Should return a different access token compared to default url with no scopes. 
- ~~Other credential types: e.g. ServiceAccountCredentials.creatScoped() should also invalidate existing AccessToken~~ Will raise separate pr for this.

Follow up items not included in this PR:
- Check if other GoogleCredentials subclasses that overrides createScoped() suffer from same issue (e.g. `ServiceAccountCredentials.creatScoped()` )

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
@@ -163,14 +163,12 @@ public class ComputeEngineCredentials extends GoogleCredentials
   @Override
   public GoogleCredentials createScoped(
       Collection<String> newScopes, Collection<String> newDefaultScopes) {
-    ComputeEngineCredentials.Builder builder =
-        (Builder)
-            this.toBuilder()
-                .setHttpTransportFactory(transportFactory)
-                .setScopes(newScopes)
-                .setDefaultScopes(newDefaultScopes)
-                .setAccessToken(null);
-    return new ComputeEngineCredentials(builder);
+    return this.toBuilder()
+        .setHttpTransportFactory(transportFactory)
+        .setScopes(newScopes)
+        .setDefaultScopes(newDefaultScopes)
+        .setAccessToken(null)
+        .build();
   }
 
   /**
--- a/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
@@ -48,6 +48,7 @@ import com.google.auth.http.HttpTransportFactory;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Joiner;
 import com.google.common.base.MoreObjects.ToStringHelper;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.errorprone.annotations.CanIgnoreReturnValue;
 import java.io.BufferedReader;
@@ -155,7 +156,7 @@ public class ComputeEngineCredentials extends GoogleCredentials
   /** Clones the compute engine account with the specified scopes. */
   @Override
   public GoogleCredentials createScoped(Collection<String> newScopes) {
-    return createScoped(newScopes, null);
+    return createScoped(newScopes, ImmutableList.of());
   }
 
   /** Clones the compute engine account with the specified scopes and default scopes. */
--- a/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
@@ -156,7 +156,11 @@ public class ComputeEngineCredentials extends GoogleCredentials
   @Override
   public GoogleCredentials createScoped(Collection<String> newScopes) {
     ComputeEngineCredentials.Builder builder =
-        this.toBuilder().setHttpTransportFactory(transportFactory).setScopes(newScopes);
+        (Builder)
+            this.toBuilder()
+                .setHttpTransportFactory(transportFactory)
+                .setScopes(newScopes)
+                .setAccessToken(null);
     return new ComputeEngineCredentials(builder);
   }
 
@@ -165,10 +169,12 @@ public class ComputeEngineCredentials extends GoogleCredentials
   public GoogleCredentials createScoped(
       Collection<String> newScopes, Collection<String> newDefaultScopes) {
     ComputeEngineCredentials.Builder builder =
-        ComputeEngineCredentials.newBuilder()
-            .setHttpTransportFactory(transportFactory)
-            .setScopes(newScopes)
-            .setDefaultScopes(newDefaultScopes);
+        (Builder)
+            ComputeEngineCredentials.newBuilder()
+                .setHttpTransportFactory(transportFactory)
+                .setScopes(newScopes)
+                .setDefaultScopes(newDefaultScopes)
+                .setAccessToken(null);
     return new ComputeEngineCredentials(builder);
   }
 
--- a/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
@@ -155,13 +155,7 @@ public class ComputeEngineCredentials extends GoogleCredentials
   /** Clones the compute engine account with the specified scopes. */
   @Override
   public GoogleCredentials createScoped(Collection<String> newScopes) {
-    ComputeEngineCredentials.Builder builder =
-        (Builder)
-            this.toBuilder()
-                .setHttpTransportFactory(transportFactory)
-                .setScopes(newScopes)
-                .setAccessToken(null);
-    return new ComputeEngineCredentials(builder);
+    return createScoped(newScopes, null);
   }
 
   /** Clones the compute engine account with the specified scopes and default scopes. */
@@ -170,7 +164,7 @@ public class ComputeEngineCredentials extends GoogleCredentials
       Collection<String> newScopes, Collection<String> newDefaultScopes) {
     ComputeEngineCredentials.Builder builder =
         (Builder)
-            ComputeEngineCredentials.newBuilder()
+            this.toBuilder()
                 .setHttpTransportFactory(transportFactory)
                 .setScopes(newScopes)
                 .setDefaultScopes(newDefaultScopes)
--- a/oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
@@ -427,8 +427,8 @@ public class GoogleCredentials extends OAuth2Credentials implements QuotaProject
   }
 
   /**
-   * If the credentials support scopes, creates a copy of the identity with the specified scopes;
-   * otherwise, returns the same instance.
+   * If the credentials support scopes, creates a copy of the identity with the specified scopes,
+   * invalidate access token (scoped); otherwise, returns the same instance.
    *
    * @param scopes Collection of scopes to request.
    * @return GoogleCredentials with requested scopes.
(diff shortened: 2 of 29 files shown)


oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -155,20 +155,20 @@ private ComputeEngineCredentials(ComputeEngineCredentials.Builder builder) {
   /** Clones the compute engine account with the specified scopes. */
   @Override
   public GoogleCredentials createScoped(Collection<String> newScopes) {
-    ComputeEngineCredentials.Builder builder =
-        this.toBuilder().setHttpTransportFactory(transportFactory).setScopes(newScopes);
-    return new ComputeEngineCredentials(builder);
+    return createScoped(newScopes, null);
```
lqiu96, line 158: nit: Perhaps use `ImmuntableList.of()` or `ImmutableSet.of()` instead of null for the second param?

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -155,20 +155,20 @@ private ComputeEngineCredentials(ComputeEngineCredentials.Builder builder) {
...
-        ComputeEngineCredentials.newBuilder()
-            .setHttpTransportFactory(transportFactory)
-            .setScopes(newScopes)
-            .setDefaultScopes(newDefaultScopes);
+        (Builder)
+            this.toBuilder()
+                .setHttpTransportFactory(transportFactory)
+                .setScopes(newScopes)
+                .setDefaultScopes(newDefaultScopes)
+                .setAccessToken(null);
```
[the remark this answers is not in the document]
zhumin8: I am trying to narrow the scope of this pr only fix this regression in behavior introduced by https://github.com/googleapis/google-auth-library-java/commit/7e268611d2c2152e84702b1c67ca846902bbe2d5. To matches the original behavior.
That said, in a follow up PR, I am considering moving the logic upstream to  `GoogleCredentials` or `OAuth2Credentials`. One thought is aside from createScoped(), does the same issue exists in creating copies and setting other properties differently, e.g. set a different universe. So perhaps  going to invalidate directly in `toBuilder()`. We can discuss further when I put up that PR.

> If a user passes in the same set of scopes (i.e. does not actually change the set of scopes for the credentials), this would invalidate the access token, right?

In this case, the old access token would probably work too, but it would be safer or at least no harm to invalidate it anyway and let the refresh workflow request the token from server again. 

> Do you think it would make sense to check to see if the scopes changed and only invalidate the access token then? If that makes sense, perhaps that could be done in GoogleCredentials so that it could apply to every single type of Credentials?

Hmm, let's discuss more in the follow-up pr, this will be a new check, and I'd rather introduce it in GoogleCredentials than individual credential classes if possible.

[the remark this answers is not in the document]
lqiu96: Sounds good. We'll keep this scope for just fixing the bug and address everything else in a follow up PR.

> In this case, the old access token would probably work too, but it would be safer or at least no harm to invalidate it anyway and let the refresh workflow request the token from server again.

Yep makes sense. I just wondering what the expected behavior is from the user POV is. Also, wondering what the behavior from other language auth libraries. I think it makes sense to always invalidate the access token even if the access token is technically valid.

oauth2_http/javatests/com/google/auth/oauth2/ComputeEngineCredentialsTest.java
```diff
@@ -228,6 +242,20 @@ public void createScoped_defaultScopes() {
...
+        ComputeEngineCredentials.newBuilder()
+            .setScopes(null)
+            .setUniverseDomain("some-universe")
+            .build();
+    ComputeEngineCredentials scopedCredentials =
+        (ComputeEngineCredentials) credentials.createScoped(null, Arrays.asList("foo"));
+
+    assertEquals("some-universe", scopedCredentials.getUniverseDomain());
+    assertTrue(scopedCredentials.isExplicitUniverseDomain());
+  }
```
lqiu96, line 257: I think this test is fine, but perhaps we could make this one and the one above more generic. IIUC, the crux of the test is that it's testing that when `createScoped(...)` is called, the internal values are still the same (i.e. universe_domain).

Maybe a test where it sets stuff like quotaProjectId, universeDomain, accessToken, refereshMargin, scope, etc and ensures that everything is still the same except accessToken and scope. Test could be something like `buildDefaultScoped_fieldsRetained_accessTokenInvalidated()` for example (name could be better) or perhaps it can be merged with `buildScoped_shouldInvalidateAccessToken` from above.

oauth2_http/javatests/com/google/auth/oauth2/MockMetadataServerTransport.java
```diff
@@ -84,7 +84,7 @@ public void setIdToken(String idToken) {
 
   @Override
   public LowLevelHttpRequest buildRequest(String method, String url) throws IOException {
-    if (url.equals(ComputeEngineCredentials.getTokenServerEncodedUrl())) {
+    if (url.startsWith(ComputeEngineCredentials.getTokenServerEncodedUrl())) {
```
lqiu96, line 87: qq, why does this need to be changed? Is this because for `getRequestMetadata_newAccessTokenWhenScoped`, the first request doesn't have any scopes, but the second request (once `getScoped()`) is called, the request has the query param scopes added to the request?

zhumin8: Yes. Basically without this change, mock transport only deal cases with no scopes specified. It will fall back to default behavior and return success code with empty content for the second request with scopes, and eventually hit [this exception](https://github.com/googleapis/google-auth-library-java/blob/cc957e3738dc2f73e86bbef0668f611eb46fe99f/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java#L287-L289). This change and the added logic below let the mock transport return a different access token when scopes provided.

oauth2_http/javatests/com/google/auth/oauth2/MockMetadataServerTransport.java
```diff
@@ -164,7 +164,11 @@ public LowLevelHttpResponse execute() throws IOException {
         // Create the JSON response
         GenericJson refreshContents = new GenericJson();
         refreshContents.setFactory(OAuth2Utils.JSON_FACTORY);
-        refreshContents.put("access_token", accessToken);
+        if (url.contains("?scopes=")) {
+          refreshContents.put("access_token", "fake access token with scope");
```
lqiu96, line 168: If possible, could the fake access token not be hard coded in the MockMetaDataServerTransport code?

Perhaps we could make the logic something like:
1. MockMetaDataServerTransport takes a map of scopes -> access token pairings/ MockMetaDataServerTransport.addAccessTokenScope(scope, accessToken) will add the the pairing
2. If url.contains("?scopes="), check the map for scopes and returns the configured access token
3. Test passes the pairing of (scope, accesstoken) to MockMetaDataServerTransport

oauth2_http/javatests/com/google/auth/oauth2/ComputeEngineCredentialsTest.java
zhumin8: I merged these two new tests together. 
As of other properties, most of them are already covered in existing tests [here](https://github.com/googleapis/google-auth-library-java/blob/b995726fa657856f02ad474220f5734ae91f0331/oauth2_http/javatests/com/google/auth/oauth2/ComputeEngineCredentialsTest.java#L194-L220), and I added a new one verifying quotaProjectId.

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -427,8 +427,8 @@ public boolean createScopedRequired() {
   }
 
   /**
-   * If the credentials support scopes, creates a copy of the identity with the specified scopes;
-   * otherwise, returns the same instance.
+   * If the credentials support scopes, creates a copy of the identity with the specified scopes,
+   * invalidate access token (scoped); otherwise, returns the same instance.
```
lqiu96, line 431:
```suggestion
   * If the credentials support scopes, creates a copy of the identity with the specified scopes,
   * invalidates the existing scoped access token; otherwise, return the same instance.
```

oauth2_http/javatests/com/google/auth/oauth2/ComputeEngineCredentialsTest.java
```diff
@@ -250,6 +277,29 @@ public void getRequestMetadata_hasAccessToken() throws IOException {
...
+    TestUtils.assertContainsBearerToken(metadata, accessToken);
+
+    assertNotNull(credentials.getAccessToken());
+    ComputeEngineCredentials scopedCredentialCopy =
+        (ComputeEngineCredentials) credentials.createScoped(Arrays.asList("foo", "bar"));
+    assertNull(scopedCredentialCopy.getAccessToken());
+    Map<String, List<String>> metadataForCopiedCredentials =
+        scopedCredentialCopy.getRequestMetadata(CALL_URI);
+    TestUtils.assertContainsBearerToken(metadataForCopiedCredentials, accessTokenWithScopes);
+  }
```
lqiu96, line 301: nit: I think one last assert which is assert that metadataForCopiedCredentials's access token does not equal metadata's accessToken.

It would be that `1/MkSJoj1xsli0AccessToken_NKPY2` != `fake access token with scope`

oauth2_http/javatests/com/google/auth/oauth2/MockMetadataServerTransport.java
```diff
@@ -63,7 +65,14 @@ public class MockMetadataServerTransport extends MockHttpTransport {
...
 
   public void setAccessToken(String accessToken) {
-    this.accessToken = accessToken;
+    setAccessToken("default", accessToken);
+  }
+
+  public void setAccessToken(String scopes, String accessToken) {
+    if (scopesToAccessToken == null) {
+      scopesToAccessToken = new HashMap<>();
+    }
```
lqiu96, line 74: nit: initialize in the constructor

lqiu96 (approved):
LGTM, but one last review from Wes

arithmetic1728 (approved):
LGTM

Issue #1451, opened 2024-07-30
SignBlob call to IAM should retry


lqiu96:
See b/341158194 for more information.

The call should retry with exponential backoff before returning an exception back to the user.

PR #1452, opened 2024-07-30
fix: Retry sign blob call  with exponential backoff

lqiu96:
Fixes https://github.com/googleapis/google-auth-library-java/issues/1451

The overall changes will be split into two parts. This part is part 1 where retries will be enabled and tests to cover successful and failed calls.

Part 2 will add additional tests to cover retries for the IAM Mock Server. Split it out to a separate PR since the IAM Mock Server will need to be refactored to support retries and variable number of responses. The refactor will touch multiple files that are non-related to adding retries to the Sign Blob RPC.

Changes:
1. Sign Blob RPC will have retries for certain 5xx status codes
2. Centralize default retry parameter values in Oauth2Utils for all retryable RPCs

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/OAuth2Utils.java
+++ b/oauth2_http/java/com/google/auth/oauth2/OAuth2Utils.java
@@ -102,11 +102,6 @@ class OAuth2Utils {
   public static final Set<Integer> TOKEN_ENDPOINT_RETRYABLE_STATUS_CODES =
       new HashSet<>(Arrays.asList(500, 503, 408, 429));
 
-  // Following guidance for IAM retries:
-  // https://cloud.google.com/iam/docs/retry-strategy#errors-to-retry
-  static final Set<Integer> IAM_RETRYABLE_STATUS_CODES =
-      new HashSet<>(Arrays.asList(500, 502, 503, 504));
-
   static class DefaultHttpTransportFactory implements HttpTransportFactory {
 
     @Override
--- a/oauth2_http/java/com/google/auth/oauth2/OAuth2Utils.java
+++ b/oauth2_http/java/com/google/auth/oauth2/OAuth2Utils.java
@@ -92,10 +92,20 @@ class OAuth2Utils {
 
   static final String TOKEN_RESPONSE_SCOPE = "scope";
 
+  static final int INITIAL_RETRY_INTERVAL_MILLIS = 1000;
+  static final double RETRY_RANDOMIZATION_FACTOR = 0.1;
+  static final double RETRY_MULTIPLIER = 2;
+  static final int DEFAULT_NUMBER_OF_RETRIES = 3;
+
   // Includes expected server errors from Google token endpoint
   // Other 5xx codes are either not used or retries are unlikely to succeed
   public static final Set<Integer> TOKEN_ENDPOINT_RETRYABLE_STATUS_CODES =
       new HashSet<>(Arrays.asList(500, 503, 408, 429));
+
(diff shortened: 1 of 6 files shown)


oauth2_http/java/com/google/auth/oauth2/TokenVerifier.java
```diff
@@ -279,9 +279,6 @@ public TokenVerifier build() {
   /** Custom CacheLoader for mapping certificate urls to the contained public keys. */
   static class PublicKeyLoader extends CacheLoader<String, Map<String, PublicKey>> {
     private static final int DEFAULT_NUMBER_OF_RETRIES = 2;
```
zhumin8, line 281: curious if there is specific guideline for TokenVerifier to default retry 2 times?

zhumin8 (commented):
LGTM

oauth2_http/java/com/google/auth/oauth2/TokenVerifier.java
lqiu96: I don't think so. I think these values were randomly selected and used. I've kept it the same for now and we can re-visit retries again in the future.

zhumin8 (approved):

oauth2_http/java/com/google/auth/oauth2/OAuth2Utils.java
```diff
@@ -92,11 +92,21 @@ class OAuth2Utils {
...
+  static final int DEFAULT_NUMBER_OF_RETRIES = 3;
+
   // Includes expected server errors from Google token endpoint
   // Other 5xx codes are either not used or retries are unlikely to succeed
   public static final Set<Integer> TOKEN_ENDPOINT_RETRYABLE_STATUS_CODES =
       new HashSet<>(Arrays.asList(500, 503, 408, 429));
 
+  // Following guidance for IAM retries:
+  // https://cloud.google.com/iam/docs/retry-strategy#errors-to-retry
+  static final Set<Integer> IAM_RETRYABLE_STATUS_CODES =
```
westarle, line 107: nit: seems like this should live in IamUtils.

westarle (commented):
This looks pretty good, I would like to see verification in the test:

1) success case makes a single request
2) 401s are not retried
3) 50x errors are retried according to policy to failure
4) 50x errors are retried according to policy to success

oauth2_http/java/com/google/auth/oauth2/OAuth2Utils.java
lqiu96: Ah, yep makes sense. I'll move it over.

lqiu96:
> This looks pretty good, I would like to see verification in the test:
> 
> 1. success case makes a single request
> 2. 401s are not retried
> 3. 50x errors are retried according to policy to failure
> 4. 50x errors are retried according to policy to success

Yep, those test cases make sense. I was running into some limitations of MockIAMCredentialsServiceTransport class where I couldn't specify that 5xx errors are retries and couldn't control how many times it would retry for.

I made some changes that to the class to support retrying and controlling the number of retry attempts. Those changes ended up touch more than just the SignBlob RPC so I made that into [Part 2](https://github.com/googleapis/google-auth-library-java/pull/1455).

I believe this PR only covers cases 1 and 2, but 3 and 4 should be tested in the second PR.

westarle (approved):
If tests in #1455 pass, this LGTM.

lqiu96:
The additional test cases added in #1455 pass. Will first merge in this PR and rebase #1455 with these changes.

PR #1455, opened 2024-08-01
chore: Control each response for calls to the IAM Mock Server

lqiu96:
This is part 2 that follows up on: https://github.com/googleapis/google-auth-library-java/pull/1452

Once Part 1 is merged, I can rebase this PR to show the diffs (mostly testing related changes) to allow us to test for retrying the IAM Sign Blob RPC.

### DIFF
--- a/oauth2_http/javatests/com/google/auth/oauth2/IamUtilsTest.java
+++ b/oauth2_http/javatests/com/google/auth/oauth2/IamUtilsTest.java
@@ -53,7 +53,8 @@ public class IamUtilsTest {
   public void sign_noRetry() throws IOException {
     byte[] expectedSignature = {0xD, 0xE, 0xA, 0xD};
 
-    // Mock this call because signing requires an access token
+    // Mock this call because signing requires an access token. The call is initialized with
+    // HttpCredentialsAdapter which will make a call to get the access token
     ServiceAccountCredentials credentials = Mockito.mock(ServiceAccountCredentials.class);
     Mockito.when(credentials.getRequestMetadata(Mockito.any())).thenReturn(ImmutableMap.of());
 
@@ -76,7 +77,8 @@ public class IamUtilsTest {
   public void sign_4xxServerError_exception() throws IOException {
     byte[] expectedSignature = {0xD, 0xE, 0xA, 0xD};
 
-    // Mock this call because signing requires an access token
+    // Mock this call because signing requires an access token. The call is initialized with
+    // HttpCredentialsAdapter which will make a call to get the access token
     ServiceAccountCredentials credentials = Mockito.mock(ServiceAccountCredentials.class);
     Mockito.when(credentials.getRequestMetadata(Mockito.any())).thenReturn(ImmutableMap.of());
 
--- a/oauth2_http/javatests/com/google/auth/oauth2/IamUtilsTest.java
+++ b/oauth2_http/javatests/com/google/auth/oauth2/IamUtilsTest.java
@@ -1,0 +1,77 @@
+package com.google.auth.oauth2;
+
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertThrows;
+import static org.junit.Assert.assertTrue;
+
+import com.google.api.client.http.HttpStatusCodes;
+import com.google.auth.ServiceAccountSigner;
+import com.google.common.collect.ImmutableMap;
+import java.io.IOException;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import org.mockito.Mockito;
+
+@RunWith(JUnit4.class)
+public class IamUtilsTest {
+
+  private static final String CLIENT_EMAIL =
+      "36680232662-vrd7ji19qe3nelgchd0ah2csanun6bnr@developer.gserviceaccount.com";
+
+  @Test
+  public void sign_noRetry() throws IOException {
+    byte[] expectedSignature = {0xD, 0xE, 0xA, 0xD};
+
+    // Mock this call because signing requires an access token
+    ServiceAccountCredentials credentials = Mockito.mock(ServiceAccountCredentials.class);
+    Mockito.when(credentials.getRequestMetadata(Mockito.any())).thenReturn(ImmutableMap.of());
+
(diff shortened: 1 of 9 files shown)


oauth2_http/javatests/com/google/auth/oauth2/IamUtilsTest.java
```diff
@@ -0,0 +1,176 @@
...
+import org.mockito.Mockito;
+
+@RunWith(JUnit4.class)
+public class IamUtilsTest {
+
+  private static final String CLIENT_EMAIL =
+      "36680232662-vrd7ji19qe3nelgchd0ah2csanun6bnr@developer.gserviceaccount.com";
+
+  @Test
+  public void sign_noRetry() throws IOException {
```
westarle, line 53: sign_success() ?

oauth2_http/javatests/com/google/auth/oauth2/IamUtilsTest.java
```diff
@@ -0,0 +1,176 @@
...
+    byte[] expectedSignature = {0xD, 0xE, 0xA, 0xD};
+
+    // Mock this call because signing requires an access token
+    ServiceAccountCredentials credentials = Mockito.mock(ServiceAccountCredentials.class);
+    Mockito.when(credentials.getRequestMetadata(Mockito.any())).thenReturn(ImmutableMap.of());
+
+    ImpersonatedCredentialsTest.MockIAMCredentialsServiceTransportFactory transportFactory =
+        new ImpersonatedCredentialsTest.MockIAMCredentialsServiceTransportFactory();
+    transportFactory.transport.addStatusCodeAndMessage(502, "Bad Gateway");
+    transportFactory.transport.addStatusCodeAndMessage(502, "Bad Gateway");
```
westarle, line 90: Should this fail 3 times to test the boundary?

oauth2_http/javatests/com/google/auth/oauth2/IamUtilsTest.java
```diff
@@ -0,0 +1,176 @@
...
+        assertThrows(
+            ServiceAccountSigner.SigningException.class,
+            () ->
+                IamUtils.sign(
+                    CLIENT_EMAIL,
+                    credentials,
+                    transportFactory.transport,
+                    expectedSignature,
+                    ImmutableMap.of()));
+    assertTrue(exception.getMessage().contains("Failed to sign the provided bytes"));
```
westarle, line 169: Can you assert that the server error is not retried?  Or is that implicit in the setup of the transport? -- if so please leave a comment about that.

Otherwise please assert a 2nd call didn't arrive, or have the transport returning a success on 2nd request.

oauth2_http/javatests/com/google/auth/oauth2/IamUtilsTest.java
lqiu96: Update to include `success`. Thanks!

oauth2_http/javatests/com/google/auth/oauth2/IamUtilsTest.java
lqiu96: I updated the description for the following test below `sign_retryThreeTimes_exception`. I believe that will test the retry boundary as it fails three times before giving up.

oauth2_http/javatests/com/google/auth/oauth2/IamUtilsTest.java
lqiu96: Sounds good. I added a method in the transport to track the number of requests being made. This tests should include an assertion `assertEquals(4, transportFactory.transport.getNumRequests());` which should ensure that a specific number of retry attempts are made and that the final 5xx errors aren't being retried.

oauth2_http/javatests/com/google/auth/oauth2/IamUtilsTest.java
```diff
@@ -0,0 +1,176 @@
...
+    Mockito.when(credentials.getRequestMetadata(Mockito.any())).thenReturn(ImmutableMap.of());
+
+    ImpersonatedCredentialsTest.MockIAMCredentialsServiceTransportFactory transportFactory =
+        new ImpersonatedCredentialsTest.MockIAMCredentialsServiceTransportFactory();
+    transportFactory.transport.setSignedBlob(expectedSignature);
+    transportFactory.transport.setTargetPrincipal(CLIENT_EMAIL);
+    transportFactory.transport.addStatusCodeAndMessage(502, "Bad Gateway");
+    transportFactory.transport.addStatusCodeAndMessage(502, "Bad Gateway");
+    transportFactory.transport.addStatusCodeAndMessage(502, "Bad Gateway");
+    transportFactory.transport.addStatusCodeAndMessage(502, "Bad Gateway");
```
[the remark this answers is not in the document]
lqiu96: Sure, done! There are a few 5xx errors, so I just swapped them to be 500, 502 and 503. It just reports the last status code that came back once all the retries are exhausted. Hopefully this should be clearer.

oauth2_http/javatests/com/google/auth/oauth2/IamUtilsTest.java
westarle: The idea here was to test the other side of the boundary: does the code allow the "last retry" to be successful?  This isn't tested now.

lqiu96: Apologies if I'm misunderstanding your point. The retry is configured to retry three times before giving up and I believe the two sides of the boundaries you are talking about are:
1. Retrying while it hasn't reached the number of attempts limit
2. Stop retrying after it has reached the number of attempts limit

The first test case `sign_retryTwoTimes_success` will receive two 5xx errors and should test the first case above.

The second test case `sign_retryThreeTimes_exception` will receive a four 5xx errors and should test the second case above. It'll retry 3 times and the 4th response will be what is sent back to the user. If it's a 5xx error, it'll throw an exception back, otherwise it'll send the response.

> does the code allow the "last retry" to be successful?

If I'm understanding this comment correctly, I'll add another test case where it'll have three 5xx errors before returning a successful response. This should return a success back to the user.

zhumin8 (approved):
LGTM

PR #1483, opened 2024-08-28
feat: add api key credential as client library authorization type

ldetmer:
Add new ApiKeyCredentials that will pass the appropriate api-key header for authorizing api requests from client libraries
 
Note: ApiKeyCredentials extends from the base Credential class as Api Keys are not considered oauth2 credentials since there is no access token which is required by GoogleCredentials/OAuth2Credentials.  This also follows definition of oauth2 credentials as specified by [Google Identity Protocols](https://developers.google.com/identity/protocols/oauth2#2.-obtain-an-access-token-from-the-google-authorization-server.) that require an access token when using Oauth 2.0 to access google APIs.


credentials/java/com/google/auth/ApiKeyCredentials.java
```diff
@@ -0,0 +1,83 @@
+/*
+ * Copyright 2015, Google Inc. All rights reserved.
```
lqiu96, line 2: I think:
```suggestion
 * Copyright 2024, Google Inc. All rights reserved.
```

credentials/java/com/google/auth/ApiKeyCredentials.java
```diff
@@ -0,0 +1,83 @@
...
+  static final String API_KEY_HEADER_KEY = "x-goog-api-key";
+  private final String apiKey;
+
+  ApiKeyCredentials(String apiKey) {
+    this.apiKey = apiKey;
+  }
+
+  public static ApiKeyCredentials create(String apiKey) {
+    return new ApiKeyCredentials(apiKey);
+  }
```
lqiu96, line 58: Do we need to check if the user passes in null or empty string ("")?

lqiu96:
I think the changes LGTM. I'll need to double check the authenticationType being `""` and the `refresh()` behavior.

credentials/java/com/google/auth/ApiKeyCredentials.java
ldetmer: thanks good catch

credentials/java/com/google/auth/ApiKeyCredentials.java
ldetmer: sounds like a good idea, added.  used Precondition as that's what I see in other apis, however it did force a new import, so let me know if you would prefer for me to check without using Precondition

ldetmer:
> I think the changes LGTM. I'll need to double check the authenticationType being `""` and the `refresh()` behavior.

FYI, Blake and I  discussed refresh with Carl and he agreed no op was correct.

credentials/java/com/google/auth/ApiKeyCredentials.java
```diff
@@ -0,0 +1,85 @@
...
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package com.google.auth;
+
+import com.google.api.client.util.*;
```
burkedavison, line 33: Replace all wildcard imports with explicit imports.

go/java-style#s3.3.1-wildcard-imports

credentials/pom.xml
```diff
@@ -54,6 +54,10 @@
       <artifactId>junit</artifactId>
       <scope>test</scope>
     </dependency>
+      <dependency>
+          <groupId>com.google.http-client</groupId>
+          <artifactId>google-http-client</artifactId>
```
burkedavison, line 59: Is this dependency needed? With the wildcard import, it's hard to see how it's being used.

lqiu96:
> > I think the changes LGTM. I'll need to double check the authenticationType being `""` and the `refresh()` behavior.
> 
> FYI, Blake and I discussed refresh with Carl and he agreed no op was correct.

Yep. I think it makes sense just wanted to confirm the behavior and I think it works as intended. I'll ask about the authenticationType with the Auth Team and once I get their input on this, I think we should be good to go.

credentials/java/com/google/auth/ApiKeyCredentials.java
```diff
@@ -0,0 +1,85 @@
...
+import java.io.IOException;
+import java.net.URI;
+import java.util.*;
+
+/**
+ * Credentials class for calling Google APIs using an API key.
+ *
```
lqiu96, line 40: Could you add a small snippet in the javadocs (or perhaps anywhere else) explaining why this extends directly from Credentials and not GoogleCredentials/ Oauth2Credentials? 

I believe the original reason is that Api Keys do not have a concept of Access Tokens, but not sure if there were additional reasons behind this.

credentials/java/com/google/auth/ApiKeyCredentials.java
```diff
@@ -0,0 +1,85 @@
...
+public class ApiKeyCredentials extends Credentials {
+  static final String API_KEY_HEADER_KEY = "x-goog-api-key";
+  private final String apiKey;
+
+  ApiKeyCredentials(String apiKey) {
+    this.apiKey = apiKey;
+  }
+
+  public static ApiKeyCredentials create(String apiKey) {
+    Preconditions.checkNotNull(apiKey, "API key cannot be null");
```
zhumin8, line 56: I am a bit hesitate on bring in `google-http-client` just for this utility method.

credentials/java/com/google/auth/ApiKeyCredentials.java
```diff
@@ -0,0 +1,85 @@
...
+/**
+ * Credentials class for calling Google APIs using an API key.
+ *
+ * <p>Uses an API key directly in the request metadata to provide authorization.
+ *
+ * <pre><code>
+ * Credentials credentials = ApiKeyCredentials.create("your api key");
+ * </code></pre>
+ */
+public class ApiKeyCredentials extends Credentials {
```
zhumin8, line 47: I suppose this class is in `google-auth-library-credentials` module unlike other credential classes we have in `google-auth-library-oauth2-http` module because it is not an OAuth2 credential?
Do you mind adding some context or links to the description section so that I can understand a bit more on this and for future tracking purposes? I find context and links to docs very helpful when tracking back changes in the past.

credentials/java/com/google/auth/ApiKeyCredentials.java
ldetmer: I added it in the PR description as its seems a design decision and wouldn't be part of java docs, but happy to add there if think it would be helpful

credentials/pom.xml
ldetmer: added it for Precondition, but since removed as per discussion below, that is overkill

credentials/java/com/google/auth/ApiKeyCredentials.java
ldetmer: make sense, removed

credentials/java/com/google/auth/ApiKeyCredentials.java
ldetmer: added note in description, let me know if you think it need any additional information

credentials/javatests/com/google/auth/ApiKeyCredentialsTest.java
```diff
@@ -0,0 +1,97 @@
...
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package com.google.auth;
+
+import static org.junit.Assert.*;
```
burkedavison, line 33: go/java-style#s3.3.1-wildcard-imports

credentials/java/com/google/auth/ApiKeyCredentials.java
lqiu96: PR Description does help, thanks! I think it may be a bit more complicated/ difficult to trace back the decision to this PR (especially if more PRs/ changes come in). I think that either a short blurb in the comments or perhaps in the original design doc would be easier to find.

credentials/java/com/google/auth/ApiKeyCredentials.java
```diff
@@ -0,0 +1,86 @@
...
+  static final String API_KEY_HEADER_KEY = "x-goog-api-key";
+  private final String apiKey;
+
+  ApiKeyCredentials(String apiKey) {
+    this.apiKey = apiKey;
+  }
+
+  public static ApiKeyCredentials create(String apiKey) {
+    if (apiKey == null || apiKey.isEmpty()) {
```
lqiu96, line 57: nit: Perhaps `if (Strings.isNullOrEmpty(apiKey)) {`

From Guava: https://guava.dev/releases/31.1-jre/api/docs/com/google/common/base/Strings.html#nullToEmpty(java.lang.String) which I think should be part of auth-library's deps.

credentials/javatests/com/google/auth/ApiKeyCredentialsTest.java
```diff
@@ -0,0 +1,97 @@
...
+    assertTrue(credentials.hasRequestMetadataOnly());
+  }
+
+  @Test
+  public void testNullApiKey_ThrowsException() {
+    try {
+      ApiKeyCredentials.create(null);
+      fail("No exception raised");
+    } catch (IllegalArgumentException e) {
+      assert (e.getMessage().contains("API key cannot be null or blank"));
```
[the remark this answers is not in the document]
ldetmer: thanks done!

credentials/java/com/google/auth/ApiKeyCredentials.java
ldetmer: ok added to java docs

credentials/java/com/google/auth/ApiKeyCredentials.java
ldetmer: unfortunately its not part of google-auth-library-credentials, i'm not sure our threshhold for adding new dependencies, as this seems a base utility jar it seems to make sense to add, wdyt?

burkedavison:
LGTM

credentials/java/com/google/auth/ApiKeyCredentials.java
lqiu96: OH, I see. I don't see any harm with adding it to the credentials sub-module. Not an issue for this change, we can always add it in a future PR.

lqiu96 (approved):
LGTM. 

@zhumin8 for approval as well

zhumin8 (approved):

burkedavison (approved):

PR #1486, opened 2024-08-28
feat: updates UserAuthorizer to support retrieving token response directly with different client auth types

lsirac:
Adds support for the following in UserAuthorizer:
- Different client auth types
- A method to retrieve the token response directly, not UserCredentials, allowing for it to be used in more ways (e.g. enables Workforce Identity Federation)
- Constructor now takes in the Builder directly


aeitzman (approved):

oauth2_http/java/com/google/auth/oauth2/OAuth2Utils.java
```diff
@@ -76,9 +77,16 @@ class OAuth2Utils {
   static final String GRANT_TYPE_JWT_BEARER = "urn:ietf:params:oauth:grant-type:jwt-bearer";
 
   static final URI TOKEN_SERVER_URI = URI.create("https://oauth2.googleapis.com/token");
+
+  static final URI WORKFORCE_IDENTITY_FEDERATION_TOKEN_SERVER_URI =
+      URI.create("https://sts.googleapis.com/v1/oauthtoken");
```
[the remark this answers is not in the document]
lsirac: Nope, only used in tests. Moved them out to test utils though.

oauth2_http/java/com/google/auth/oauth2/OAuth2Utils.java
```diff
@@ -257,5 +259,26 @@ static PrivateKey privateKeyFromPkcs8(String privateKeyPkcs8) throws IOException
...
+   * password. The credentials are encoded in Base64 format and prefixed with the "Basic " scheme
+   * identifier.
+   *
+   * @param username The username for authentication.
+   * @param password The password for authentication.
+   * @return The Basic Authentication header value.
+   * @throws IllegalArgumentException if either username or password is null or empty.
+   */
+  static String generateBasicAuthHeader(String username, String password) {
+    if (username == null || username.isEmpty() || password == null || password.isEmpty()) {
```
lqiu96, line 275: nit: Perhaps we can use Guava's [isNullOrEmpty](https://guava.dev/releases/32.1.1-jre/api/docs/com/google/common/base/Strings.html#isNullOrEmpty(java.lang.String)) for username and password

oauth2_http/java/com/google/auth/oauth2/UserAuthorizer.java
```diff
@@ -285,61 +295,36 @@ public UserCredentials getCredentialsFromCode(String code, URI baseUri) throws I
...
+   * @param code The authorization code received from the OAuth2 authorization server.
+   * @param callbackUri The URI to which the authorization server redirected the user after granting
+   *     authorization.
+   * @param additionalParameters Additional parameters to include in the token exchange request.
+   * @return A {@link TokenResponseWithConfig} object containing the access token, refresh token (if
+   *     granted), and configuration details used in the OAuth flow.
+   * @throws IOException If an error occurs during the token exchange process.
+   */
+  public TokenResponseWithConfig getTokenResponseFromAuthCodeExchange(
+      String code, URI callbackUri, @Nullable Map<String, String> additionalParameters)
```
lqiu96, line 323: qq, can we make additionalParameters required to be non-null? If there are no additionalParameters needed, the user can just pass in an empty map. We wouldn't need the null check below in `getCredentialsFromCodeInternal` for additionalParameters

oauth2_http/java/com/google/auth/oauth2/UserAuthorizer.java
```diff
@@ -52,10 +52,22 @@
...
 
+  /**
+   * Represents the client authentication types as specified in RFC 7591.
+   *
+   * <p>For more details, see <a href="https://tools.ietf.org/html/rfc7591">RFC 7591</a>.
+   */
+  public enum ClientAuthenticationType {
+    CLIENT_SECRET_POST,
+    CLIENT_SECRET_BASIC,
+    NONE
```
sai-sunder-s, line 68: None is not used?

sai-sunder-s (approved):

oauth2_http/java/com/google/auth/oauth2/UserAuthorizer.java
lsirac: I can remove `@Nullable` but I can't remove the null check. This is existing code that is used by getCredentialsFromCode. Removing the null check would be a breaking change

oauth2_http/java/com/google/auth/oauth2/UserAuthorizer.java
```diff
@@ -285,61 +295,35 @@ public UserCredentials getCredentialsFromCode(String code, URI baseUri) throws I
...
+   *
+   * @param code The authorization code received from the OAuth2 authorization server.
+   * @param callbackUri The URI to which the authorization server redirected the user after granting
+   *     authorization.
+   * @param additionalParameters Additional parameters to include in the token exchange request.
+   * @return A {@link TokenResponseWithConfig} object containing the access token, refresh token (if
+   *     granted), and configuration details used in the OAuth flow.
+   * @throws IOException If an error occurs during the token exchange process.
+   */
+  public TokenResponseWithConfig getTokenResponseFromAuthCodeExchange(
```
lqiu96, line 322: qq, is it possible to have this return `UserCredentials` instead of `TokenResponseWithConfig`? I see from the above method that the TokenResponseWithConfig fields are mapped to a UserCredentials and wonder if the same can be done here.

I believe TokenResponseWithConfig is only exposed here and I would prefer not to expose a new class if possible

lqiu96 (approved):

lqiu96:
Thanks @lsirac, LGTM. Added some last comment if you could take a look as well.

oauth2_http/java/com/google/auth/oauth2/UserAuthorizer.java
lsirac: Unfortunately it is not, as UserCredentials is specific to Google and has several methods that would not work. Specifically did it like this to enable support for Workforce Identity Federation.

PR #1489, opened 2024-08-29
fix: Invalidate the SA's AccessToken when createScoped() is called

lqiu96:
See b/353616562 for more information. This seems to be a very similar issue to https://github.com/googleapis/google-auth-library-java/issues/1387.

Invalidating the Access Token via `setAccessToken(null)` from `createScoped(...)` should not break any existing workflows. Copying the existing Access Token while modifying the interval fields that make up the Credentials is a regression in behavior that was introduced in v1.14.0 of the Auth-Library.

Example flow for the current state (pseudocode):
```
Credentials credentials = GoogleCredentials.getADC();
Storage.create(..., credentials) // Access Token is retrieved with the scopes from ADC
System.out.println(credentials.getAccessToken()) // This is a valid access token with the ADC's scopes

credentials = credentials.createScoped("{NEW SCOPES}");
Storage.update(..., credentials) // Access Token is cached and re-used, but the scopes are different
System.out.println(credentials.getAccessToken()) // This is an invalid access token as it still has the ADC's scopes
```

Without this change, users would be recieving a 403 Access Denied error message. This change to invalidate the access token would force the auth library to retrieve a new valid access token.

Example flow for the new state (pseudocode):
```
Credentials credentials = GoogleCredentials.getADC();
Storage.create(..., credentials) // Access Token is retrieved with the scopes from ADC
System.out.println(credentials.getAccessToken()) // This is a valid access token with the ADC's scopes

credentials = credentials.createScoped("{NEW SCOPES}");
Storage.update(..., credentials) // Access Token is retrieved with the new scopes
System.out.println(credentials.getAccessToken()) // This is an valid access token as it has the new scopes
```


oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -612,14 +612,15 @@ public GoogleCredentials createScoped(Collection<String> newScopes) {
   }
 
   /**
-   * Clones the service account with the specified scopes.
+   * Clones the service account with the specified scopes. The Access Token is invalidated even if
```
suztomo, line 615: When this draft is ready for review, would you do:

1. Would you add a sentence about the purpose of invalidate access token?
2. In the pull request description, would you add a rationale `setAccessToken(null)` does not break existing code that use the `createScoped` method?

lqiu96: Thanks, added.

oauth2_http/javatests/com/google/auth/oauth2/ServiceAccountCredentialsTest.java
```diff
@@ -1705,6 +1705,37 @@ public void onFailure(Throwable exception) {
...
+            .setPrivateKeyId(PRIVATE_KEY_ID)
+            .setProjectId(PROJECT_ID)
+            .setQuotaProjectId("my-quota-project-id")
+            .setHttpTransportFactory(transportFactory)
+            .setUseJwtAccessWithScope(true)
+            .setScopes(SCOPES)
+            .build();
+    // Explicitly call `refresh()` to simulate an API request. Creating a GoogleCredentials object
+    // does not immediately get the Access Token
+    credentials.refresh();
```
zhumin8, line 1727: nit: since you are setting up mock transport, perhaps call `credentials.getRequestMetadata(CALL_URI)`  instead of calling `credentials.refresh()` which user won't need to call directly. Similar calls in tests [here](https://github.com/googleapis/google-auth-library-java/blob/f07d19a505e0ff8c43c3366bcda6350df6dcb3bc/oauth2_http/javatests/com/google/auth/oauth2/ServiceAccountCredentialsTest.java#L635).

zhumin8 (approved):
LGTM. 
Just a small comment with the test.

oauth2_http/javatests/com/google/auth/oauth2/ServiceAccountCredentialsTest.java
lqiu96: Thanks, updated.

PR #1503, opened 2024-09-13
feat: add metric headers

zhumin8:
context: b/339259830 and [go/send-auth-metrics-java](http://goto.google.com/send-auth-metrics-java)

Changes include:
- expose `Credentials` type via `getMetricsCredentialType()`. Override this method for UserCredentials, ServiceAccountCredentials, ImpersonatedCredentials, and ComputeEngineCredentials. This is used in both token request and token usage flows.
- add metric headers for each of the in-scope token requests. Below are examples of each request flow with added metrics:
  - User credentials request (at/id):  “gl-java/19.0.1 auth/1.24.3 cred-type/u”
  - SA credentials, VM credentials or Impersonated credentials requests (at/id): “gl-java/19.0.1 auth/1.24.3 auth-request-type/at cred-type/sa”
  - MDS ping (This is used in ADC during the credential detection): “gl-java/19.0.1 auth/1.24.3 auth-request-type/mds”
- What is not tracked: ComputeEngineCredentials getUniverseDomain and getAccount does not send metrics header; TPC flows does not send metrics headers.

Related pr: adding for cred_type for token usage requests
https://github.com/googleapis/sdk-platform-java/pull/3186

### DIFF
--- a/credentials/java/com/google/auth/CredentialTypeForMetrics.java
+++ b/credentials/java/com/google/auth/CredentialTypeForMetrics.java
@@ -31,13 +31,26 @@
 
 package com.google.auth;
 
+/**
+ * Defines the different types of credentials that can be used for metrics.
+ *
+ * <p>Each credential type is associated with a label that is used for reporting purposes. Add new
+ * enum constant only when corresponding configs established.
+ *
+ * <p>Credentials with type {@code CredentialTypeForMetrics.DO_NOT_SEND} is default value for
+ * credential implementations that do not set type specifically. It is not expected to send metrics.
+ *
+ * <p>
+ *
+ * @see #getLabel()
+ */
 public enum CredentialTypeForMetrics {
   USER_CREDENTIALS("u"),
   SERVICE_ACCOUNT_CREDENTIALS_AT("sa"),
   SERVICE_ACCOUNT_CREDENTIALS_JWT("jwt"),
   VM_CREDENTIALS("mds"),
   IMPERSONATED_CREDENTIALS("imp"),
-  UNKNOWN("unknown");
+  DO_NOT_SEND("do not send type to metrics");
 
   private String label;
 
--- a/credentials/java/com/google/auth/CredentialTypeForMetrics.java
+++ b/credentials/java/com/google/auth/CredentialTypeForMetrics.java
@@ -31,7 +31,7 @@
 
 package com.google.auth;
 
-public enum CredentialType {
+public enum CredentialTypeForMetrics {
   USER_CREDENTIALS("u"),
   SERVICE_ACCOUNT_CREDENTIALS_AT("sa"),
   SERVICE_ACCOUNT_CREDENTIALS_JWT("jwt"),
@@ -41,7 +41,7 @@ public enum CredentialType {
 
   private String label;
 
-  private CredentialType(String label) {
+  private CredentialTypeForMetrics(String label) {
     this.label = label;
   }
 
--- a/credentials/java/com/google/auth/CredentialTypeForMetrics.java
+++ b/credentials/java/com/google/auth/CredentialTypeForMetrics.java
@@ -50,7 +50,7 @@ public enum CredentialTypeForMetrics {
   SERVICE_ACCOUNT_CREDENTIALS_JWT("jwt"),
   VM_CREDENTIALS("mds"),
   IMPERSONATED_CREDENTIALS("imp"),
-  DO_NOT_SEND("do not send type to metrics");
+  DO_NOT_SEND("dns");
 
   private String label;
 
--- a/credentials/java/com/google/auth/Credentials.java
+++ b/credentials/java/com/google/auth/Credentials.java
@@ -45,7 +45,7 @@ public abstract class Credentials implements Serializable {
 
   public static final String GOOGLE_DEFAULT_UNIVERSE = "googleapis.com";
 
-  private CredentialType credentialType = CredentialType.UNKNOWN;
+  private CredentialTypeForMetrics credentialTypeForMetrics = CredentialTypeForMetrics.UNKNOWN;
 
   /**
    * A constant string name describing the authentication technology.
@@ -72,12 +72,12 @@ public abstract class Credentials implements Serializable {
     return GOOGLE_DEFAULT_UNIVERSE;
   }
 
-  public CredentialType getCredentialType() {
-    return this.credentialType;
+  public CredentialTypeForMetrics getCredentialType() {
+    return this.credentialTypeForMetrics;
   }
 
-  public void setCredentialType(CredentialType credentialType) {
-    this.credentialType = credentialType;
+  public void setCredentialType(CredentialTypeForMetrics credentialTypeForMetrics) {
+    this.credentialTypeForMetrics = credentialTypeForMetrics;
   }
 
   /**
--- a/credentials/java/com/google/auth/Credentials.java
+++ b/credentials/java/com/google/auth/Credentials.java
@@ -45,7 +45,7 @@ public abstract class Credentials implements Serializable {
 
   public static final String GOOGLE_DEFAULT_UNIVERSE = "googleapis.com";
 
-  private String credentialType = "unknown";
+  private CredentialType credentialType = CredentialType.UNKNOWN;
 
   /**
    * A constant string name describing the authentication technology.
@@ -72,11 +72,11 @@ public abstract class Credentials implements Serializable {
     return GOOGLE_DEFAULT_UNIVERSE;
   }
 
-  public String getCredentialType() {
+  public CredentialType getCredentialType() {
     return this.credentialType;
   }
 
-  public void setCredentialType(String credentialType) {
+  public void setCredentialType(CredentialType credentialType) {
     this.credentialType = credentialType;
   }
 
--- a/credentials/java/com/google/auth/Credentials.java
+++ b/credentials/java/com/google/auth/Credentials.java
@@ -45,7 +45,7 @@ public abstract class Credentials implements Serializable {
 
   public static final String GOOGLE_DEFAULT_UNIVERSE = "googleapis.com";
 
-  private CredentialTypeForMetrics credentialTypeForMetrics = CredentialTypeForMetrics.UNKNOWN;
+  private CredentialTypeForMetrics credentialTypeForMetrics = CredentialTypeForMetrics.DO_NOT_SEND;
 
   /**
    * A constant string name describing the authentication technology.
@@ -72,11 +72,24 @@ public abstract class Credentials implements Serializable {
     return GOOGLE_DEFAULT_UNIVERSE;
   }
 
-  public CredentialTypeForMetrics getCredentialType() {
+  /**
+   * Gets the credential type used for internal metrics header.
+   *
+   * @return a enum value for credential type
+   */
+  public CredentialTypeForMetrics getMetricsCredentialType() {
     return this.credentialTypeForMetrics;
   }
 
-  public void setCredentialType(CredentialTypeForMetrics credentialTypeForMetrics) {
+  /**
+   * Sets the credential type for metrics.
+   *
+   * <p>The default is {@code CredentialTypeForMetrics.DO_NOT_SEND}. For a credential that is
+   * established to track for metrics, this default should be overridden.
+   *
+   * @param credentialTypeForMetrics The credential type to be used for metrics.
+   */
+  public void setMetricsCredentialType(CredentialTypeForMetrics credentialTypeForMetrics) {
     this.credentialTypeForMetrics = credentialTypeForMetrics;
   }
 
--- a/credentials/java/com/google/auth/Credentials.java
+++ b/credentials/java/com/google/auth/Credentials.java
@@ -45,6 +45,8 @@ public abstract class Credentials implements Serializable {
 
   public static final String GOOGLE_DEFAULT_UNIVERSE = "googleapis.com";
 
+  private String credentialType = "unknown";
+
   /**
    * A constant string name describing the authentication technology.
    *
@@ -68,6 +70,14 @@ public abstract class Credentials implements Serializable {
    */
   public String getUniverseDomain() throws IOException {
     return GOOGLE_DEFAULT_UNIVERSE;
+  }
+
+  public String getCredentialType() {
+    return this.credentialType;
+  }
+
+  public void setCredentialType(String credentialType) {
+    this.credentialType = credentialType;
   }
 
   /**
--- a/credentials/java/com/google/auth/Credentials.java
+++ b/credentials/java/com/google/auth/Credentials.java
@@ -72,16 +72,15 @@ public abstract class Credentials implements Serializable {
 
   /**
    * Gets the credential type used for internal metrics header.
-    *
-    * <p>The default is {@code CredentialTypeForMetrics.DO_NOT_SEND}. For a credential that is
-    * established to track for metrics, this default should be overridden.
+   *
+   * <p>The default is {@code CredentialTypeForMetrics.DO_NOT_SEND}. For a credential that is
+   * established to track for metrics, this default should be overridden.
    *
    * @return a enum value for credential type
    */
   public CredentialTypeForMetrics getMetricsCredentialType() {
     return CredentialTypeForMetrics.DO_NOT_SEND;
   }
-
 
   /**
    * Get the current request metadata, refreshing tokens if required.
--- a/credentials/java/com/google/auth/Credentials.java
+++ b/credentials/java/com/google/auth/Credentials.java
@@ -45,8 +45,6 @@ public abstract class Credentials implements Serializable {
 
   public static final String GOOGLE_DEFAULT_UNIVERSE = "googleapis.com";
 
-  private CredentialTypeForMetrics credentialTypeForMetrics = CredentialTypeForMetrics.DO_NOT_SEND;
-
   /**
    * A constant string name describing the authentication technology.
    *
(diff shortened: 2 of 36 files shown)


credentials/java/com/google/auth/Credentials.java
```diff
@@ -45,6 +45,8 @@ public abstract class Credentials implements Serializable {
 
   public static final String GOOGLE_DEFAULT_UNIVERSE = "googleapis.com";
 
+  private CredentialType credentialType = CredentialType.UNKNOWN;
```
lqiu96, line 48: Just to confirm: Unknown is default for the other Credential types that aren't currently on our list, right?

credentials/java/com/google/auth/Credentials.java
```diff
@@ -70,6 +72,14 @@ public String getUniverseDomain() throws IOException {
     return GOOGLE_DEFAULT_UNIVERSE;
   }
 
+  public CredentialType getCredentialType() {
+    return this.credentialType;
+  }
+
+  public void setCredentialType(CredentialType credentialType) {
+    this.credentialType = credentialType;
+  }
```
lqiu96, line 81: Can you add javadocs for these public methods?

lqiu96, line 81: Hmm, I wish we could make the setter package-private but it looks like it's in a different module and it needs to be public.

I guess the other option would be to expose this CredentialType stuff only for Oauth2Credentials. That probably leaves a bunch of classes like ApiKeys needing to hold their own implementation.

I think it's probably fine. Maybe we can find another way...

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -334,13 +338,22 @@ public IdToken idTokenWithAudience(String targetAudience, List<IdTokenProvider.O
...
-  private HttpResponse getMetadataResponse(String url) throws IOException {
+  private HttpResponse getMetadataResponse(String url, RequestType requestType) throws IOException {
     GenericUrl genericUrl = new GenericUrl(url);
     HttpRequest request =
         transportFactory.create().createRequestFactory().buildGetRequest(genericUrl);
     JsonObjectParser parser = new JsonObjectParser(OAuth2Utils.JSON_FACTORY);
     request.setParser(parser);
     request.getHeaders().set(METADATA_FLAVOR, GOOGLE);
+    // do not send metric header for getUniverseDomain and getAccount
+    if (requestType != RequestType.UNSPECIFIED) {
```
lqiu96, line 349: qq, I see that UNSPECIFIED is also being used to gate some logic in `getGoogleCredentialsMetricsHeader`. Are both unspecified parts doing the same thing?

From a glance, it looks like they're doing two different things (one is not adding a value to the header and this one is not sending the metric headers at all). Thoughts on adding a different header like `RequestType.UNTRACKED` for this use case or maybe having a helper function `shouldSendMetricHeader()`?

oauth2_http/java/com/google/auth/oauth2/MetricsUtils.java
```diff
@@ -67,4 +68,44 @@ private static String getAuthLibraryVersion() {
...
+    private RequestType(String label) {
+      this.label = label;
+    }
+
+    public String getLabel() {
+      return label;
+    }
+  }
+
+  static String getGoogleCredentialsMetricsHeader(
```
lqiu96, line 89: Could you a small section highlighting the difference between RequestType.UNSPECIFIED, CredentialType.UNKNOWN, and the normal flow?

What happens when both RequestType.UNSPECIFIED and CredentialType.UNKNOWN come in? Or is that not possible? If not, do we want to check that and throw and error?

oauth2_http/java/com/google/auth/oauth2/MetricsUtils.java
```diff
@@ -67,4 +68,44 @@ private static String getAuthLibraryVersion() {
...
+      return String.format(
+          "%s %s/%s",
+          MetricsUtils.getLanguageAndAuthLibraryVersions(),
+          "auth-request-type",
+          requestType.getLabel());
+    }
+    return String.format(
+        "%s %s/%s %s/%s",
+        MetricsUtils.getLanguageAndAuthLibraryVersions(),
+        "auth-request-type",
```
lqiu96, line 106: nit: Maybe `auth-request-type` and `cred-type` as package-private constants? It looks mostly to be contained, but may be useful in tests.

oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java
```diff
@@ -508,6 +511,12 @@ public AccessToken refreshAccessToken() throws IOException {
     HttpRequest request = requestFactory.buildPostRequest(url, requestContent);
     adapter.initialize(request);
     request.setParser(parser);
+    request
+        .getHeaders()
+        .set(
+            MetricsUtils.API_CLIENT_HEADER,
+            MetricsUtils.getGoogleCredentialsMetricsHeader(
+                RequestType.ACCESS_TOKEN_REQUEST, getCredentialType()));
```
lqiu96, line 519: nit: Maybe we can extract setting the metrics header out to a common helper function? If it makes sense, it could be something like:
```
MetricsUtils.setMetricHeader(request, {header-value})
```

Might be easier to add a test for. I'm also fine with this.

oauth2_http/java/com/google/auth/oauth2/MetricsUtils.java
zhumin8: You are right that "both RequestType.UNSPECIFIED and CredentialType.UNKNOWN" is not currently intended for auth metrics. However, I am a bit hesitant to throw error for this utility methods, as it would need signature changes to methods sending the requests (IOException that these requests methods throws is not appropriate).
I am adding a case for this and simply sending language&lib versions. This is not actually used now, but does not hurt either.
Also simplified the logic to make it a bit more clear.

credentials/java/com/google/auth/CredentialTypeForMetrics.java
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * @see #getLabel()
+ */
+public enum CredentialTypeForMetrics {
+  USER_CREDENTIALS("u"),
+  SERVICE_ACCOUNT_CREDENTIALS_AT("sa"),
+  SERVICE_ACCOUNT_CREDENTIALS_JWT("jwt"),
+  VM_CREDENTIALS("mds"),
+  IMPERSONATED_CREDENTIALS("imp"),
+  DO_NOT_SEND("do not send type to metrics");
```
lqiu96, line 53: nit: How about `dns` to match the others?

credentials/java/com/google/auth/Credentials.java
zhumin8: Yes. And to make it clear, updated this default to "DO_NOT_SEND"

oauth2_http/java/com/google/auth/oauth2/MetricsUtils.java
lqiu96: We could throw something like a RunTimeException (possibly a IllegalStateException). It's just that it's possible for customers to configure the CredentialType because we have a public setter on the Credentials and currently can't hide it.

I think the worst case is that our metrics may not e 100% accurate if users tinker with it. It probably would be better than crashing the application so I think I'm ok with not throwing anything.

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -334,13 +338,20 @@ public IdToken idTokenWithAudience(String targetAudience, List<IdTokenProvider.O
...
-  private HttpResponse getMetadataResponse(String url) throws IOException {
+  private HttpResponse getMetadataResponse(String url, RequestType requestType) throws IOException {
     GenericUrl genericUrl = new GenericUrl(url);
     HttpRequest request =
         transportFactory.create().createRequestFactory().buildGetRequest(genericUrl);
     JsonObjectParser parser = new JsonObjectParser(OAuth2Utils.JSON_FACTORY);
     request.setParser(parser);
     request.getHeaders().set(METADATA_FLAVOR, GOOGLE);
+    // do not send metric header for getUniverseDomain and getAccount
+    if (requestType != RequestType.UNTRACKED) {
```
lqiu96, line 349: Just to confirm: this is not sending *any* header. So we also won't be sending `MetricsUtils.getLanguageAndAuthLibraryVersions()`. This is intended right?

oauth2_http/javatests/com/google/auth/oauth2/MockMetadataServerTransport.java
```diff
@@ -101,31 +103,41 @@ public void setIdToken(String idToken) {
...
+      this.request = getMockRequestForServiceAccount(url);
+      return this.request;
     } else if (isSignRequestUrl(url)) {
-      return getMockRequestForSign(url);
+      this.request = getMockRequestForSign(url);
+      return this.request;
     } else if (isIdentityDocumentUrl(url)) {
-      return getMockRequestForIdentityDocument(url);
+      this.request = getMockRequestForIdentityDocument(url);
+      return this.request;
```
lqiu96, line 123: why is this being assigned to request before returning?

oauth2_http/javatests/com/google/auth/oauth2/TestUtils.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.List;
+import java.util.Map;
+
+public class TestUtils {
+  static void validateMetricsHeader(
+      Map<String, List<String>> headers, String requestType, String credentialType) {
+    assertTrue(headers.containsKey(MetricsUtils.API_CLIENT_HEADER));
+    String actualMetricsValue = headers.get(MetricsUtils.API_CLIENT_HEADER).get(0);
+    String expectedMetricsValue;
+    if (requestType.equals("unspecified")) {
```
lqiu96, line 46: nit: it was changed to `untracked`, right?

lqiu96: Oh wait nvm.

oauth2_http/java/com/google/auth/oauth2/MetricsUtils.java
```diff
@@ -67,4 +71,50 @@ private static String getAuthLibraryVersion() {
     }
     return version;
   }
+
+  public enum RequestType {
+    ACCESS_TOKEN_REQUEST("at"),
+    ID_TOKEN_REQUEST("it"),
+    METADATA_SERVER_PIN("mds"),
+    UNSPECIFIED("unspecified"),
+    UNTRACKED("untracked");
```
lqiu96, line 80: Can you add a comment about the difference between UNSPECIFIED and UNTRACKED for future development? 

If I'm understanding it correctly, `UNSPECIFIED` is only omit the  `auth-request-type/REQUEST_TYPE` portion, but `UNTRACKED` omits the entire header?

oauth2_http/javatests/com/google/auth/oauth2/MockMetadataServerTransport.java
zhumin8: The request is assigned in this mock transport so that it can be exposed with `getRequest()`, so that headers are easily accessible in tests. Similarly to MockTokenServerTransport [here](https://github.com/googleapis/google-auth-library-java/blob/16510064e861868f649b6bc8fdc54b8a39890812/oauth2_http/javatests/com/google/auth/oauth2/MockTokenServerTransport.java#L179-L194).

credentials/java/com/google/auth/Credentials.java
```diff
@@ -70,6 +72,27 @@ public String getUniverseDomain() throws IOException {
...
+   *
+   * <p>The default is {@code CredentialTypeForMetrics.DO_NOT_SEND}. For a credential that is
+   * established to track for metrics, this default should be overridden.
+   *
+   * @param credentialTypeForMetrics The credential type to be used for metrics.
+   */
+  public void setMetricsCredentialType(CredentialTypeForMetrics credentialTypeForMetrics) {
+    this.credentialTypeForMetrics = credentialTypeForMetrics;
+  }
+
```
lqiu96, line 95: On second thought, I think it may be possible to remove `setMetricsCredentialType()`? I'd rather not add a public setter and I think we can do it this way:

Credentials' `getMetricsCredentialType()` returns `DO_NOT_SEND` by default:
```
public CredentialTypeForMetrics getMetricsCredentialType() {
    return CredentialTypeForMetrics.DO_NOT_SEND;
}
```

Each subclass would override this value: 
i.e. ComputeEngineCredentials' `getMetricsCredentialType()`:
```
public CredentialTypeForMetrics getMetricsCredentialType() {
    return CredentialTypeForMetrics.VM_CREDENTIALS;
}
```

When we add a new Metrics for Credentials, we'll just need to override this value. Do you think there are any issues with this?

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
lqiu96: Ok gotcha, I think that makes sense. Thanks!

oauth2_http/java/com/google/auth/oauth2/MetricsUtils.java
```diff
@@ -67,4 +71,50 @@ private static String getAuthLibraryVersion() {
     }
     return version;
   }
+
+  public enum RequestType {
+    ACCESS_TOKEN_REQUEST("at"),
+    ID_TOKEN_REQUEST("it"),
+    METADATA_SERVER_PIN("mds"),
```
lqiu96, line 78: METADATA_SERVER_PING?

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -334,13 +338,20 @@ public IdToken idTokenWithAudience(String targetAudience, List<IdTokenProvider.O
     return IdToken.create(rawToken);
   }
 
-  private HttpResponse getMetadataResponse(String url) throws IOException {
+  private HttpResponse getMetadataResponse(String url, RequestType requestType) throws IOException {
```
lqiu96, line 341: I know I told you to swap from UNSPECIFIED to UNTRACKED, but from reading the comments below as additional context, I think my opinion has slightly changed and I'm flip-flopping (sorry!). Let me know what you think. 

To me, it seems a bit odd that CredentialsMetrics.DO_NOT_SEND does not send the CRED_TYPE and RequestType.UNSPECIFIED does not send the AUTH_REQUEST_TYPE, but RequestType.UNTRACKED doesn't send both (it doesn't send the header at all). I would think that CredentialMetrics only impacts CRED_TYPE and RequestType for AUTH_REQUEST_TYPE.

What if we change the calls to something like this:
`getMetadataResponse(buildHttpRequest(url, requestType, shouldAddMetricsHeader));`

We add a new helper function `buildHttpRequest(...)` that takes in an additional bool flag where `universe_domain` and `getAccount` calls set the flag to be false. The flag gates adding the MetricsHeader.

oauth2_http/java/com/google/auth/oauth2/MetricsUtils.java
```diff
@@ -67,4 +71,50 @@ private static String getAuthLibraryVersion() {
...
+    UNSPECIFIED("unspecified"), // should not send request type info
+    UNTRACKED("untracked"); // should not add metric header
```
lqiu96, line 80: nit: If the change above for ComputeEngineCredentials makes sense, UNTRACKED sounds more accurate to me. The RequestType isn't tracked and therefore isn't added. Would it be possible to swap UNSPECIFIED and UNTRACKED?

oauth2_http/javatests/com/google/auth/oauth2/MockMetadataServerTransport.java
lqiu96: Ok gotcha. I wish we could use like interceptors or something. Something to fix in the future for sure, but I think this is fine for now.

credentials/java/com/google/auth/Credentials.java
zhumin8: Thanks for pointing this out. In fact, I switched away from this approach assuming issues for SA with 2 flow types. But I just took another look and this should work fine.

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
zhumin8: I don't feel too strongly about this, will make the change.

oauth2_http/javatests/com/google/auth/oauth2/TestUtils.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.List;
+import java.util.Map;
+
+public class TestUtils {
+  static void validateMetricsHeader(
+      Map<String, List<String>> headers, String requestType, String credentialType) {
+    assertTrue(headers.containsKey(MetricsUtils.API_CLIENT_HEADER));
+    String actualMetricsValue = headers.get(MetricsUtils.API_CLIENT_HEADER).get(0);
+    String expectedMetricsValue;
+    if (requestType.equals("untracked")) {
```
lqiu96, line 46: nit: Can we use the enum value here?

lqiu96 (approved):
LGTM. Added two questions in the tests. Thanks for addressing the comments and questions!

oauth2_http/javatests/com/google/auth/oauth2/TestUtils.java
zhumin8: note `requestType` and `credentialType` here are String values. They represent part of the expected value in this utils method. I am leaning towards checking against plain text here

oauth2_http/javatests/com/google/auth/oauth2/TestUtils.java
```diff
@@ -0,0 +1,72 @@
...
+      expectedMetricsValue =
+          String.format(
+              "%s %s/%s",
+              MetricsUtils.getLanguageAndAuthLibraryVersions(),
+              MetricsUtils.CRED_TYPE,
+              credentialType);
+    } else if (credentialType.isEmpty()) {
```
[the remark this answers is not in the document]
zhumin8: updating to be consistent with `requestType` above.

PR #1528, opened 2024-10-04
feat: service sccount to service account impersonation to support universe domain

zhumin8:
for context: b/340602527

Changes in this pr:
- Override `getUniverseDomain()` to grab source credentials’s universe domain (UD) by default. Always use source credentials UD, not explicit provided UD. (In current design, impersonated credentials may not have universe domain in the outer layer. relay on UD from source credential. This may change in future)
- Fix `isDefaultUniverseDomain()` in `GoogleCredentials` to account for `getUniverseDomain()` overrides in child classes.
- In refreshAccessToken(), use endpoint url pattern to account for TPC case.
  - note that I choose to bypass this refreshIfExpired step because it wrongly steps into code path meant only for OAuth2 token request (GDU flow). Filed #1534 to address this separately. But for GDU flow here, this refresh step is redundant because the SSJ will get re-generated at [initialize request](https://github.com/googleapis/google-auth-library-java/blob/a987ecd06fd25a0048cdb3da6d1df4d029d85d79/oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java#L558). Also skip this step for SA GDU with SSJ flow.
- Throw IllegalStateException if UD is explicitly set (with parent class setter) and not matching source credential's UD

- Fix toBuilder() to invoke super, and fix related issue with createScoped. (see #1489, #1428); Also fix equals() to compare super first.

Not in this pr: 
- idtoken and signBlob endpoint changes are out-of-scope for this pr, will raise separate pr for it.

sa-to-sa impersonation is successfully E2E tested for TPC usage according to [go/prptst-testing-service-account-impersonation](http://goto.google.com/prptst-testing-service-account-impersonation).

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
@@ -263,7 +263,7 @@ public class GoogleCredentials extends OAuth2Credentials implements QuotaProject
   /**
    * Checks if universe domain equals to {@link Credentials#GOOGLE_DEFAULT_UNIVERSE}.
    *
-   * @return true if universeDomain equals to {@link Credentials#GOOGLE_DEFAULT_UNIVERSE}, false
+   * @return true if universe domain equals to {@link Credentials#GOOGLE_DEFAULT_UNIVERSE}, false
    *     otherwise
    */
   boolean isDefaultUniverseDomain() throws IOException {
--- a/oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
@@ -266,8 +266,8 @@ public class GoogleCredentials extends OAuth2Credentials implements QuotaProject
    * @return true if universeDomain equals to {@link Credentials#GOOGLE_DEFAULT_UNIVERSE}, false
    *     otherwise
    */
-  boolean isDefaultUniverseDomain() {
-    return this.universeDomain.equals(Credentials.GOOGLE_DEFAULT_UNIVERSE);
+  boolean isDefaultUniverseDomain() throws IOException {
+    return getUniverseDomain().equals(Credentials.GOOGLE_DEFAULT_UNIVERSE);
   }
 
   /**
--- a/oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
@@ -483,8 +483,7 @@ public class GoogleCredentials extends OAuth2Credentials implements QuotaProject
     return this;
   }
 
-  public static class Builder<BuilderType extends Builder<BuilderType>>
-      extends OAuth2Credentials.Builder {
+  public static class Builder extends OAuth2Credentials.Builder {
     @Nullable protected String quotaProjectId;
     @Nullable protected String universeDomain;
 
@@ -510,14 +509,14 @@ public class GoogleCredentials extends OAuth2Credentials implements QuotaProject
     }
 
     @CanIgnoreReturnValue
-    public BuilderType setQuotaProjectId(String quotaProjectId) {
+    public Builder setQuotaProjectId(String quotaProjectId) {
       this.quotaProjectId = quotaProjectId;
-      return (BuilderType) this;
+      return this;
     }
 
-    public BuilderType setUniverseDomain(String universeDomain) {
+    public Builder setUniverseDomain(String universeDomain) {
       this.universeDomain = universeDomain;
-      return (BuilderType) this;
+      return this;
     }
 
     public String getQuotaProjectId() {
@@ -530,9 +529,9 @@ public class GoogleCredentials extends OAuth2Credentials implements QuotaProject
 
     @Override
     @CanIgnoreReturnValue
-    public BuilderType setAccessToken(AccessToken token) {
+    public Builder setAccessToken(AccessToken token) {
       super.setAccessToken(token);
-      return (BuilderType) this;
+      return this;
     }
   }
 }
--- a/oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
@@ -483,7 +483,8 @@ public class GoogleCredentials extends OAuth2Credentials implements QuotaProject
     return this;
   }
 
-  public static class Builder extends OAuth2Credentials.Builder {
+  public static class Builder<BuilderType extends Builder<BuilderType>>
+      extends OAuth2Credentials.Builder {
     @Nullable protected String quotaProjectId;
     @Nullable protected String universeDomain;
 
@@ -509,14 +510,14 @@ public class GoogleCredentials extends OAuth2Credentials implements QuotaProject
     }
 
     @CanIgnoreReturnValue
-    public Builder setQuotaProjectId(String quotaProjectId) {
+    public BuilderType setQuotaProjectId(String quotaProjectId) {
       this.quotaProjectId = quotaProjectId;
-      return this;
+      return (BuilderType) this;
     }
 
-    public Builder setUniverseDomain(String universeDomain) {
+    public BuilderType setUniverseDomain(String universeDomain) {
       this.universeDomain = universeDomain;
-      return this;
+      return (BuilderType) this;
     }
 
     public String getQuotaProjectId() {
@@ -529,9 +530,9 @@ public class GoogleCredentials extends OAuth2Credentials implements QuotaProject
 
     @Override
     @CanIgnoreReturnValue
-    public Builder setAccessToken(AccessToken token) {
+    public BuilderType setAccessToken(AccessToken token) {
       super.setAccessToken(token);
-      return this;
+      return (BuilderType) this;
     }
   }
 }
--- a/oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java
@@ -484,8 +484,8 @@ public class ImpersonatedCredentials extends GoogleCredentials
   /**
    * Gets the universe domain for the credential.
    *
-   * @return An explicit universe domain if it was explicitly provided, invokes the super
-   *     implementation otherwise
+   * @return An explicit universe domain if it was explicitly provided, use the universe domain from
+   *     source credentials otherwise
    */
   @Override
   public String getUniverseDomain() throws IOException {
@@ -495,6 +495,12 @@ public class ImpersonatedCredentials extends GoogleCredentials
     return this.sourceCredentials.getUniverseDomain();
   }
 
+  /**
+   * Checks if universe domain equals to {@link Credentials#GOOGLE_DEFAULT_UNIVERSE}.
+   *
+   * @return true if universeDomain equals to {@link Credentials#GOOGLE_DEFAULT_UNIVERSE}, false
+   *     otherwise
+   */
   @Override
   boolean isDefaultUniverseDomain() {
     try {
@@ -513,7 +519,8 @@ public class ImpersonatedCredentials extends GoogleCredentials
           this.sourceCredentials.createScoped(Arrays.asList(CLOUD_PLATFORM_SCOPE));
     }
 
-    // for nonGDU uses self-signed JWT and will get refreshed at initialize request step
+    // skip for nonGDU because it uses self-signed JWT
+    // and will get refreshed at initialize request step
     if (isDefaultUniverseDomain()) {
       try {
         this.sourceCredentials.refreshIfExpired();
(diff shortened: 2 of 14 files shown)


oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java
```diff
@@ -430,6 +428,7 @@ public GoogleCredentials createScoped(Collection<String> scopes) {
         .setHttpTransportFactory(this.transportFactory)
         .setQuotaProjectId(this.quotaProjectId)
         .setIamEndpointOverride(this.iamEndpointOverride)
+        .setAccessToken(null)
```
blakeli0, line 431: Why do we have to explicitly set it to null?
Also thanks to the newly introduced constructor below `protected Builder(ImpersonatedCredentials credentials)`, I think this can be simplified to 
```
toBuilder()
        .setScopes(new ArrayList<>(scopes))
        .build()
```

oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java
```diff
@@ -501,10 +535,18 @@ public AccessToken refreshAccessToken() throws IOException {
...
+    try {
+      endpointUrl =
+          this.iamEndpointOverride != null
+              ? this.iamEndpointOverride
+              : String.format(
+                  OAuth2Utils.IAM_ACCESS_TOKEN_ENDPOINT_FORMAT,
+                  getUniverseDomain(),
+                  this.targetPrincipal);
+    } catch (IOException e) {
+      throw new IOException("Error getting Universe Domain", e);
```
blakeli0, line 548: This re-throw does not seem to provide much additional value, can we just let the `IOException` bubble up without wrapping and rethrowing?

oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java
```diff
@@ -482,17 +481,52 @@ private ImpersonatedCredentials(Builder builder) {
...
       this.sourceCredentials =
           this.sourceCredentials.createScoped(Arrays.asList(CLOUD_PLATFORM_SCOPE));
     }
 
-    try {
-      this.sourceCredentials.refreshIfExpired();
-    } catch (IOException e) {
-      throw new IOException("Unable to refresh sourceCredentials", e);
+    // skip for nonGDU because it uses self-signed JWT
+    // and will get refreshed at initialize request step
```
blakeli0, line 523: I'm still not sure why this call is required for GDU but not required for nonGDU, is there any feature that is only supported in GDU?

oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java
```diff
@@ -482,17 +481,52 @@ private ImpersonatedCredentials(Builder builder) {
...
 
+  /**
+   * Gets the universe domain for the credential.
+   *
+   * @return An explicit universe domain if it was explicitly provided, use the universe domain from
+   *     source credentials otherwise
+   */
+  @Override
+  public String getUniverseDomain() throws IOException {
+    if (isExplicitUniverseDomain()) {
```
blakeli0, line 492: This is for the scenario that customers may create an `ImpersonatedCredentials` object and set `universeDomain` explicitly?

oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java
```diff
@@ -482,17 +481,52 @@ private ImpersonatedCredentials(Builder builder) {
...
+  }
+
+  /**
+   * Checks if universe domain equals to {@link Credentials#GOOGLE_DEFAULT_UNIVERSE}.
+   *
+   * @return true if universeDomain equals to {@link Credentials#GOOGLE_DEFAULT_UNIVERSE}, false
+   *     otherwise
+   */
+  @Override
+  boolean isDefaultUniverseDomain() {
```
blakeli0, line 505: We should probably change the super class's [implementation](https://github.com/googleapis/google-auth-library-java/blob/89201558db913d9a71b3acccbab8eb0045ada6de/oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java#L270) to what you have here.
From 
```
this.universeDomain.equals(Credentials.GOOGLE_DEFAULT_UNIVERSE)
```
to
```
getUniverseDomain().equals(Credentials.GOOGLE_DEFAULT_UNIVERSE)
```
so that we don't have to duplicate the effort here. Make sure it does not break other scenarios though.

oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java
```diff
@@ -724,6 +787,12 @@ public Calendar getCalendar() {
       return this.calendar;
     }
 
+    @Override
+    public Builder setUniverseDomain(String universeDomain) {
```
blakeli0, line 791: I don't think we have to override this method as it is already public in the parent. I noticed that it seems we are overriding the builder methods in other subclasses of the builder, but I think that was a mistake.

oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java
zhumin8: > Why do we have to explicitly set it to null?

This is invoking the parent class builder which sets: 
https://github.com/googleapis/google-auth-library-java/blob/89201558db913d9a71b3acccbab8eb0045ada6de/oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java#L648
We have #1428, #1489 fixing similar situations in other credential classes.

You are right about the other setters.

oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java
zhumin8: I am convinced that this is not needed for nonGDU, reasons in description above.
For GDU, I am not 100% convinced that this is required before initialize step. But out of precautious, I do not want to change the flow for GDU in this pr. I will raise a separate issue to investigate for GDU flow.

oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java
zhumin8: Yes.

oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java
zhumin8: I debated this in my mind. One benefit I see overriding the setter allows for a fluent builder API where method calls can be chained without requiring explicit casting to ImpersonatedCredentials.  In this case, I feel it provides a slightly better experience for users setting explicit UD
So user can do 
```
        ImpersonatedCredentials.newBuilder()
            .setUniverseDomain("explicit.domain.com");
```
instead of casting:
```
        (ImpersonatedCredentials) ImpersonatedCredentials.newBuilder()
            .setUniverseDomain("explicit.domain.com");
```
While generally it's best to avoid overriding public methods if possible, override setters in Builders is kind of a special case and more acceptable. .e.g. overriding Builder setters seems to be a pattern in other libraries like Guava ([example](https://github.com/google/guava/blob/d40589de41baa3b2ce8442b24366bc57339b9871/guava/src/com/google/common/collect/ImmutableList.java#L840))

One other concern is removing other overrides in other credential classes will be breaking change to customers code. 
That said, I don't feel too strongly about this, if you have any other concerns, I can remove this. What do you think?

oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java
zhumin8: Good point! Let me try it out.

zhumin8: I checked and this should not conflict with other use cases, this `isDefaultUniverseDomain()` is only used here and from ServiceAccountCredentials. In fact, I think use getUniverseDomain() inside it should be the expected behavior, otherwise it won't accurately reflect status for subclasses. e.g. ComputeEngineCredentials which also have [custom logic](https://github.com/googleapis/google-auth-library-java/blob/89201558db913d9a71b3acccbab8eb0045ada6de/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java#L224) for getting universe domain.

I am a bit hesitant on how to deal with the exception situation here. @blakeli0 Do you have any suggestions?

In parent class GoogleCredentials,
```
 // throwing exception in isDefaultUniverseDomain would be breakiing change
  boolean isDefaultUniverseDomain() {
    try {
     // getUniverseDomain() throws exception
      return getUniverseDomain().equals(Credentials.GOOGLE_DEFAULT_UNIVERSE);
    } catch (IOException ignored) {
         logger.warn("Error getting universe domain: " + e.getMessage(), e);
         return false; // this might be misleading
    }
  }
```

blakeli0: Throwing exception in `isDefaultUniverseDomain` is not a breaking change because it is package private. I think it should be fine.

zhumin8: ah- somehow missed this. Thank you.

oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java
blakeli0: I don't think customers have to cast the credential to `ImpersonatedCredentials` to `setUniverseDomain()`, I verified in my local that it works. Can you pleas double check it?
For the Guava example, yes they are doing it because there are some additional logics, but we are not doing anything special other than calling `super.setUniverseDomain()`. So I think we can delete it at this moment and add it later if needed.

oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java
blakeli0: > This is invoking the parent class builder
Ah, makes sense, thanks! Basically we want to invalidate existing tokens when creating a new type of credential.

oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java
TimurSadykov: as comment states - nonGDU is always SSJ, no token endpoint, no refresh call

TimurSadykov: to answer Blake's question - GDU has token endpoint, nonGDU does not :) 

SSJ is a primary flow is GDU now as well. This should be a fallback scenario if SSJ is disabled, execute conditionally if SSJ flag is disabled.

oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java
TimurSadykov: Confirming that explicit overrides were added to avoid casting.

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -992,13 +992,19 @@ public void getRequestMetadata(
...
+      if (this.useJwtAccessWithScope || !isDefaultUniverseDomain()) {
+        // This will call getRequestMetadata(URI uri), which handles self-signed JWT logic.
+        // Self-signed JWT doesn't use network, so here we do a blocking call to improve
+        // efficiency. executor will be ignored since it is intended for async operation.
+        blockingGetToCallback(uri, callback);
+      } else {
+        super.getRequestMetadata(uri, executor, callback);
+      }
+    } catch (IOException e) {
+      // Wrap it here to avoid breaking change.
```
TimurSadykov, line 1005: What would be a breaking change?

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -266,8 +266,8 @@ protected boolean isExplicitUniverseDomain() {
    * @return true if universeDomain equals to {@link Credentials#GOOGLE_DEFAULT_UNIVERSE}, false
```
TimurSadykov, line 266: update docs?

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
zhumin8: added to comment, to throw IOException would be breaking.

oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java
blakeli0: Min and I confirmed that we do need explicit casting right now. However, the problem is that the parent builder's setters should return the child types. I created a [demo PR](https://github.com/googleapis/google-auth-library-java/pull/1536) for it .  Also see how protobuf's [GeneratedMessageV3](https://github.com/protocolbuffers/protobuf/blob/main/java/core/src/main/java/com/google/protobuf/GeneratedMessageV3.java#L158-L159) does it.

That being said, this should be separate problem to solve, so it's OK to override the parent's setters right now.

blakeli0 (approved):

zhumin8:
@blakeli0 @TimurSadykov Can you please take another look? Changes since you last reviewed:
Updated with changes to reflect internal discussion to throw for explicit set universe domain not matching source credential's ud.
Also updated condition to skip `refreshIfExpired()` to include gdu w/ ssj flow. 
Added tests for these changes.

oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java
zhumin8: updated condition to also skip for GDU SSJ flow.

PR #1558, opened 2024-10-26
feat: Support transport and binding-enforcement MDS parameters.

rmehta19:


rmehta19:
cc: @rockspore

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -696,6 +725,14 @@ public Collection<String> getDefaultScopes() {
       return defaultScopes;
     }
 
+    public String getTransport() {
```
westarle, line 728: Please provide javadoc on these, pointing to publicly documented Metadata Service parameters.

rmehta19: Discussed offline: We will update public MDS docs when the feature is non-experimental. Since the usage of the S2A + MTLS + bound tokens flow is guarded by an environment variable (EXPERIMENTAL_GOOGLE_API_USE_S2A). Until the public docs are updated, we can link to AIP here and https://github.com/googleapis/google-auth-library-java/pull/1400 (AIP already linked in this merged PR).

rmehta19:
@lqiu96 , @zhumin8 : Would be able to take a look at this change and leave any feedback? Thanks!

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -200,6 +208,12 @@ String createTokenUrlWithScopes() {
     if (!scopes.isEmpty()) {
       tokenUrl.set("scopes", Joiner.on(',').join(scopes));
     }
+    if (!transport.isEmpty()) {
+      tokenUrl.set("transport", transport);
+    }
+    if (!bindingEnforcement.isEmpty()) {
+      tokenUrl.set("binding-enforcement", bindingEnforcement);
+    }
```
lqiu96, line 216: I think this should be null and empty check. Setter can set a null value unless we want to add a check there

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -643,6 +657,9 @@ public static class Builder extends GoogleCredentials.Builder {
     private Collection<String> scopes;
     private Collection<String> defaultScopes;
 
+    private String transport = "";
+    private String bindingEnforcement = "";
```
lqiu96, line 661: nit:
```suggestion
    private String transport;
    private String bindingEnforcement;
```

lqiu96:
Update to use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/):
`feat: Support transport and binding-enforcement MDS parameters`

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
rmehta19: Thanks for pointing that out, done in https://github.com/googleapis/google-auth-library-java/pull/1558/commits/9d46bbd880b6240bb7a828510254b075a60c7188

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
rmehta19: Done in https://github.com/googleapis/google-auth-library-java/pull/1558/commits/3cb6e9852e77c77944d1418eec03bb5fdd970760

rmehta19:
Thanks for the review @lqiu96 !

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -191,7 +197,10 @@ public final Collection<String> getScopes() {
   }
 
   /**
-   * If scopes is specified, add "?scopes=comma-separated-list-of-scopes" to the token url.
+   * If scopes is specified, add "?scopes=comma-separated-list-of-scopes" to the token url. If
+   * transport is specified, add "?transport=xyz" to the token url; xyz is one of "alts" or "mtls".
+   * If bindingEnforcement is specified, add "?binding-enforcement=xyz" to the token url; xyz is one
+   * of "iam-policy" or "on".
```
lqiu96, line 203: If these have a limited set of options, would it be possible we have a Transport Enum and a BindingEnforcement enum?

The transport setter would take in a set of Transport enums to use.

rmehta19: > If these have a limited set of options
Yes there are a limited set of options. Transport can be "alts" or "mtls". BindingEnforcement can be "on" or "iam-policy"

Thanks for suggesting this @lqiu96. This is something that we can do. It would require exposing a public `Transport` enum and `BindingEnforcement` enum so that the logic in sdk-platform-java can set these values. Are you ok with adding these public enums to the auth library?

lqiu96: > It would require exposing a public Transport enum and BindingEnforcement enum so that the logic in sdk-platform-java can set these values. Are you ok with adding these public enums to the auth library?

Ah I missed this. I think that's fine and would be better to cover the combination of transport and bindingEnforcement configurations. The invalid configurations between the two enums can be handled by the server side logic, but at least this would help users choose the correct configurations.

Just to confirm, these enums aren't expect to change much (or at all)

rmehta19: >  I think that's fine and would be better to cover the combination of transport and bindingEnforcement configurations. The invalid configurations between the two enums can be handled by the server side logic, but at least this would help users choose the correct configurations.

SG

> Just to confirm, these enums aren't expect to change much (or at all)

Yes, we don't expect them to change, as the design has been finalized and approved by all stakeholders, and implementation has begun. When this feature is non-experimental, the enums would be confirmed to not change.

@lqiu96 in the latest commit, I added these enums.

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -200,6 +218,16 @@ String createTokenUrlWithScopes() {
     if (!scopes.isEmpty()) {
       tokenUrl.set("scopes", Joiner.on(',').join(scopes));
     }
+    if (transport == Transport.MTLS) {
+      tokenUrl.set("transport", "mtls");
```
blakeli0, line 222: You can add fields to the Enum class, so that each Enum can have its own label, then you can just do `tokenUrl.set("transport", transport.getLabel())`. See this [example](https://www.baeldung.com/java-enum-values#adding-constructor).

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -109,6 +109,16 @@ public class ComputeEngineCredentials extends GoogleCredentials
   static final int MAX_COMPUTE_PING_TRIES = 3;
   static final int COMPUTE_PING_CONNECTION_TIMEOUT_MS = 500;
 
+  public enum Transport {
```
blakeli0, line 112: I know the it is called `transport` in the token url, but since Transport in GAPIC has different meaning (the value of it is either gRPC or httpJson), can we name this differently? Maybe `AuthTransport`? Javadocs for what ALTS and MTLS mean would also be appreciated.

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
rmehta19: Thanks for pointing this out! Done in latest commit.

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
rmehta19: Renamed to AuthTransport in latest commit and added some javadocs. Thanks!

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -109,6 +109,40 @@ public class ComputeEngineCredentials extends GoogleCredentials
...
+    public String getLabel() {
+      return label;
+    }
+  }
+
+  public enum BindingEnforcement {
+    // Binding enforcement will always happen, irrespective of the IAM policy.
+    ON("on"),
+    // Binding enforcement will depend on IAM policy.
+    IAMPOLICY("iam-policy");
```
lqiu96, line 133:
```suggestion
    IAM_POLICY("iam-policy");
```

I believe the convention is underscore to separate the words for constants

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -684,6 +735,18 @@ public Builder setQuotaProjectId(String quotaProjectId) {
...
+    public Builder setTransport(AuthTransport transport) {
+      this.transport = transport;
+      return this;
+    }
+
+    @CanIgnoreReturnValue
+    public Builder setBindingEnforcement(BindingEnforcement bindingEnforcement) {
+      this.bindingEnforcement = bindingEnforcement;
+      return this;
+    }
```
lqiu96, line 748: can you some small/ simple javadocs for these public setters?

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -696,6 +759,14 @@ public Collection<String> getDefaultScopes() {
...
     }
 
+    public AuthTransport getTransport() {
+      return transport;
+    }
+
+    public BindingEnforcement getBindingEnforcement() {
+      return bindingEnforcement;
+    }
+
```
lqiu96, line 769: same here. also if it's returning AuthTransport, can it be `getAuthTransport()`?

oauth2_http/javatests/com/google/auth/oauth2/ComputeEngineCredentialsTest.java
```diff
@@ -190,6 +190,98 @@ public void buildTokenUrlWithScopes_defaultScopes() {
     assertEquals("bar", scopes.toArray()[1]);
   }
 
+  @Test
+  public void buildTokenUrl_nullTransport() {
```
lqiu96, line 194: nit: one additional test where both transport and bindingenforcements are null

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -109,6 +109,40 @@ public class ComputeEngineCredentials extends GoogleCredentials
   static final int MAX_COMPUTE_PING_TRIES = 3;
   static final int COMPUTE_PING_CONNECTION_TIMEOUT_MS = 500;
 
+  public enum AuthTransport {
```
lqiu96, line 112: I think AuthTransport is ok for this. I would prefer if this could be more specific if possible.

Is there a term for alts and mtls type of transport that we can use? Does `TlsTransport` make sense? I don't know if TLS has a transport concept or if that would be confusing.

`GoogleMtlsTransport`?

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
rmehta19: Done in https://github.com/googleapis/google-auth-library-java/pull/1558/commits/f36a19cd4e94898e896950304296160e2a88d685

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
rmehta19: Done in https://github.com/googleapis/google-auth-library-java/pull/1558/commits/8a2d967552c6322f7b6514dcde4b7ba190a7128a

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
rmehta19: Done in https://github.com/googleapis/google-auth-library-java/pull/1558/commits/8a2d967552c6322f7b6514dcde4b7ba190a7128a

oauth2_http/javatests/com/google/auth/oauth2/ComputeEngineCredentialsTest.java
rmehta19: Done in https://github.com/googleapis/google-auth-library-java/pull/1558/commits/ce74e45f2cb2fba09486fe4faccac933950ebd09

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
rmehta19: TLDR: WDYT of `GoogleAuthTransport`?

> I think AuthTransport is ok for this. I would prefer if this could be more specific if possible.

We generally call this "transport type": describes how the client is authenticating to Google APIS either via GFE (using `mtls`) or via DirectPath (using `alts`). However Blake mentioned above that transport has a different meaning in Client libraries (grpc vs http), so it would be good to clarify to something more specific, which is how we landed on `AuthTransport` initially.

> Is there a term for alts and mtls type of transport that we can use? Does TlsTransport make sense? I don't know if TLS has a transport concept or if that would be confusing.

TLS is the security protocol being used, it's used in both mtls and alts. mtls is mutually authenticated TLS, alts is (in short) "Google-optimized" mtls (see [public docs](https://cloud.google.com/docs/security/encryption-in-transit/application-layer-transport-security)).

In the enum name, I think we want to focus on how clients are authenticating to Google APIS, not details of the protocol being used (mtls vs alts). How about `GoogleAuthTransport`? I think this captures that the enum is listing different ways to ways clients can authenticate to Google APIs.

> GoogleMtlsTransport? Though I'm not sure mtls/atls are google specific terms.
I think this name would be ok. One thing is that mtls via GFE strictly adheres to the TLS standard (i.e. not Google specific). Whereas alts is a [Google specific protocol](https://cloud.google.com/docs/security/encryption-in-transit/application-layer-transport-security).

I've done this in https://github.com/googleapis/google-auth-library-java/pull/1558/commits/8ce3a4d6782d58b13a90fbf7899b28a02419adf6, let me know if the name sounds ok.

rmehta19:
Friendly ping @lqiu96, @blakeli0. Please let me know if there's anything else to address on this PR.

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -109,6 +109,40 @@ public class ComputeEngineCredentials extends GoogleCredentials
   static final int MAX_COMPUTE_PING_TRIES = 3;
   static final int COMPUTE_PING_CONNECTION_TIMEOUT_MS = 500;
 
+  public enum GoogleAuthTransport {
```
lqiu96, line 112: Since GoogleAuthTransport and BindingEnforcement live in ComputeEngineCredentials and not any S2A specific classes, can they have some simple top level javadocs? Ideally if there is public docs that we can link to (this can also be added in the future). 

Would be helpful to answer the combo of when/why users can/ should use them?

lqiu96 (approved):
Added a nit small request for a bit more javadocs on public enums, lgtm otherwise.

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
rmehta19: > can they have some simple top level javadocs?

Done

> Ideally if there is public docs that we can link to (this can also be added in the future).

When the feature (S2A + MTLS + bound tokens) is non-experimental, we will be able to link to public MDS docs. For now, I have marked in the javadocs: "Experimental Feature".

> Would be helpful to answer the combo of when/why users can/ should use them? Also to explain what Google Auth Transport is in this case

Done. we can modify these javadocs again when the feature is non-experimental (public MDS docs are ready).

rmehta19:
> Added a nit small request for a bit more javadocs on public enums, lgtm otherwise.

Thanks @lqiu96 ! Addressed. Would you be able to merge if the changes look good to you?

lqiu96:
> > Added a nit small request for a bit more javadocs on public enums, lgtm otherwise.
> 
> Thanks @lqiu96 ! Addressed. Would you be able to merge if the changes look good to you?

Yep lgtm, feel free to merge

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -109,6 +109,57 @@ public class ComputeEngineCredentials extends GoogleCredentials
   static final int MAX_COMPUTE_PING_TRIES = 3;
   static final int COMPUTE_PING_CONNECTION_TIMEOUT_MS = 500;
 
+  // Experimental Feature:
+  // Behavior of setting {@link GoogleAuthTransport} / {@link BindingEnforcement}:
+  // - MTLS-bound token where binding enforcement depends on IAM policy: MTLS / {}, {} / IAM_POLICY,
+  // MTLS / IAM_POLICY
+  // - MTLS-bound token where bindings are always enforced: {} / ON, MTLS / ON
+  // - DirectPath bound token: ALTS / {}
```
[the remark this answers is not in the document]
rmehta19: Done.

rmehta19:
> Yep lgtm, feel free to merge

Thanks @lqiu96 ! I don't have write permissions on this repo; would you be able to merge?

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -109,6 +109,62 @@ public class ComputeEngineCredentials extends GoogleCredentials
   static final int MAX_COMPUTE_PING_TRIES = 3;
   static final int COMPUTE_PING_CONNECTION_TIMEOUT_MS = 500;
 
+  // Experimental Feature:
+  //
```
lqiu96, line 113: can you remove this?

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -109,6 +109,62 @@ public class ComputeEngineCredentials extends GoogleCredentials
...
+   * <p>{@link GoogleAuthTransport} specifies how to authenticate to Google APIs.
+   *
+   * <p>Behavior of setting {@link GoogleAuthTransport} / {@link BindingEnforcement}:
+   *
+   * <p>MTLS-bound token where binding enforcement depends on IAM policy: MTLS / {}, {} /
+   * IAM_POLICY, MTLS / IAM_POLICY
+   *
+   * <p>MTLS-bound token where bindings are always enforced: {} / ON, MTLS / ON
+   *
+   * <p>DirectPath bound token: ALTS / {}
```
lqiu96, line 127: Since this also shares logic with BindingEnforcement, thoughts on duplicating this in the javadocs for BindingsEnforcement as well?

lqiu96: Or if there is a better way to split it up, I'm also fine with that

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
rmehta19: Thanks for catching! removed.

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
rmehta19: Yes, duplicating this in `BindingEnforcement` javadocs SGTM. done.

PR #1562, opened 2024-10-30
feat: Implement ClientSideCredentialAccessBoundaryFactory

nbayati:
See: [go/client-side-cab-client-library-java](http://goto.google.com/client-side-cab-client-library-java)
This PR creates the ClientSideCredentialAccessBoundaryFactory class, and implements the function to fetch the intermediary token and session key from the STS endpoint. It completes the item `Fetching and parsing intermediate tokens` in the execution plan. [go/offline-cab-sdk-execution-plan](http://goto.google.com/offline-cab-sdk-execution-plan)

### DIFF
--- a/cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
+++ b/cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
@@ -60,7 +60,7 @@ public final class ClientSideCredentialAccessBoundaryFactory {
     this.tokenExchangeEndpoint = builder.tokenExchangeEndpoint;
   }
 
-  private void fetchCredentials() throws IOException {
+  private void refreshIntermediateCredentials() throws IOException {
     try {
       this.sourceCredential.refreshIfExpired();
     } catch (IOException e) {
--- a/cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
+++ b/cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
@@ -38,6 +38,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import com.google.auth.Credentials;
 import com.google.auth.http.HttpTransportFactory;
 import com.google.auth.oauth2.AccessToken;
+import com.google.auth.oauth2.CredentialAccessBoundary;
 import com.google.auth.oauth2.GoogleCredentials;
 import com.google.auth.oauth2.OAuth2Utils;
 import com.google.auth.oauth2.StsRequestHandler;
@@ -60,7 +61,7 @@ public final class ClientSideCredentialAccessBoundaryFactory {
     this.tokenExchangeEndpoint = builder.tokenExchangeEndpoint;
   }
 
-  private void refreshIntermediateCredentials() throws IOException {
+  private void refreshCredentials() throws IOException {
     try {
       this.sourceCredential.refreshIfExpired();
     } catch (IOException e) {
@@ -100,6 +101,16 @@ public final class ClientSideCredentialAccessBoundaryFactory {
     }
   }
 
+  private void refreshCredentialsIfRequired() {
+    // TODO(negarb): Implement refreshCredentialsIfRequired
+    throw new UnsupportedOperationException("refreshCredentialsIfRequired is not yet implemented.");
+  }
+
+  public AccessToken generateToken(CredentialAccessBoundary accessBoundary) {
+    // TODO(negarb/jiahuah): Implement generateToken
+    throw new UnsupportedOperationException("generateToken is not yet implemented.");
+  }
+
   public static Builder newBuilder() {
     return new Builder();
   }
@@ -113,11 +124,12 @@ public final class ClientSideCredentialAccessBoundaryFactory {
     private Builder() {}
 
     /**
-     * Sets the required source credential used to acquire the intermediary credential.
+     * Sets the required source credential.
      *
      * @param sourceCredential the {@code GoogleCredentials} to set
      * @return this {@code Builder} object
      */
+    @CanIgnoreReturnValue
     public Builder setSourceCredential(GoogleCredentials sourceCredential) {
       this.sourceCredential = sourceCredential;
       return this;
--- a/cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
+++ b/cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2024 Google LLC
+ * Copyright 2024, Google LLC
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions are
--- a/cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
+++ b/cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
@@ -47,7 +47,6 @@ import com.google.common.base.Strings;
 import com.google.errorprone.annotations.CanIgnoreReturnValue;
 import java.io.IOException;
 
-
 public final class ClientSideCredentialAccessBoundaryFactory {
   private final GoogleCredentials sourceCredential;
   private final transient HttpTransportFactory transportFactory;
@@ -77,7 +76,6 @@ public final class ClientSideCredentialAccessBoundaryFactory {
         StsTokenExchangeRequest.newBuilder(
                 sourceAccessToken.getTokenValue(), OAuth2Utils.TOKEN_TYPE_ACCESS_TOKEN)
             .setRequestTokenType(OAuth2Utils.TOKEN_TYPE_ACCESS_BOUNDARY_INTERMEDIARY_TOKEN)
-
             .build();
 
     StsRequestHandler handler =
--- a/cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
+++ b/cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
@@ -47,12 +47,13 @@ import com.google.common.base.Strings;
 import com.google.errorprone.annotations.CanIgnoreReturnValue;
 import java.io.IOException;
 
+
 public final class ClientSideCredentialAccessBoundaryFactory {
   private final GoogleCredentials sourceCredential;
   private final transient HttpTransportFactory transportFactory;
   private final String tokenExchangeEndpoint;
-  private String acceessBoundarySessionKey;
-  private AccessToken intermediaryAccessToken;
+  private String accessBoundarySessionKey;
+  private AccessToken intermediateAccessToken;
 
   private ClientSideCredentialAccessBoundaryFactory(Builder builder) {
     this.transportFactory = builder.transportFactory;
@@ -76,6 +77,7 @@ public final class ClientSideCredentialAccessBoundaryFactory {
         StsTokenExchangeRequest.newBuilder(
                 sourceAccessToken.getTokenValue(), OAuth2Utils.TOKEN_TYPE_ACCESS_TOKEN)
             .setRequestTokenType(OAuth2Utils.TOKEN_TYPE_ACCESS_BOUNDARY_INTERMEDIARY_TOKEN)
+
             .build();
 
     StsRequestHandler handler =
@@ -84,8 +86,8 @@ public final class ClientSideCredentialAccessBoundaryFactory {
             .build();
 
     StsTokenExchangeResponse response = handler.exchangeToken();
-    this.acceessBoundarySessionKey = response.getAccessBoundarySessionKey();
-    this.intermediaryAccessToken = response.getAccessToken();
+    this.accessBoundarySessionKey = response.getAccessBoundarySessionKey();
+    this.intermediateAccessToken = response.getAccessToken();
 
     // The STS endpoint will only return the expiration time for the intermediary token
     // if the original access token represents a service account.
@@ -93,7 +95,7 @@ public final class ClientSideCredentialAccessBoundaryFactory {
     // When no expires_in is returned, we can copy the source credential's expiration time.
     if (response.getAccessToken().getExpirationTime() == null) {
       if (sourceAccessToken.getExpirationTime() != null) {
-        this.intermediaryAccessToken =
+        this.intermediateAccessToken =
             new AccessToken(
                 response.getAccessToken().getTokenValue(), sourceAccessToken.getExpirationTime());
       }
--- a/cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
+++ b/cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
@@ -61,6 +61,22 @@ public final class ClientSideCredentialAccessBoundaryFactory {
     this.tokenExchangeEndpoint = builder.tokenExchangeEndpoint;
   }
 
+  /**
+   * Refreshes the source credential and exchanges it for an intermediary access token using the STS
+   * endpoint.
+   *
+   * <p>If the source credential is expired, it will be refreshed. A token exchange request is then
+   * made to the STS endpoint. The resulting intermediary access token and access boundary session
+   * key are stored. The intermediary access token's expiration time is determined as follows:
+   *
+   * <ol>
+   *   <li>If the STS response includes `expires_in`, that value is used.
+   *   <li>Otherwise, if the source credential has an expiration time, that value is used.
+   *   <li>Otherwise, the intermediary token will have no expiration time.
+   * </ol>
+   *
+   * @throws IOException If an error occurs during credential refresh or token exchange.
+   */
   private void refreshCredentials() throws IOException {
     try {
       this.sourceCredential.refreshIfExpired();
@@ -70,7 +86,7 @@ public final class ClientSideCredentialAccessBoundaryFactory {
 
     AccessToken sourceAccessToken = sourceCredential.getAccessToken();
     if (sourceAccessToken == null || Strings.isNullOrEmpty(sourceAccessToken.getTokenValue())) {
-      throw new IOException("The source credential does not have an access token.");
+      throw new IllegalStateException("The source credential does not have an access token.");
     }
 
     StsTokenExchangeRequest request =
--- a/cab-token-generator/pom.xml
+++ b/cab-token-generator/pom.xml
@@ -20,27 +20,22 @@
     <dependency>
       <groupId>com.google.auth</groupId>
       <artifactId>google-auth-library-oauth2-http</artifactId>
-      <version>${project.version}</version>
     </dependency>
     <dependency>
       <groupId>com.google.auth</groupId>
       <artifactId>google-auth-library-credentials</artifactId>
-      <version>1.29.1-SNAPSHOT</version>
     </dependency>
     <dependency>
       <groupId>com.google.http-client</groupId>
       <artifactId>google-http-client</artifactId>
-      <version>1.45.0</version>
     </dependency>
     <dependency>
       <groupId>com.google.errorprone</groupId>
       <artifactId>error_prone_annotations</artifactId>
-      <version>2.35.1</version>
     </dependency>
     <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
-      <version>33.3.1-android</version>
     </dependency>
   </dependencies>
 
--- a/cab-token-generator/pom.xml
+++ b/cab-token-generator/pom.xml
@@ -16,6 +16,7 @@
     <dependency>
       <groupId>com.google.auth</groupId>
       <artifactId>google-auth-library-oauth2-http</artifactId>
+      <version>${project.version}</version>
     </dependency>
   </dependencies>
 
--- a/cab-token-generator/pom.xml
+++ b/cab-token-generator/pom.xml
@@ -6,17 +6,41 @@
   <parent>
     <groupId>com.google.auth</groupId>
     <artifactId>google-auth-library-parent</artifactId>
-    <version>1.29.1-SNAPSHOT</version><!-- {x-version-update:google-auth-library-parent:current} -->
+    <version>1.29.1-SNAPSHOT
+    </version><!-- {x-version-update:google-auth-library-parent:current} -->
   </parent>
 
   <artifactId>cab-token-generator</artifactId>
   <name>Google Auth Library for Java - Cab Token Generator</name>
 
+  <build>
+    <sourceDirectory>java</sourceDirectory>
+  </build>
   <dependencies>
     <dependency>
       <groupId>com.google.auth</groupId>
       <artifactId>google-auth-library-oauth2-http</artifactId>
       <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>com.google.auth</groupId>
+      <artifactId>google-auth-library-credentials</artifactId>
+      <version>1.29.1-SNAPSHOT</version>
+    </dependency>
+    <dependency>
+      <groupId>com.google.http-client</groupId>
+      <artifactId>google-http-client</artifactId>
+      <version>1.45.0</version>
+    </dependency>
+    <dependency>
+      <groupId>com.google.errorprone</groupId>
+      <artifactId>error_prone_annotations</artifactId>
+      <version>2.35.1</version>
+    </dependency>
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+      <version>33.3.1-android</version>
     </dependency>
   </dependencies>
 
--- a/cab-token-generator/pom.xml
+++ b/cab-token-generator/pom.xml
@@ -6,16 +6,12 @@
   <parent>
     <groupId>com.google.auth</groupId>
     <artifactId>google-auth-library-parent</artifactId>
-    <version>1.29.1-SNAPSHOT</version>
+    <version>1.29.1-SNAPSHOT</version><!-- {x-version-update:google-auth-library-parent:current} -->
   </parent>
 
   <artifactId>cab-token-generator</artifactId>
+  <name>Google Auth Library for Java - Cab Token Generator</name>
 
-  <properties>
-    <maven.compiler.source>22</maven.compiler.source>
-    <maven.compiler.target>22</maven.compiler.target>
-    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-  </properties>
   <dependencies>
     <dependency>
       <groupId>com.google.auth</groupId>
--- a/cab-token-generator/pom.xml
+++ b/cab-token-generator/pom.xml
@@ -1,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>com.google.auth</groupId>
+    <artifactId>google-auth-library-parent</artifactId>
+    <version>1.29.1-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>cab-token-generator</artifactId>
+
+  <properties>
+    <maven.compiler.source>22</maven.compiler.source>
+    <maven.compiler.target>22</maven.compiler.target>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+  </properties>
+  <dependencies>
+    <dependency>
+      <groupId>com.google.auth</groupId>
+      <artifactId>google-auth-library-oauth2-http</artifactId>
+    </dependency>
+  </dependencies>
+
+</project>
--- a/cab-token-generator/src/main/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
+++ b/cab-token-generator/src/main/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
@@ -1,7 +1,38 @@
+/*
+ * Copyright 2021 Google LLC
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met:
+ *
+ *    * Redistributions of source code must retain the above copyright
+ * notice, this list of conditions and the following disclaimer.
+ *    * Redistributions in binary form must reproduce the above
+ * copyright notice, this list of conditions and the following disclaimer
+ * in the documentation and/or other materials provided with the
+ * distribution.
+ *
+ *    * Neither the name of Google LLC nor the names of its
+ * contributors may be used to endorse or promote products derived from
+ * this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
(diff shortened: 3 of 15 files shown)


oauth2_http/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -0,0 +1,147 @@
+package com.google.auth.credentialaccessboundary;
```
lqiu96, line 1: I believe this class is intended to go under the `cab-token-generator` module, not under the existing oauth2_http module, right? This looks like it's adding it to a different package under the same module.

oauth2_http/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -0,0 +1,147 @@
...
+import java.io.IOException;
+
+public final class ClientSideCredentialAccessBoundaryFactory {
+  private final GoogleCredentials sourceCredential;
+  private final transient HttpTransportFactory transportFactory;
+  private final String tokenExchangeEndpoint;
+  private String acceessBoundarySessionKey;
+  private AccessToken intermediaryAccessToken;
+
+  private ClientSideCredentialAccessBoundaryFactory(Builder builder) {
```
lqiu96, line 25: If possible, can we move the validation logic into the builder's build() method?

oauth2_http/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -0,0 +1,147 @@
...
+                + "credential.");
+      }
+    } catch (IOException e) {
+      // Throwing an IOException would be a breaking change, so wrap it here.
+      throw new IllegalStateException(
+          "Error occurred when attempting to retrieve source credential universe domain.", e);
+    }
+    String TOKEN_EXCHANGE_URL_FORMAT = "https://sts.{universe_domain}/v1/token";
+    this.tokenExchangeEndpoint =
+        TOKEN_EXCHANGE_URL_FORMAT.replace("{universe_domain}", universeDomain);
```
lqiu96, line 54: nit:
```suggestion
    String TOKEN_EXCHANGE_URL_FORMAT = "https://sts.%s/v1/token";
    this.tokenExchangeEndpoint = String.format(TOKEN_EXCHANGE_URL_FORMAT, universeDomain);
```
Can you check if `TOKEN_EXCHANGE_URL_FORMAT` exists in Oauth2Utils or make it constant in this class?

oauth2_http/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  public void fetchCredentials() throws IOException {
+    try {
+      this.sourceCredential.refreshIfExpired();
+    } catch (IOException e) {
+      throw new IOException("Unable to refresh the provided source credential.", e);
```
lqiu96, line 61: qq, why catch and rethrow this IOException? Is the message from the earlier IOException not detailed enough?

Also, can you add Javadocs for this public method?

oauth2_http/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    AccessToken sourceAccessToken = sourceCredential.getAccessToken();
+    if (sourceAccessToken == null || sourceAccessToken.getTokenValue() == null) {
```
lqiu96, line 65: nit: Can we use Guava's [isNullorEmpty](https://guava.dev/releases/31.1-jre/api/docs/com/google/common/base/Strings.html#isNullOrEmpty(java.lang.String))?

oauth2_http/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -0,0 +1,147 @@
...
+      // Throwing an IOException would be a breaking change, so wrap it here.
+      throw new IllegalStateException(
+          "Error occurred when attempting to retrieve source credential universe domain.", e);
+    }
+    String TOKEN_EXCHANGE_URL_FORMAT = "https://sts.{universe_domain}/v1/token";
+    this.tokenExchangeEndpoint =
+        TOKEN_EXCHANGE_URL_FORMAT.replace("{universe_domain}", universeDomain);
+  }
+
+  public void fetchCredentials() throws IOException {
```
lqiu96, line 57: If this is supposed to be public facing, can we add javadocs for this? Otherwise can this be package-private or private if it's intended for internal use?

oauth2_http/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
nbayati: Ah yes, you're right. I'll create a new module and move the code there.

oauth2_http/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
nbayati: Good catch. I believe this will be for internal use and we'll have another function wrapped around it to fetch and refresh credentials that will be public facing.

oauth2_http/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
nbayati: Good question. I saw this was the way it was handled in `DownscopedCredentials.java` and assumed that was necessary. @lsirac might know more about it.
 https://github.com/googleapis/google-auth-library-java/blob/7c5ed2fb2b1f6ddc74f3ec4873d2915f58d7128f/oauth2_http/java/com/google/auth/oauth2/DownscopedCredentials.java#L137

oauth2_http/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
nbayati: Sounds good, done!

cab-token-generator/pom.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>com.google.auth</groupId>
+    <artifactId>google-auth-library-parent</artifactId>
```
lqiu96, line 8: Can you add the release-please tag: `<!-- {x-version-update:google-auth-library-parent:current} -->` so this will be bumped?

cab-token-generator/pom.xml
```diff
@@ -0,0 +1,26 @@
...
+    <version>1.29.1-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>cab-token-generator</artifactId>
```
lqiu96, line 12: Add a <name> tag after this: `<name>Google Auth Library for Java - Cab Token Generator</name>`

cab-token-generator/pom.xml
```diff
@@ -0,0 +1,26 @@
...
+
+  <properties>
+    <maven.compiler.source>22</maven.compiler.source>
+    <maven.compiler.target>22</maven.compiler.target>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+  </properties>
```
lqiu96, line 18: I believe this is already mentioned in the parent pom and you don't need add it here.

cab-token-generator/src/main/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -0,0 +1,181 @@
+/*
+ * Copyright 2021 Google LLC
```
lqiu96, line 2:
```suggestion
 * Copyright 2024 Google LLC
```

oauth2_http/java/com/google/auth/oauth2/CredentialAccessBoundary.java
```diff
@@ -67,7 +67,7 @@ public final class CredentialAccessBoundary {
   /**
    * Internal method that returns the JSON string representation of the credential access boundary.
    */
-  String toJson() {
+  public String toJson() {
```
lqiu96, line 70: I don't think I see this being used anywhere. Does this need to change?

oauth2_http/java/com/google/auth/oauth2/StsTokenExchangeResponse.java
```diff
@@ -142,9 +151,22 @@ public StsTokenExchangeResponse.Builder setScopes(List<String> scopes) {
       return this;
     }
 
+    @CanIgnoreReturnValue
+    public StsTokenExchangeResponse.Builder setAccessBoundarySessionKey(
+        String accessBoundarySessionKey) {
+      this.accessBoundarySessionKey = accessBoundarySessionKey;
+      return this;
+    }
```
lqiu96, line 159: javadocs for the public method

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -0,0 +1,181 @@
...
+import com.google.auth.oauth2.GoogleCredentials;
+import com.google.auth.oauth2.OAuth2Utils;
+import com.google.auth.oauth2.StsRequestHandler;
+import com.google.auth.oauth2.StsTokenExchangeRequest;
+import com.google.auth.oauth2.StsTokenExchangeResponse;
+import com.google.common.base.Strings;
+import com.google.errorprone.annotations.CanIgnoreReturnValue;
+import java.io.IOException;
+
+public final class ClientSideCredentialAccessBoundaryFactory {
```
aeitzman, line 50: Nit: Javadoc comment for this class, I think it's fine to wait until a later PR though in case anything changes.

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -0,0 +1,181 @@
...
+  private final GoogleCredentials sourceCredential;
+  private final transient HttpTransportFactory transportFactory;
+  private final String tokenExchangeEndpoint;
+  private String acceessBoundarySessionKey;
```
aeitzman, line 54: typo - should be "access..." not "accees"

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -0,0 +1,181 @@
...
+  private AccessToken intermediaryAccessToken;
```
aeitzman, line 55: nit: intermediateAccessToken

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -0,0 +1,181 @@
...
+
+  private ClientSideCredentialAccessBoundaryFactory(Builder builder) {
+    this.transportFactory = builder.transportFactory;
+    this.sourceCredential = builder.sourceCredential;
+    this.tokenExchangeEndpoint = builder.tokenExchangeEndpoint;
+  }
+
+  private void fetchCredentials() throws IOException {
```
aeitzman, line 63: nit: refreshCredentials?

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
nbayati: Sounds good. I'll add the docs in a later PR.

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
aeitzman: there are two separate flows here right? There is a function that needs to refresh intermediate credentials (this one), and one that refresh a downscoped credential - I think those two could have different names but both still be refreshXYZCredentials() - i.e. call this "refreshIntermediateCredentials()". Fetch implies to me that should also return the credential, which this function doesn't

aeitzman (approved):

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -0,0 +1,181 @@
...
+  public static class Builder {
+    private GoogleCredentials sourceCredential;
+    private HttpTransportFactory transportFactory;
+    private String universeDomain;
+    private String tokenExchangeEndpoint;
+
+    private Builder() {}
+
+    /**
+     * Sets the required source credential used to acquire the intermediary credential.
```
lsirac, line 116: Do we need to expose that there is an intermediary token here? If we don't need to here / anywhere else, I would omit this to avoid confusion.

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -0,0 +1,181 @@
...
+     *
+     * @param sourceCredential the {@code GoogleCredentials} to set
+     * @return this {@code Builder} object
+     */
+    public Builder setSourceCredential(GoogleCredentials sourceCredential) {
```
lsirac, line 121: nit: Missing @CanIgnoreReturnValue
```suggestion
    @CanIgnoreReturnValue
    public Builder setSourceCredential(GoogleCredentials sourceCredential) {
```

cab-token-generator/pom.xml
```diff
@@ -0,0 +1,47 @@
...
+      <version>2.35.1</version>
+    </dependency>
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+      <version>33.3.1-android</version>
+    </dependency>
+  </dependencies>
+
+</project>
```
lsirac, line 47:
```suggestion
</project>

```

oauth2_http/java/com/google/auth/oauth2/OAuth2Utils.java
```diff
@@ -70,11 +70,15 @@
 import java.util.Set;
 
 /** Internal utilities for the com.google.auth.oauth2 namespace. */
-class OAuth2Utils {
+public class OAuth2Utils {
```
lsirac, line 73: Not sure if it makes sense for this to be public @lqiu96

oauth2_http/java/com/google/auth/oauth2/StsTokenExchangeResponse.java
```diff
@@ -40,25 +40,33 @@
 import javax.annotation.Nullable;
 
 /**
- * Defines an OAuth 2.0 token exchange successful response. Based on
- * https://tools.ietf.org/html/rfc8693#section-2.2.1.
+ * Represents an OAuth 2.0 token exchange request to the Google Security Token Service (STS), as
```
lsirac, line 43: This should be updated -- looks like its the Request docs

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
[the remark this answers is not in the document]
nbayati: I think I'll merge this PR without unit tests to unblock Jiahua's PR and add the unit tests in a future PR.

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
nbayati: Good point! I don't think we'll expose the intermediary token, I'll update the comment.

oauth2_http/java/com/google/auth/oauth2/StsTokenExchangeResponse.java
nbayati: Ah yes! 🤦  Thanks for catching this!

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -0,0 +1,193 @@
...
+    this.sourceCredential = builder.sourceCredential;
+    this.tokenExchangeEndpoint = builder.tokenExchangeEndpoint;
+  }
+
+  private void refreshCredentials() throws IOException {
+    try {
+      this.sourceCredential.refreshIfExpired();
+    } catch (IOException e) {
+      throw new IOException("Unable to refresh the provided source credential.", e);
+    }
```
lqiu96, line 69: Can you remind me why we are re-throwing this exception? If there is no reason, I think we should have `refreshIfExpired()` throw the exception instead of catching and then re-throw

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -0,0 +1,193 @@
...
+
+    AccessToken sourceAccessToken = sourceCredential.getAccessToken();
+    if (sourceAccessToken == null || Strings.isNullOrEmpty(sourceAccessToken.getTokenValue())) {
+      throw new IOException("The source credential does not have an access token.");
```
lqiu96, line 73: I think this probably should be an IllegalStateException rather than IOException. I don't think there is any IO here.

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -0,0 +1,193 @@
...
+
+    StsTokenExchangeResponse response = handler.exchangeToken();
+    this.accessBoundarySessionKey = response.getAccessBoundarySessionKey();
+    this.intermediateAccessToken = response.getAccessToken();
+
+    // The STS endpoint will only return the expiration time for the intermediary token
+    // if the original access token represents a service account.
+    // The intermediary token's expiration time will always match the source credential expiration.
+    // When no expires_in is returned, we can copy the source credential's expiration time.
+    if (response.getAccessToken().getExpirationTime() == null) {
```
lqiu96, line 95: qq, I believe above has a null check for `getAccessToken()`. Does there need to be a null check here as well?

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -0,0 +1,193 @@
...
+            .build();
+
+    StsTokenExchangeResponse response = handler.exchangeToken();
+    this.accessBoundarySessionKey = response.getAccessBoundarySessionKey();
+    this.intermediateAccessToken = response.getAccessToken();
+
+    // The STS endpoint will only return the expiration time for the intermediary token
+    // if the original access token represents a service account.
+    // The intermediary token's expiration time will always match the source credential expiration.
+    // When no expires_in is returned, we can copy the source credential's expiration time.
```
lqiu96, line 94: is `expires_in` mapped to `expirationTime`? If so, could we update the comment to make that clearer?

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -0,0 +1,193 @@
...
+  private String accessBoundarySessionKey;
+  private AccessToken intermediateAccessToken;
+
+  private ClientSideCredentialAccessBoundaryFactory(Builder builder) {
+    this.transportFactory = builder.transportFactory;
+    this.sourceCredential = builder.sourceCredential;
+    this.tokenExchangeEndpoint = builder.tokenExchangeEndpoint;
+  }
+
+  private void refreshCredentials() throws IOException {
```
lqiu96, line 64: Would you be able to write some quick comments that describes some of the behavior regarding the source access token and the response access token and how they form the intermediate access token?

From what I gather from the code, but it would be helpful for a comment to help confirm:
Intermediate access token is usually the STS response, unless the response doesn't have an expiration time. If not, then a new access token is created with the source access token's expiration time unless that value also doesn't exist.

What happens when both access tokens don't have an expiration time? Will this end up failing downstream? Will this just live forever?

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -0,0 +1,193 @@
...
+
+  private void refreshCredentialsIfRequired() {
+    // TODO(negarb): Implement refreshCredentialsIfRequired
+    throw new UnsupportedOperationException("refreshCredentialsIfRequired is not yet implemented.");
+  }
+
+  public AccessToken generateToken(CredentialAccessBoundary accessBoundary) {
+    // TODO(negarb/jiahuah): Implement generateToken
+    throw new UnsupportedOperationException("generateToken is not yet implemented.");
+  }
```
lqiu96, line 112: can we keep this private for now and update the visibility when needed? Just to give us flexibility to change the method name/ params if needed.

cab-token-generator/pom.xml
```diff
@@ -0,0 +1,47 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>com.google.auth</groupId>
+    <artifactId>google-auth-library-parent</artifactId>
+    <version>1.29.1-SNAPSHOT
+    </version><!-- {x-version-update:google-auth-library-parent:current} -->
```
lqiu96, line 10:
```suggestion
    <version>1.30.0</version><!-- {x-version-update:google-auth-library-parent:current} -->
```

This version is going to change as we cut new releases of auth. Last thing before we merge will be to update this to the latest version.

cab-token-generator/pom.xml
```diff
@@ -0,0 +1,47 @@
...
+    <dependency>
+      <groupId>com.google.errorprone</groupId>
+      <artifactId>error_prone_annotations</artifactId>
+      <version>2.35.1</version>
+    </dependency>
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+      <version>33.3.1-android</version>
+    </dependency>
```
lqiu96, line 44: I believe these versions can be managed by the parent pom. I think you can remove the versions here.

oauth2_http/java/com/google/auth/oauth2/OAuth2Utils.java
lqiu96: One option for this to see if we can keep this package-private and move some of these constants to STSRequestHandler. It looks like `TOKEN_TYPE_ACCESS_BOUNDARY_INTERMEDIARY_TOKEN` and `TOKEN_EXCHANGE_URL_FORMAT` are new additions could be live in the individual classes (i.e StsREquestHandler or DownscopedCredentials).

TOKEN_TYPE_ACCESS_TOKEN is used in a few places, but if it makes sense we can move them around. I don't know if `TOKEN_TYPE_ACCESS_TOKEN` applies to non-CAB use cases.

Otherwise, I think I'm ok with making it public given the module constraints we have.

oauth2_http/java/com/google/auth/oauth2/StsRequestHandler.java
```diff
@@ -175,6 +193,11 @@ private StsTokenExchangeResponse buildResponse(GenericData responseData) throws
       String scope = OAuth2Utils.validateString(responseData, "scope", PARSE_ERROR_PREFIX);
       builder.setScopes(Arrays.asList(scope.trim().split("\\s+")));
     }
+    if (responseData.containsKey("access_boundary_session_key")) {
+      builder.setAccessBoundarySessionKey(
+          OAuth2Utils.validateString(
+              responseData, "access_boundary_session_key", PARSE_ERROR_PREFIX));
+    }
```
lqiu96, line 200: nit: Perhaps `access_boundary_session_key` as a package-private constant and we can use this in the tests as well?

lsirac (approved):

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
lqiu96: Ok, I'm fine with this for now, but I'd rather have this be captured via logs (that is probably a separate project).

CC: @zhumin8

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -0,0 +1,193 @@
...
+                sourceAccessToken.getTokenValue(), OAuth2Utils.TOKEN_TYPE_ACCESS_TOKEN)
+            .setRequestTokenType(OAuth2Utils.TOKEN_TYPE_ACCESS_BOUNDARY_INTERMEDIARY_TOKEN)
+            .build();
+
+    StsRequestHandler handler =
+        StsRequestHandler.newBuilder(
+                tokenExchangeEndpoint, request, transportFactory.create().createRequestFactory())
+            .build();
+
+    StsTokenExchangeResponse response = handler.exchangeToken();
```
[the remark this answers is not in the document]
lqiu96: > we've already handled cases where sourceCredential.refreshIfExpired() fails or the access token is empty.

Those are two separate failures scenarios, right? Sorry, I should have been clearer on my above comment. The `handler.exchangeToken()` call also throws an IOException (https://github.com/googleapis/google-auth-library-java/blob/387b46315fa27424bd001625b3c53907d7a5ce1d/oauth2_http/java/com/google/auth/oauth2/StsRequestHandler.java#L96), which I just wanted to confirm if there needed any checks for.

oauth2_http/java/com/google/auth/oauth2/StsRequestHandler.java
lqiu96: Ah, the reason I was suggesting it is because I saw a test below had:
`private static final String ACCESS_BOUNDARY_SESSION_KEY = "accessBoundarySessionKey";` and assumed it was for the key (it looks like it's for the value instead).

nit: IMO, ACCESS_BOUNDARY_SESSION_KEY should reference `access_boundary_session_key`. If the constant probably be renamed to be something like ACCESS_BOUNDARY_SESSION_KEY_VALUE

oauth2_http/java/com/google/auth/oauth2/OAuth2Utils.java
nbayati: I'm going to leave this comment open and keep it as-is for now so I can merge this PR and unblock #1571. I'll address this comment in the next PR.

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
nbayati: A null check isn't needed here. The StsTokenExchangeResponse constructor enforces a non-null accessToken using checkNotNull. An exception would have been thrown earlier if it were null.
https://github.com/googleapis/google-auth-library-java/blob/387b46315fa27424bd001625b3c53907d7a5ce1d/oauth2_http/java/com/google/auth/oauth2/StsTokenExchangeResponse.java#L55-L62

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
nbayati: As we discussed, I'm marking this comment as resolved since this PR is not merging to the main branch right now.

cab-token-generator/pom.xml
nbayati: I'll leave this comment open as a reminder to update the version before the feature is merged to the main branch.

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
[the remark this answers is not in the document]
nbayati: Yes, you're right. However, since refreshCredentials() already declares throws IOException, we don't need an explicit check around exchangeToken(). The exception will propagate up automatically and the consuming code can handle it.

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
nbayati: Added documentation to the function. Please let me know if additional clarification is needed.

lqiu96: Thanks, that makes sense! Is there any issue if the intermediary token doesn't have an expiration time?

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -0,0 +1,209 @@
...
+      }
+
+      // Default to GDU when not supplied.
+      if (Strings.isNullOrEmpty(universeDomain)) {
+        this.universeDomain = Credentials.GOOGLE_DEFAULT_UNIVERSE;
+      }
+
+      // Ensure source credential's universe domain matches.
+      try {
+        if (!universeDomain.equals(sourceCredential.getUniverseDomain())) {
```
lqiu96, line 194: qq, does CAB plan to restrict the sourceCredentials to a subset of GoogleCredentials (i.e. only Compute or SA)?

Or can any GoogleCredential be used?

lqiu96 (approved):
LGTM. This PR's changes are going into the `client-side-cab` feature branch and not main. The unimplemented methods and additional changes will be addressed in future PRs.

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
nbayati: Currently, client-side CAB only works with service account tokens. In the future we might add support for other types of tokens.

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
nbayati: I believe the intermediary token should still work as expected without any issues even if there's no expiration time.

lqiu96: Sounds good, could we confirm that behavior. If it does fail, could you document that or link to docs that detail the behavior?

PR #1566, opened 2024-11-01
feat: ImpersonatedCredentials to support universe domain for idtoken and signblob

zhumin8:
follow up to https://github.com/googleapis/google-auth-library-java/pull/1528.

idtoken and sign flow are tested E2E according to TPC test guide for sa-to-sa impersonation.


oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java
```diff
@@ -345,12 +345,19 @@ public void setTransportFactory(HttpTransportFactory httpTransportFactory) {
...
+          getAccount(),
+          sourceCredentials,
+          getUniverseDomain(),
+          transportFactory.create(),
+          toSign,
+          ImmutableMap.of("delegates", this.delegates));
+    } catch (IOException e) {
+      // Throwing an IOException would be a breaking change, so wrap it here.
+      // This should not happen for this credential type.
+      throw new IllegalStateException(e);
```
lqiu96, line 359: Can we throw a SigningException here for consistency just like ComputeEngineCredentials?

If not, can the IllegalStateException have an error message that is the same as the one in ComputeEngineCredentials for consistency?

oauth2_http/javatests/com/google/auth/oauth2/ImpersonatedCredentialsTest.java
```diff
@@ -908,6 +908,38 @@ public void sign_serverError_throws() {
     }
   }
 
+  @Test
+  public void sign_sameAs_nonGDU() {
```
lqiu96, line 912: Is it possible for there be a test for ImpersonatedCredentials `sign_universeDomainException()`?

oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java
zhumin8: I've opted to throw IllegalStateException similar to https://github.com/googleapis/google-auth-library-java/blob/6668022a66692ee2330bee794a755120b0b0120a/oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java#L337
because in these cases, this exception will never be reached. Unlike in `ComputeEngineCredentials`, where `getUniverseDomain()` may actually throw this exception ([ref](https://github.com/googleapis/google-auth-library-java/blob/6668022a66692ee2330bee794a755120b0b0120a/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java#L218-L220)).
Added message in [a8d466f](https://github.com/googleapis/google-auth-library-java/pull/1566/commits/a8d466f3951686ddf0f6a7ae429cab505b337c54)

zhumin8: Here is my thought on this, I opted for IlligalStateException similar to ExternalAccountCredentials [here](https://github.com/googleapis/google-auth-library-java/blob/6668022a66692ee2330bee794a755120b0b0120a/oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java#L334-L337)
- Currently, unlike `ComputeEngineCredentials` where IOException can actually be thrown ([ref](https://github.com/googleapis/google-auth-library-java/blob/6668022a66692ee2330bee794a755120b0b0120a/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java#L218-L220)), `ImpersonatedCredentials` should not throw exception on [`getUniverseDomain()`](https://github.com/googleapis/google-auth-library-java/blob/6668022a66692ee2330bee794a755120b0b0120a/oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java#L494) calls. (this method throws because we do not want breaking changes when introducing the override) Because neither of the allowed source credential types (sa, u, external credentials) throw exception. 
- We are wrapping this with try-catch block to avoid breaking change. Throwing `SigningException` if it can happen for say, future allowed source credentials, it seems to imply a behavior change that should change the method signature. But for now, since we do not expect any of the allowed source credential to be throwing, it is acceptable to wrap and not add exception to method signature. Which IllegalStateException seems more appropriate.

Added message in [a8d466f](https://github.com/googleapis/google-auth-library-java/pull/1566/commits/a8d466f3951686ddf0f6a7ae429cab505b337c54)

oauth2_http/javatests/com/google/auth/oauth2/ImpersonatedCredentialsTest.java
zhumin8: added in [083557d](https://github.com/googleapis/google-auth-library-java/pull/1566/commits/083557d8f8a0a655b3a785e222816b58b193591c)

oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java
lqiu96: > ImpersonatedCredentials should not throw exception on [getUniverseDomain()](https://github.com/googleapis/google-auth-library-java/blob/6668022a66692ee2330bee794a755120b0b0120a/oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java#L494) calls

IIUC (I may be completely off on this as I don't know Impersonation at all), the sourceCredential in the ImpersonatedCredential could be of *any* type, right? Unless there is a limitation that the underlying sourceCredential for an ImpersonatedCredential can't be of ComputeEngineCredential. I am assuming that if a user tries to impersonate a ComputeEngineCredential, the call `getUniverseDomain()` may end up throwing an IOException. 

> Throwing SigningException if it can happen for say, future allowed source credentials, it seems to imply a behavior change that should change the method signature

I think SigningException is a RuntimeException which shouldn't require adding it as part of the method signature.

zhumin8: > an ImpersonatedCredential can't be of ComputeEngineCredential

This is my understanding. sourceCredential can be user or sa ([ref](https://github.com/googleapis/google-auth-library-java/blob/a8d466f3951686ddf0f6a7ae429cab505b337c54/oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java#L400-L410)), or a couple of external account types ([ref](https://github.com/googleapis/google-auth-library-java/blob/e3caf05831011dc05d3a8b01ebf79305eda70183/oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java#L281-L296)). So ComputeEngineCredential cannot be sourceCredential (and I don't know if impersonate a GCE cred have a use case?).

> I think SigningException shouldn't require adding it as part of the method signature.

Right, no signature change required, but I was concerned about behavior change.

zhumin8:
@TimurSadykov can you please also take a look?

oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java
lqiu96: Ah, I was looking at this [method](https://github.com/googleapis/google-auth-library-java/blob/a8d466f3951686ddf0f6a7ae429cab505b337c54/oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java#L140) which seems to allow any type of credential to be passed in. 

I only see this comment that tries to prevent other Credential types:
```
   * @param sourceCredentials the source credential used to acquire the impersonated credentials. It
   *     should be either a user account credential or a service account credential.
```

Seems a bit odd that certain static methods are checking for Credential types and others aren't. Maybe there is a reason for this... If not, probably something we can backfill and fix in a different PR.

I'm assuming in some downstream use case, some functionality will fail when using ImpersonatedCredentials with ComputeEngineCredentials as the source.  We probably don't have any users that have this setup (Impersonating a Compute Credentials), I just don't know enough about Impersonation to be sure about that. Would you know if this is the case?

> Right, no signature change required, but I was concerned about behavior change.

I am just thinking about keeping these consistent. As of now, I think they anything that signs should be either SigningException or IllegalStateException due to `getUniverseDomain()` call (even if ImpersonatedCredentials may not end up ever throwing it).

zhumin8: I lean to agree with you on keeping consistent.

On the " using ImpersonatedCredentials with ComputeEngineCredentials as the source", I also find it a bit odd that this [method](https://github.com/googleapis/google-auth-library-java/blob/a8d466f3951686ddf0f6a7ae429cab505b337c54/oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java#L140) you are quoting only specifies allowed source credential types in javadoc but not enforced. @TimurSadykov By any chance you know any context about it?  Was this a miss or intentional? Also, in general, is it fair to assume ComputeEngineCredentials should not be source credential for ImpersonatedCredentials?

zhumin8:
@sai-sunder-s If you can take a look. 
@aeitzman @lsirac These changes makes changes ImpersonatedCredentials, can you take a look to confirm if this change works fine with the BYOID flows?

oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java
zhumin8: 1. done in [a8d466f](https://github.com/googleapis/google-auth-library-java/pull/1566/commits/a8d466f3951686ddf0f6a7ae429cab505b337c54)
2. filed https://github.com/googleapis/google-auth-library-java/issues/1602

sai-sunder-s (approved):

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -599,11 +596,18 @@ public byte[] sign(byte[] toSign) {
...
+          this.getUniverseDomain(),
+          transportFactory.create(),
+          toSign,
+          Collections.<String, Object>emptyMap());
     } catch (SigningException ex) {
       throw ex;
     } catch (RuntimeException ex) {
       throw new SigningException("Signing failed", ex);
+    } catch (IOException ex) {
+      throw new SigningException("Failed to sign: Error obtaining universe domain", ex);
```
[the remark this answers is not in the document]
lqiu96: The `IAMUtils.sign()` method catches IOExceptions inside the method and re-throws it as `ServiceAccountSigner.SigningException`. I believe the only place that can throw IOException is from the `getUniverseDomain()` call, which should happen before we enter the `sign()` method.

oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java
```diff
@@ -345,12 +345,19 @@ public void setTransportFactory(HttpTransportFactory httpTransportFactory) {
...
+          getAccount(),
+          sourceCredentials,
+          getUniverseDomain(),
+          transportFactory.create(),
+          toSign,
+          ImmutableMap.of("delegates", this.delegates));
+    } catch (IOException ex) {
+      // Throwing an IOException would be a breaking change, so wrap it here.
+      // This should not happen for this credential type.
+      throw new IllegalStateException("Failed to sign: Error obtaining universe domain", ex);
```
lqiu96, line 359:
```suggestion
      throw new SigningException("Failed to sign: Error obtaining universe domain", ex);
```

lqiu96 (approved):

oauth2_http/java/com/google/auth/oauth2/IamUtils.java
```diff
@@ -62,10 +62,14 @@
  * features like signing.
  */
 class IamUtils {
-  private static final String SIGN_BLOB_URL_FORMAT =
-      "https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/%s:signBlob";
-  private static final String ID_TOKEN_URL_FORMAT =
-      "https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/%s:generateIdToken";
+
+  // iam credentials endpoints are to be formatted with universe domain and client email
```
aeitzman, line 66: Nit

```suggestion
  // IAM credentials endpoints are to be formatted with universe domain and client email.
```

oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java
```diff
@@ -775,8 +783,9 @@ public ImpersonatedCredentials build() {
         return new ImpersonatedCredentials(this);
       } catch (IOException e) {
         // throwing exception would be breaking change. catching instead.
-        // this should never happen.
-        throw new IllegalStateException(e);
+        // this should never happen because ImpersonatedCredential can only be SA or User
```
aeitzman, line 786:
```suggestion
        // This should never happen because ImpersonatedCredential can only be SA or User
```

oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java
```diff
@@ -775,8 +783,9 @@ public ImpersonatedCredentials build() {
         return new ImpersonatedCredentials(this);
       } catch (IOException e) {
         // throwing exception would be breaking change. catching instead.
```
aeitzman, line 785: Nit, I know this wasn't changed in the PR, but just cleaning up the comments.
```suggestion
        // Throwing an exception would be breaking change, so the exception is caught instead.
```

oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java
```diff
@@ -775,8 +783,9 @@ public ImpersonatedCredentials build() {
...
-        // this should never happen.
-        throw new IllegalStateException(e);
+        // this should never happen because ImpersonatedCredential can only be SA or User
+        // Credentials.
+        throw new SigningException("Signing failed", e);
```
aeitzman, line 788: Why is the exception type changed here?

aeitzman (approved):

oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java
zhumin8: Thanks for capturing this! This is actually a mistake on my side made in 2fec328d trying to address [this feedback](https://github.com/googleapis/google-auth-library-java/pull/1566#discussion_r1882521720)

The intended change is done in https://github.com/googleapis/google-auth-library-java/pull/1566/commits/3dbc8e908ac72219533d60c13cd4d599400e1bdf. I am reverting this accidental change in [40e2f9e](https://github.com/googleapis/google-auth-library-java/pull/1566/commits/40e2f9ebb6ccdcce659d39eefde0cb24071ed3f2)

PR #1571, opened 2024-11-04
feat: Add the generated ClientSideAccessBoundaryProto class for the Client-Side CAB feature

huangjiahua:
See: [go/client-side-cab-client-library-java](http://goto.google.com/client-side-cab-client-library-java)

This PR introduces the `ClientSideAccessBoundaryProto` Protobuf Java class, necessary for serializing the Client-Side Access Boundary (CAB) token. Following the Auth SDK team's recommendation, the Protobuf definition itself isn't included in the repository, as it's anticipated to remain stable.

To support this generated class and the Client-Side CAB functionality, two new dependencies, `protobuf` and `cel`, have been added to the `cab-token-generator` package.

To address Maven build issues, I've excluded `findbugs` and `guava` from the cel dependency. 

---


huangjiahua:
Will move the generated code to the new package #1562 creates

huangjiahua:
Moved to the new package. Ready for review

cab-token-generator/java/com/google/auth/credentialaccessboundary/protobuf/ClientSideAccessBoundaryProto.java
```diff
@@ -0,0 +1,1817 @@
+// Generated by the protocol buffer compiler.  DO NOT EDIT!
+// source: client_side_access_boundary.proto
```
lqiu96, line 2: How we plan on on keeping this up to date? Where does this `client_side_access_boundary.proto` file live and how does it get updated?

huangjiahua: I got the suggestion from @aeitzman to manually generate this class first and not include the source proto file for now. We do have a task to figure out a way to automate the protobuf workflow later in the project.

pom.xml
```diff
@@ -77,6 +77,8 @@
     <project.findbugs.version>3.0.2</project.findbugs.version>
     <deploy.autorelease>false</deploy.autorelease>
     <project.error-prone.version>2.35.1</project.error-prone.version>
+    <project.protobuf.version>4.28.3</project.protobuf.version>
```
lqiu96, line 80: The Java SDK protos are generated with Protoc 25.5. Libraries-Bom will set the Protobuf-Java runtime version to [4.28.3](https://github.com/googleapis/java-cloud-bom/blob/18b9062657f953eb44d2e9cf4945c6bfbf1574aa/libraries-bom/pom.xml#L62-L71). I think this should be 3.25.5 since this is part of the BOM.

lqiu96:
> To address Maven build issues, I've excluded findbugs and guava from the cel dependency.

Would you be able to share the build issues? I'm curious what the issues are cause off the top of my mind, I don't think we need to exclude them

huangjiahua:
> > To address Maven build issues, I've excluded findbugs and guava from the cel dependency.
> 
> Would you be able to share the build issues? I'm curious what the issues are cause off the top of my mind, I don't think we need to exclude them

Sure, let me reproduce the errors.

If I don exclude `findbugs`, we get:

```
[ERROR] Duplicate classes found:
[ERROR] 
[ERROR]   Found in:
[ERROR]     com.google.code.findbugs:jsr305:jar:3.0.2:compile
[ERROR]     com.google.code.findbugs:annotations:jar:3.0.1:compile
[ERROR]   Duplicate classes:
[ERROR]     javax/annotation/RegEx$Checker.class
[ERROR]     javax/annotation/MatchesPattern$Checker.class
[ERROR]     javax/annotation/meta/When.class
[ERROR]     javax/annotation/Nonnull$Checker.class
[ERROR]     javax/annotation/Nonnegative$Checker.class
```

If we don't exclude `guava`, we get:

```
[ERROR] Rule 2: org.apache.maven.enforcer.rules.dependency.RequireUpperBoundDeps failed with message:
[ERROR] Failed while enforcing RequireUpperBoundDeps. The error(s) are [
[ERROR] Require upper bound dependencies error for com.google.guava:guava:33.3.1-android paths to dependency are:
[ERROR] +-com.google.auth:cab-token-generator:1.29.1-SNAPSHOT
[ERROR]   +-com.google.guava:guava:33.3.1-android
[ERROR] and
[ERROR] +-com.google.auth:cab-token-generator:1.29.1-SNAPSHOT
[ERROR]   +-com.google.auth:google-auth-library-oauth2-http:1.29.1-SNAPSHOT
[ERROR]     +-com.google.guava:guava:33.3.1-android (managed) <-- com.google.guava:guava:33.3.1-android
[ERROR] and
[ERROR] +-com.google.auth:cab-token-generator:1.29.1-SNAPSHOT
[ERROR]   +-com.google.http-client:google-http-client:1.45.0
[ERROR]     +-com.google.guava:guava:33.3.1-android (managed) <-- com.google.guava:guava:30.1.1-android
[ERROR] and
[ERROR] +-com.google.auth:cab-token-generator:1.29.1-SNAPSHOT
[ERROR]   +-dev.cel:cel:0.8.0
[ERROR]     +-com.google.guava:guava:33.3.1-android (managed) <-- com.google.guava:guava:33.3.1-jre
[ERROR] and
[ERROR] +-com.google.auth:cab-token-generator:1.29.1-SNAPSHOT
[ERROR]   +-com.google.http-client:google-http-client:1.45.0
[ERROR]     +-io.opencensus:opencensus-contrib-http-util:0.31.1
[ERROR]       +-com.google.guava:guava:33.3.1-android (managed) <-- com.google.guava:guava:29.0-android
[ERROR] and
[ERROR] +-com.google.auth:cab-token-generator:1.29.1-SNAPSHOT
[ERROR]   +-dev.cel:cel:0.8.0
[ERROR]     +-com.google.protobuf:protobuf-java-util:4.28.2
[ERROR]       +-com.google.guava:guava:33.3.1-android (managed) <-- com.google.guava:guava:32.0.1-jre
[ERROR] ]
```

pom.xml
huangjiahua: Actually downgrading it to 3.25.5 causes `RequireUpperBoundDeps` issue since CEL depends on 4.28.2:

```
[ERROR] Rule 2: org.apache.maven.enforcer.rules.dependency.RequireUpperBoundDeps failed with message:
[ERROR] Failed while enforcing RequireUpperBoundDeps. The error(s) are [
[ERROR] Require upper bound dependencies error for com.google.protobuf:protobuf-java:3.25.5 paths to dependency are:
[ERROR] +-com.google.auth:cab-token-generator:1.29.1-SNAPSHOT
[ERROR]   +-com.google.protobuf:protobuf-java:3.25.5
[ERROR] and
[ERROR] +-com.google.auth:cab-token-generator:1.29.1-SNAPSHOT
[ERROR]   +-dev.cel:cel:0.8.0
[ERROR]     +-com.google.protobuf:protobuf-java:3.25.5 (managed) <-- com.google.protobuf:protobuf-java:4.28.2
[ERROR] and
[ERROR] +-com.google.auth:cab-token-generator:1.29.1-SNAPSHOT
[ERROR]   +-dev.cel:cel:0.8.0
[ERROR]     +-com.google.protobuf:protobuf-java-util:4.28.2
[ERROR]       +-com.google.protobuf:protobuf-java:3.25.5 (managed) <-- com.google.protobuf:protobuf-java:4.28.2
[ERROR] ]
```

cab-token-generator/java/com/google/auth/credentialaccessboundary/protobuf/ClientSideAccessBoundaryProto.java
aeitzman: The proto lives in google3 right now, and we do not expect it to get updated very often. For context, the credential access boundary definition used by the existing online cab feature is also based off a Google3 protobuf file, and has not changed since launch in 2021.

lqiu96: > We do have a task to figure out a way to automate the protobuf workflow later in the project.

Sure, I think that's fine for now to unblock development.

Regarding this generated file, was this generated with an older Protoc version? I think the later versions have a comment that describes the protoc version (i.e. `// Protobuf Java Version: 3.25.5`). If so, can we generate this with Protoc 25.5?

lqiu96:
Took a look at the logs above. I'm not entirely sure why CEL is bringing in findbugs-annotations instead of findbug-jsr305. Either way, I think that should be fine to ignore since we don't depend on it in the auth library at all. 

For Guava, I think the issue is due to the guava flavor. It looks to be treating guava-android and guava-jre as two different versions and can't tell that they're actually the same version. I think we are also fine to exclude this.

I'll do a bit more digging to confirm this. If the above is confirmed, a few small comments to explain the exclusion would be great

cab-token-generator/java/com/google/auth/credentialaccessboundary/protobuf/ClientSideAccessBoundaryProto.java
huangjiahua: Done

huangjiahua:
> I'll do a bit more digging to confirm this. If the above is confirmed, a few small comments to explain the exclusion would be great

I added a comment here. I also added the protobuf to the exclusion list because CEL is using 4.28.2 which will trigger `RequireUpperBoundDeps` error.

pom.xml
```diff
@@ -77,6 +77,8 @@
     <project.findbugs.version>3.0.2</project.findbugs.version>
     <deploy.autorelease>false</deploy.autorelease>
     <project.error-prone.version>2.35.1</project.error-prone.version>
+    <project.protobuf.version>3.25.5</project.protobuf.version>
```
lqiu96, line 80: Thanks, I will need to double check this because I know CEL brings in Protobuf-Java 4.x. 

I just need to confirm compatibility. I'll coordinate with CEL and confirm.

cab-token-generator/java/com/google/auth/credentialaccessboundary/protobuf/ClientSideAccessBoundaryProto.java
blakeli0: > The proto lives in google3 right now, and we do not expect it to get updated very often

Agree that this is fine for now to unblock the development. But I want to emphasize that introducing a new dependency `protobuf-java` in auth and a protoc gen code without a `.proto` file will cause frictions to the overall protobuf maintenance strategy in Java Client Libraries. Even if the content do not update at all, we may need to regenerate it with a newer version of protoc due to compatibility issues.
For example, if we want to upgrade `protobuf-java` to a version that is not compatible with 3.25.5 gen code anymore, we would need to manually regenerate this file from somewhere.
That being said, we have a lot of experience in automating proto generation process, please let us know when you start the task, it would be great if it can be integrated into our existing client library generation process.

blakeli0:
> necessary for serializing the Client-Side Access Boundary (CAB) token

Are we making a grpc call or rest call? If it is not a grpc call, I don't think we have to introduce the proto gen code into auth. We can create our own POJO object to serialize the JSON response. We may have to update the POJO when the proto changes, but since it does not change often, I think the risk is very low.

huangjiahua:
> Are we making a grpc call or rest call?

The proto here is not for grpc call. This proto is for serializing the CAB restriction for the CAB token generation, so protobuf is still required.

blakeli0:
> > Are we making a grpc call or rest call?
> 
> The proto here is not for grpc call. This proto is for serializing the CAB restriction for the CAB token generation, so protobuf is still required.

If that's the case, I don't think protobuf is required. We can use Gson for serialization as well.

huangjiahua:
> If that's the case, I don't think protobuf is required. We can use Gson for serialization as well.

@blakeli0 Does Gson support serializing to protobuf wire format? I couldn't find any documentation on this use case.

blakeli0:
> > If that's the case, I don't think protobuf is required. We can use Gson for serialization as well.
> 
> @blakeli0 Does Gson support serializing to protobuf wire format? I couldn't find any documentation on this use case.

I might be missing some context. If it is not used for making gRPC calls, why do we have to serialize it to protobuf wire format?

huangjiahua:
> > > If that's the case, I don't think protobuf is required. We can use Gson for serialization as well.
> > 
> > 
> > @blakeli0 Does Gson support serializing to protobuf wire format? I couldn't find any documentation on this use case.
> 
> I might be missing some context. If it is not used for making gRPC calls, why do we have to serialize it to protobuf wire format?

Yeah, I should've included the context: The client-side CAB token is designed to include an encrypted protobuf wired format. One of the reasons the token need to be a protobuf wired format is because it needs to carry a compiled CEL expression in AST format. The server will need to verify the token using the same protobuf. More on [go/client-side-cab-design-doc](http://goto.google.com/client-side-cab-design-doc).

blakeli0:
> > > > If that's the case, I don't think protobuf is required. We can use Gson for serialization as well.
> > > 
> > > 
> > > @blakeli0 Does Gson support serializing to protobuf wire format? I couldn't find any documentation on this use case.
> > 
> > 
> > I might be missing some context. If it is not used for making gRPC calls, why do we have to serialize it to protobuf wire format?
> 
> Yeah, I should've included the context: The client-side CAB token is designed to include an encrypted protobuf wired format. One of the reasons the token need to be a protobuf wired format is because it needs to carry a compiled CEL expression in AST format. The server will need to verify the token using the same protobuf. More on [go/client-side-cab-design-doc](http://goto.google.com/client-side-cab-design-doc).

I see, that makes sense. It might be too late to change the design, why does it have to be protobuf wired format? Can JSON format suffice? I think all other types of tokens are sent as Strings directly through rest calls.

blakeli0 (approved):
Discussed offline that this PR can be merged for now to unblock development. Before this feature goes GA/merges to main, we need to make sure we have a proper strategy of maintaining the protobuf gencode and protobuf runtime dependency in this repo.

PR #1573, opened 2024-11-07
fix: Make it explicit that there is a network call to MDS to get SecureSessionAgentConfig

rmehta19:
Make the classes non-final while we're here to allow for mocking in tests if needed.

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/S2A.java
+++ b/oauth2_http/java/com/google/auth/oauth2/S2A.java
@@ -58,7 +58,7 @@ import javax.annotation.concurrent.ThreadSafe;
  * <p>This is an experimental utility.
  */
 @ThreadSafe
-public final class S2A {
+public class S2A {
   static final String S2A_PLAINTEXT_ADDRESS_JSON_KEY = "plaintext_address";
   static final String S2A_MTLS_ADDRESS_JSON_KEY = "mtls_address";
   static final String S2A_CONFIG_ENDPOINT_POSTFIX =
@@ -81,14 +81,9 @@ public final class S2A {
     this.config = getS2AConfigFromMDS();
   }
 
-  /** @return the mTLS S2A Address from the mTLS config. */
-  public String getMtlsS2AAddress() {
-    return config.getMtlsAddress();
-  }
-
-  /** @return the plaintext S2A Address from the mTLS config. */
-  public String getPlaintextS2AAddress() {
-    return config.getPlaintextAddress();
+  /** @return the cached S2AConfig. */
+  public S2AConfig getConfigFromMDS() {
+    return config;
   }
 
   public static Builder newBuilder() {
--- a/oauth2_http/java/com/google/auth/oauth2/SecureSessionAgent.java
+++ b/oauth2_http/java/com/google/auth/oauth2/SecureSessionAgent.java
@@ -82,7 +82,7 @@ public class SecureSessionAgent {
    * This method makes a network call to MDS to get the {@link SecureSessionAgentConfig} which
    * contains the plaintext and mtls address to reach the S2A (Secure Session Agent).
    *
-   * @return the cached SecureSessionAgentConfig.
+   * @return a SecureSessionAgentConfig.
    */
   public SecureSessionAgentConfig getConfig() {
     return getSecureSessionAgentConfigFromMDS();
--- a/oauth2_http/java/com/google/auth/oauth2/SecureSessionAgent.java
+++ b/oauth2_http/java/com/google/auth/oauth2/SecureSessionAgent.java
@@ -72,18 +72,20 @@ public class SecureSessionAgent {
   private static final String MDS_MTLS_ENDPOINT =
       ComputeEngineCredentials.getMetadataServerUrl() + S2A_CONFIG_ENDPOINT_POSTFIX;
 
-  private SecureSessionAgentConfig config;
-
   private transient HttpTransportFactory transportFactory;
 
   SecureSessionAgent(SecureSessionAgent.Builder builder) {
     this.transportFactory = builder.getHttpTransportFactory();
-    this.config = getSecureSessionAgentConfigFromMDS();
   }
 
-  /** @return the cached SecureSessionAgentConfig. */
-  public SecureSessionAgentConfig getConfigFromMDS() {
-    return config;
+  /**
+   * This method makes a network call to MDS to get the {@link SecureSessionAgentConfig} which
+   * contains the plaintext and mtls address to reach the S2A (Secure Session Agent).
+   *
+   * @return the cached SecureSessionAgentConfig.
+   */
+  public SecureSessionAgentConfig getConfig() {
+    return getSecureSessionAgentConfigFromMDS();
   }
 
   public static Builder newBuilder() {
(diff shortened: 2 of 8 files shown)


oauth2_http/java/com/google/auth/oauth2/S2A.java
```diff
@@ -58,7 +58,7 @@
  * <p>This is an experimental utility.
  */
 @ThreadSafe
-public final class S2A {
+public class S2A {
```
lqiu96, line 61: On a second look, do you have concerns if we can make this name written out specifically (SecureSessionAgent.java) and (SecureSessionAgentConfig.java) as S2A may not be well known and specific to Google?

oauth2_http/java/com/google/auth/oauth2/S2A.java
```diff
@@ -81,14 +81,9 @@ public final class S2A {
...
-    return config.getMtlsAddress();
-  }
-
-  /** @return the plaintext S2A Address from the mTLS config. */
-  public String getPlaintextS2AAddress() {
-    return config.getPlaintextAddress();
+  /** @return the cached S2AConfig. */
+  public S2AConfig getConfigFromMDS() {
+    return config;
   }
```
lqiu96, line 87: What I had in mind was
```
public SecureSessionAgentConfig getConfig() {
  return getS2AConfigFromMDS();
}
```
as this `getConfig()` call is the explicit call that will make the network call.

oauth2_http/java/com/google/auth/oauth2/S2A.java
```diff
@@ -81,14 +81,9 @@ public final class S2A {
...
 
   public static Builder newBuilder() {
```
lqiu96, line 89: Also, if you want, you can include a static `create()` helper that creates a default instance (i.e. `SecureSessionAgent.create()`).

oauth2_http/java/com/google/auth/oauth2/S2A.java
rmehta19: That's fine with me, done in https://github.com/googleapis/google-auth-library-java/pull/1573/commits/e6b6e18582b7538a2ea6082c13885355016fbcd4

oauth2_http/java/com/google/auth/oauth2/S2A.java
rmehta19: I see, so right now the config is being set in the builder and is a member of the class. The current setup ensures that only one call to MDS is made (since there is no need to make the request to MDS multiple times) no matter how many times a user calls getConfigFromMds.

I'm fine with this though, it shifts responsibility to the user to store config and not call getConfig() more than once to save latency of their application. This seems aligned with what you mentioned in our offline discussion yesterday about how it should be explicit a network call is being made.

Done in: https://github.com/googleapis/google-auth-library-java/pull/1573/commits/381efe4531e481a2f4a36fbaa70dc94ce55ecc38

oauth2_http/java/com/google/auth/oauth2/S2A.java
rmehta19: Done in https://github.com/googleapis/google-auth-library-java/pull/1573/commits/7ce029110e5752ae469e46174c97e30d65caa29c

oauth2_http/java/com/google/auth/oauth2/SecureSessionAgent.java
```diff
@@ -58,7 +58,7 @@
  * <p>This is an experimental utility.
  */
 @ThreadSafe
-public final class S2A {
+public class SecureSessionAgent {
```
lqiu96, line 61: Can we keep final or does that break something?

oauth2_http/java/com/google/auth/oauth2/SecureSessionAgent.java
```diff
@@ -72,23 +72,24 @@ public final class S2A {
...
+   */
+  public SecureSessionAgentConfig getConfig() {
+    return getSecureSessionAgentConfigFromMDS();
   }
 
-  /** @return the plaintext S2A Address from the mTLS config. */
-  public String getPlaintextS2AAddress() {
-    return config.getPlaintextAddress();
+  public static SecureSessionAgent create() {
+    return newBuilder().build();
```
lqiu96, line 92:
```suggestion
  /** @return default instance of SecureSessionAgent */
  public static SecureSessionAgent create() {
    return newBuilder().build();
```

or something similar

oauth2_http/java/com/google/auth/oauth2/SecureSessionAgentConfig.java
```diff
@@ -33,7 +33,7 @@
 import com.google.errorprone.annotations.CanIgnoreReturnValue;
 
 /** Holds an mTLS configuration (consists of address of S2A) retrieved from the Metadata Server. */
-final class S2AConfig {
+public class SecureSessionAgentConfig {
```
lqiu96, line 36: Can this continue to be final?

lqiu96 (approved):
LGTM. Added a few nits/ questions

lqiu96:
@rmehta19  Also can you update the title with `fix: {title}`?

oauth2_http/java/com/google/auth/oauth2/SecureSessionAgentConfig.java
rmehta19: final will let us mock in tests (if that's something we want to test in sdk PR). If not, I'm fine keeping this final

rmehta19:
@lqiu96 , Thanks for the review! Addressed comments.

oauth2_http/java/com/google/auth/oauth2/SecureSessionAgent.java
lqiu96: Can you point me to where we need to mock it? I think we should be able to find other ways around, but just want to check.

rmehta19: In https://github.com/googleapis/sdk-platform-java/pull/3326 if we want to test: `createS2ASecuredChannelCredentials` function added. I was thinking Mockito would be the simplest way to mock this lib, in which case I believe the class would have to be non-final

lqiu96: Do you have an example of how we're testing in the PR right now?

One option that comes to mind for testing this without mocks:
Update the method to: `createS2ASecuredChannelCredentials(SecureSessionAgent s2a)`

The tests for createS2ASecuredChannelCredentials would pass version of SecureSessionAgent (i.e. SecureSessionAgentTest which overrides the `getConfig()` and returns SecureSessionAgentConfig with the needed values). 

I *think* it's possible to test it without having to mock it.

lqiu96:
> @lqiu96 , Thanks for the review! Addressed comments.

LGTM. Let me know if you feel strongly about final/ non-final. I'll approve again and we can merge then

lqiu96:
FYI the linter is failing. I think you need to run `mvn fmt:format` again

oauth2_http/java/com/google/auth/oauth2/SecureSessionAgent.java
lqiu96: Ah ignore the above, I was mistaken. I am fine with removing the final keyword.

rmehta19:
> > @lqiu96 , Thanks for the review! Addressed comments.
> 
> LGTM. Let me know if you feel strongly about final/ non-final. I'll approve again and we can merge then

As discussed offline, let's keep it final so that we can test it from Gax where it is used

PR #1583, opened 2024-11-23
feat: Implement refreshCredentialsIfRequired for intermediate token r…

nbayati:
This PR implements `refreshCredentialsIfRequired` which will be called by generateToken(). The refresh logic was implemented based on the [design doc]([go/client-side-cab-client-library-java](http://goto.google.com/client-side-cab-client-library-java)), which takes into account the `refreshMargin` and `minimumTokenLifetime`, and decides on either async or synchronous refresh if one is needed.

Notes:
- This PR merges the changes to a feature branch and not the main branch.
- Unit tests to follow.


cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -45,41 +46,65 @@
...
   private final String tokenExchangeEndpoint;
   private String accessBoundarySessionKey;
   private AccessToken intermediateAccessToken;
+  private final Duration minimumTokenLifetime;
+  private final Duration refreshMargin;
+  private static final Duration DEFAULT_REFRESH_MARGIN = Duration.ofMinutes(30);
+  private static final Duration DEFAULT_MINIMUM_TOKEN_LIFETIME = Duration.ofMinutes(3);
+  private final Object refreshLock = new Object[0]; // Lock for refresh operations
+  @Nullable private SettableFuture<Void> currentRefreshFuture;
+  private final ExecutorService backgroundExecutor = Executors.newSingleThreadExecutor();
```
lqiu96, line 71: This isn't being lifecycle managed. If we create this, then we'll need to close this as well.

To keep with the existing Auth flow, let's use `MoreExecutors.directExecutor()`

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -101,25 +126,108 @@ private void refreshCredentials() throws IOException {
             .build();
 
     StsTokenExchangeResponse response = handler.exchangeToken();
-    this.accessBoundarySessionKey = response.getAccessBoundarySessionKey();
-    this.intermediateAccessToken = response.getAccessToken();
 
-    // The STS endpoint will only return the expiration time for the intermediary token
+    synchronized (refreshLock) {
```
lqiu96, line 130: qq, why do we need a lock here?

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -101,25 +126,108 @@ private void refreshCredentials() throws IOException {
...
+            throw (IOException) cause;
+          } else {
+            throw new IOException("Asynchronous refresh failed.", cause);
+          }
+        }
+      } else {
+        // No asynchronous refresh is running, perform a synchronous refresh.
+        refreshCredentials();
+      }
+    }
```
lqiu96, line 196: I'm a bit concerned with this. I don't think we should be holding a lock over a network call (refreshCredentials) or while waiting for the future (currentRefreshFuture.get()).

Is there a reason why this needs to obtain the lock?

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -45,41 +46,65 @@
...
   private final transient HttpTransportFactory transportFactory;
   private final String tokenExchangeEndpoint;
   private String accessBoundarySessionKey;
   private AccessToken intermediateAccessToken;
+  private final Duration minimumTokenLifetime;
+  private final Duration refreshMargin;
+  private static final Duration DEFAULT_REFRESH_MARGIN = Duration.ofMinutes(30);
+  private static final Duration DEFAULT_MINIMUM_TOKEN_LIFETIME = Duration.ofMinutes(3);
+  private final Object refreshLock = new Object[0]; // Lock for refresh operations
+  @Nullable private SettableFuture<Void> currentRefreshFuture;
```
lqiu96, line 70: I think this needs to volatile

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -101,25 +126,108 @@ private void refreshCredentials() throws IOException {
...
+        && sourceAccessToken.getExpirationTime() != null) {
+      return new AccessToken(
+          intermediateToken.getTokenValue(), sourceAccessToken.getExpirationTime());
+    }
+    return intermediateToken; // Return original if no modification needed
+  }
+
+  private void startAsynchronousRefresh() {
+    // Obtain the lock before checking or modifying currentRefreshFuture to prevent race conditions.
+    synchronized (refreshLock) {
```
lqiu96, line 155: qq, why do we need to obtain the lock here. I think for this it's probably fine for async, but I'm not putting together why it's needed.

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -45,41 +46,65 @@
...
 public final class ClientSideCredentialAccessBoundaryFactory {
   private final GoogleCredentials sourceCredential;
   private final transient HttpTransportFactory transportFactory;
   private final String tokenExchangeEndpoint;
   private String accessBoundarySessionKey;
   private AccessToken intermediateAccessToken;
+  private final Duration minimumTokenLifetime;
+  private final Duration refreshMargin;
+  private static final Duration DEFAULT_REFRESH_MARGIN = Duration.ofMinutes(30);
+  private static final Duration DEFAULT_MINIMUM_TOKEN_LIFETIME = Duration.ofMinutes(3);
```
lqiu96, line 68: Is it possible for DEFAULT_REFRESH_MARGIN to be less than DEFAULT_MINIMUM_TOKEN_LIFETIME? Do we need any validation for user configurations?

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
nbayati: From [comment thread](https://docs.google.com/document/d/1YLYAlae_AAOtb9Mw-TRawi5df_8EfHv-hO04HgZgiYA/edit?disco=AAABYwpOifM) in the design doc:

> We still need to synchronize access to the intermediary token and session key. Concurrent access by the background refresh task and main user thread could lead to inconsistencies. Because the background refresh task updates these values separately, there's a chance the main thread could grab a new token with an old key, causing problems.

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
nbayati: The lock here prevents multiple concurrent asynchronous refreshes from starting. Without the lock, the following scenario could occur:
1. Thread A checks currentRefreshFuture and finds it null or done.
2. Before Thread A can assign a new future to currentRefreshFuture, Thread B also checks and finds it null or done.
3. Both Thread A and Thread B start a refresh, wasting resources and potentially leading to inconsistencies if they complete out of order and overwrite each other's results.

The lock ensures that only one asynchronous refresh can be in progress at any given time.

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
nbayati: I'm not used to async programming so there's a good chance I've over-used the lock! My thought process was that we need to obtain the lock before checking/setting the value of `currentRefreshFuture` because it can be updated by another thread. We could only hold the lock while checking the value though, something like:
```
    Boolean isAsyncInProgress = false;
    synchronized (refreshLock) {
      isAsyncInProgress = currentRefreshFuture != null && !currentRefreshFuture.isDone();
    }
    if (isAsyncInProgress) {
      try {
        ... 
      }
    } else {
      // No asynchronous refresh is running, perform a synchronous refresh.
      refreshCredentials();
    }
  }
```

Would this address your concern?

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
lqiu96: Ok, I think I understand. I think there are a few things regarding this.

1. When either an async or sync call is made, we really only want one execution/ call to go through at a time. Even if `refreshCredentialsIfRequired()` is called repeatedly, then one call really should be made. For that, I believe we would want functionality similar to [this](https://github.com/googleapis/google-auth-library-java/blob/9a7c2e08a6fd891d802e3fb7729d163841bae524/oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java#L259). It will find the existing execution or create a new one as a transaction. If multiple calls are made at the same time, each of the calls will find the same refresh action while it's still active.
2. For the sync flow, we should kick off the refresh action [here](https://github.com/googleapis/google-auth-library-java/blob/eb4b793cca84bd64c6de00db0303e9e10c566aae/cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java#L222) via the createRefreshTask linked above and block until the future is resolved via `blockingRefresh()`.
3. For the async flow, it should just create a refresh action via createRefreshTask

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
nbayati: The default value of refresh margin is 30 minutes and the default or minimum token lifetime is 3 minutes.
Technically the users could set those values in a way that the refresh margin ends up being less than the minimum token lifetime.

lqiu96: Is this something that we want to guard against? Or is this fine to allow users to do this?

cab-token-generator/javatests/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactoryTest.java
```diff
@@ -0,0 +1,311 @@
...
+        fail("Should fail with universe domain mismatch.");
+      } catch (IllegalArgumentException e) {
+        assertEquals(
+            "The client side access boundary credential's universe domain must be the same as the source credential.",
+            e.getMessage());
+      }
+    }
+  }
+
+  private static GoogleCredentials getServiceAccountSourceCredentials(boolean canRefresh)
```
lqiu96, line 288: nit: Can this take in a `MockTokenServerTransportFactory` param with the error/ serviceAccount configured instead of a boolean? I think that would be cleaner as this method would just be creating the Credentials instead of creating credentials and configuring the response

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
nbayati: No we're not going to enforce it. The only thing we need to check is that the entered values are positive. The users might want to intentionally configure those values in a way that they only have blocking refresh, meaning refresh margin would be smaller than minimum token lifetime, and that's ok.

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -151,6 +394,48 @@ public Builder setSourceCredential(GoogleCredentials sourceCredential) {
...
+     * than this value, CAB token generation will be blocked and a refresh will be initiated. This
+     * ensures that generated CAB tokens have a sufficient lifetime for use.
+     *
+     * @param minimumTokenLifetime The minimum acceptable lifetime for a generated CAB token. Must
+     *     be positive.
+     * @return This {@code Builder} object.
+     * @throws IllegalArgumentException if minimumTokenLifetime is negative or zero.
+     */
+    @CanIgnoreReturnValue
+    public Builder setMinimumTokenLifetime(Duration minimumTokenLifetime) {
```
lqiu96, line 411: Might need to consider null checks for Duration as well

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -151,6 +394,48 @@ public Builder setSourceCredential(GoogleCredentials sourceCredential) {
...
+     * <p>This duration specifies how far in advance of the intermediate access token's expiration
+     * time an asynchronous refresh should be initiated. If not provided, it will default to 30
+     * minutes.
+     *
+     * @param refreshMargin The refresh margin. Must be positive.
+     * @return This {@code Builder} object.
+     * @throws IllegalArgumentException if refreshMargin is negative or zero.
+     */
+    @CanIgnoreReturnValue
+    public Builder setRefreshMargin(Duration refreshMargin) {
```
lqiu96, line 431: Might need to consider null checks for Duration as well

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -151,6 +394,48 @@ public Builder setSourceCredential(GoogleCredentials sourceCredential) {
...
+     *
+     * @param minimumTokenLifetime The minimum acceptable lifetime for a generated CAB token. Must
+     *     be positive.
+     * @return This {@code Builder} object.
+     * @throws IllegalArgumentException if minimumTokenLifetime is negative or zero.
+     */
+    @CanIgnoreReturnValue
+    public Builder setMinimumTokenLifetime(Duration minimumTokenLifetime) {
+      if (minimumTokenLifetime.isNegative() || minimumTokenLifetime.isZero()) {
+        throw new IllegalArgumentException("Minimum token lifetime must be positive.");
```
lqiu96, line 413: nit: I think `must be greater than zero` sounds better (Ignore if this is official verbiage)

lqiu96 (approved):
LGTM, thanks! Added a few smaller comments but the overall flow is similar to Oauth2Credentials

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -206,4 +491,27 @@ public ClientSideCredentialAccessBoundaryFactory build() {
       return new ClientSideCredentialAccessBoundaryFactory(this);
     }
   }
+
+  @VisibleForTesting
+  String getAccessBoundarySessionKey() {
+
```
aeitzman, line 497: nit: remove newline

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -206,4 +491,27 @@ public ClientSideCredentialAccessBoundaryFactory build() {
...
+    return intermediateCredentials != null
+        ? intermediateCredentials.accessBoundarySessionKey
+        : null;
+  }
+
+  @VisibleForTesting
+  AccessToken getIntermediateAccessToken() {
```
aeitzman, line 504: nit: Just to stay consistent with the style of our other classes, can the order be:

Main Class stuff
Utility Static Classes
Builder Stuff

So move these getters above the utility/builder classes and move the NewBuilder method down with the rest of the builder logic

aeitzman (approved):

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -44,42 +45,195 @@
...
+import com.google.common.util.concurrent.ListenableFutureTask;
+import com.google.common.util.concurrent.MoreExecutors;
 import com.google.errorprone.annotations.CanIgnoreReturnValue;
 import java.io.IOException;
+import java.time.Duration;
+import java.util.Date;
+import java.util.concurrent.ExecutionException;
+import javax.annotation.Nullable;
 
 public final class ClientSideCredentialAccessBoundaryFactory {
```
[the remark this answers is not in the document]
nbayati: Removed the final keyword.

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -44,42 +45,195 @@
...
+  }
+
+  private RefreshType determineRefreshType() {
+    if (intermediateCredentials == null
+        || intermediateCredentials.intermediateAccessToken == null) {
+      // A blocking refresh is needed if the intermediate access token doesn't exist.
+      return RefreshType.BLOCKING;
+    }
+
+    AccessToken localAccessToken = intermediateCredentials.intermediateAccessToken;
```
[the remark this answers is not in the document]
nbayati: Renamed.

nbayati:
I Renamed `refreshCredentials()` to `fetchIntermediateCredentials()` to better reflect its updated functionality. This method now returns the fetched intermediate credentials instead of updating the stored values directly. The update is handled by `finishRefreshTask`.

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -101,41 +255,130 @@ private void refreshCredentials() throws IOException {
...
+   * successful. The single-flight "slot" is cleared, allowing subsequent refresh attempts. Any
+   * exceptions during the refresh are caught and suppressed to prevent indefinite blocking of
+   * subsequent refresh attempts.
+   */
+  private void finishRefreshTask(ListenableFuture<IntermediateCredentials> finishedTask) {
+    synchronized (refreshLock) {
+      try {
+        this.intermediateCredentials = Futures.getDone(finishedTask);
+      } catch (Exception e) {
+        // noop
```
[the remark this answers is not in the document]
nbayati: Good catch! OAuth2Credentials had implemented this function like this, I think maybe to avoid throwing exceptions in the async mode? I'm not sure! I did some testing and the way it's implemented right now causes the blocking refresh to fail silently without notifying anywhere that it has failed. I'm adding some error handling and will send a new commit soon.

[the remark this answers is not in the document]
nbayati: Fixed! Now the blocking refresh will correctly propagate the exception if something goes wrong during the refresh.

cab-token-generator/javatests/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactoryTest.java
```diff
@@ -0,0 +1,575 @@
...
+            .build();
+
+    ClientSideCredentialAccessBoundaryFactory factory =
+        ClientSideCredentialAccessBoundaryFactory.newBuilder()
+            .setUniverseDomain(universeDomain)
+            .setSourceCredential(sourceCredentials)
+            .setHttpTransportFactory(mockStsTransportFactory)
+            .build();
+
+    factory.fetchIntermediateCredentials();
```
lqiu96, line 151: Can this call return an exception since `foobar` is a non-valid universe domain? I think we can keep the verify domain test case below to ensure that the url is constructed properly

cab-token-generator/javatests/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactoryTest.java
```diff
@@ -0,0 +1,575 @@
...
+            .setSourceCredential(sourceCredentials)
+            .setHttpTransportFactory(mockStsTransportFactory)
+            .build();
+
+    try {
+      factory.fetchIntermediateCredentials(); // Expecting an IOException
+      fail("Should fail as the source credential should not be able to be refreshed.");
+    } catch (IOException e) {
+      assertEquals("Unable to refresh the provided source credential.", e.getMessage());
+    }
```
lqiu96, line 178: nit: If possible, can we convert the try .. catch logic with assertThrows()?

cab-token-generator/javatests/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactoryTest.java
```diff
@@ -0,0 +1,575 @@
...
+    IntermediateCredentials intermediateCredentials = factory.fetchIntermediateCredentials();
+    AccessToken intermediateAccessToken = intermediateCredentials.getIntermediateAccessToken();
+
+    assertEquals(
+        mockStsTransportFactory.transport.getAccessToken(),
+        intermediateAccessToken.getTokenValue());
+
+    // Validate that the expires_in has been copied from the source credential.
+    assertEquals(
+        Objects.requireNonNull(sourceCredentials.getAccessToken()).getExpirationTime(),
```
lqiu96, line 205: nit: can you bring this `Objects.requireNonNull(sourceCredentials.getAccessToken())` (I think it would be easier to read).

Perhaps with `assertnotnull`?

cab-token-generator/javatests/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactoryTest.java
```diff
@@ -0,0 +1,575 @@
...
+    // Simulate multiple calls to refreshCredentialsIfRequired. In blocking mode, each call should
+    // trigger a new request to the STS endpoint.
+    int numRefresh = 3;
+    for (int i = 0; i < numRefresh; i++) {
+      factory.refreshCredentialsIfRequired();
+    }
+
+    // Verify that the total number of requests to the STS endpoint is the initial request plus the
+    // number of subsequent refresh calls.
+    assertEquals(1 + numRefresh, mockStsTransportFactory.transport.getRequestCount());
```
lqiu96, line 264: This is only the case for the test because we specify that all subsequent calls will be BLOCKING? IIUC, normally the first request would be blocking and the other three calls would be NONE since the token should be valid?

cab-token-generator/javatests/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactoryTest.java
```diff
@@ -0,0 +1,575 @@
...
+    // necessary because the async task runs on a separate thread.
+    try {
+      Thread.sleep(100);
+    } catch (InterruptedException e) {
+      throw new IOException(e);
+    }
+
+    // After the delay, the request count should be 2, indicating that the async refresh has made
+    // a single request to the STS endpoint.
+    assertEquals(2, mockStsTransportFactory.transport.getRequestCount());
```
lqiu96, line 298: qq, why is this 2 and not 4 as similar to BLOCKING?

cab-token-generator/javatests/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactoryTest.java
```diff
@@ -0,0 +1,575 @@
...
+    }
+
+    // Verify that only the initial call resulted in an immediate STS request. The async refresh
+    // is still pending.
+    assertEquals(1, mockStsTransportFactory.transport.getRequestCount());
+
+    // Introduce a small delay to allow the asynchronous refresh task to complete.  This is
+    // necessary because the async task runs on a separate thread.
+    try {
+      Thread.sleep(100);
```
lqiu96, line 291: can we make this a slightly larger value to reduce any potential flakiness (i.e. 1 sec?)

cab-token-generator/javatests/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactoryTest.java
nbayati: Correct. The async calls will re-use the same refreshTask, so only 1 additional call to the STS is made. I've updated the comment to explain that.

cab-token-generator/javatests/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactoryTest.java
nbayati: Yes, exactly.

cab-token-generator/javatests/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactoryTest.java
nbayati: `ClientSideCredentialAccessBoundaryFactory` doesn't have a way of checking if a domain is valid or not. The only thing that the builder checks is that the domain on factory matches the source credential and we have a test that covers that scenario (`public void builder_universeDomainMismatch_throws()`) Since we're using a mock STS transport factory, it doesn't throw an exception here.

lqiu96: I see. It looks like they're failing at two different spots (one is in the builder and one is in the refresh logic).

If possible, can the `GoogleCredentials` object above be mocked? Mock the `GoogleCredentials.refresh()` call so that it will throw an exception (i.e. url with the custom endpoint turns out to be invalid)?

lqiu96: Oh wait, I see this scenario is covered in `refreshCredentialsIfRequired_sourceCredentialCannotRefresh_throwsIOException`.

cab-token-generator/javatests/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactoryTest.java
```diff
@@ -0,0 +1,573 @@
...
+            .setSourceCredential(sourceCredentials)
+            .setHttpTransportFactory(mockStsTransportFactory)
+            .build();
+
+    try {
+      factory.refreshCredentialsIfRequired(); // Expecting an IOException
+      fail("Should fail as the source credential should not be able to be refreshed.");
+    } catch (IOException e) {
+      assertEquals("Unable to refresh the provided source credential.", e.getMessage());
+    }
```
lqiu96, line 368: Can this be down with assertThrows?

cab-token-generator/javatests/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactoryTest.java
```diff
@@ -0,0 +1,573 @@
...
+  @Test
+  public void builder_noSourceCredential_throws() {
+    try {
+      ClientSideCredentialAccessBoundaryFactory.newBuilder()
+          .setHttpTransportFactory(OAuth2Utils.HTTP_TRANSPORT_FACTORY)
+          .build();
+      fail("Should fail as the source credential is null.");
+    } catch (NullPointerException e) {
+      assertEquals("Source credential must not be null.", e.getMessage());
+    }
```
lqiu96, line 381: assertThrows here as well?

cab-token-generator/javatests/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactoryTest.java
```diff
@@ -0,0 +1,573 @@
...
+          .setSourceCredential(sourceCredentials)
+          .setUniverseDomain("differentUniverseDomain")
+          .build();
+
+      fail("Should fail with universe domain mismatch.");
+    } catch (IllegalArgumentException e) {
+      assertEquals(
+          "The client side access boundary credential's universe domain must be the same as the source credential.",
+          e.getMessage());
+    }
```
lqiu96, line 475: same here if possible

cab-token-generator/javatests/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactoryTest.java
```diff
@@ -0,0 +1,573 @@
...
+                  Thread.currentThread().interrupt();
+                } catch (IOException e) {
+                  throw new RuntimeException(e);
+                }
+              });
+      threads[i].start();
+    }
+    for (Thread thread : threads) {
+      thread.join(); // Wait for each thread to complete
+    }
```
lqiu96, line 571: Can this join() have a timeout? Say like 5 seconds for the tests so they don't run forever in the CI?

lqiu96 (approved):
LGTM

aeitzman (approved):

PR #1586, opened 2024-12-05
feat: add client logging with slf4j

zhumin8:
This Pr contains changes for adding client logging capability to auth.
see [go/java-client-logging-design](http://goto.google.com/java-client-logging-design)

Some logging setups are mirror of same setting in Gax ([pr 3403](https://github.com/googleapis/sdk-platform-java/pull/3403))

Changes includes:
- add slf4j as optional dependency, enable logging only when GOOGLE_SDK_JAVA_LOGGING=true
- user app should add SLF4J + binding dependencies accordingly. For SLF4J 1x, we record extra info with MDC; for SLF4J2x, we record extra info with KeyValuePairs. More user guide to be added via README (see [draft](https://docs.google.com/document/d/1g8HJCstkEEZZc73gFlQO8uPGs07SkWKkr7NG-Yg4bvM/edit?tab=t.0#heading=h.9t58aw3up7to)). 
- If env var not true, or no binding present, default to no-op.
- Add log for request and response made to auth endpoints for UserCredentials, ServiceAccountCredentials, ComputeEngineCredentials, ImpersonatedCredentials. For token values included, added hash in log. Note this PR mainly focuses on adding logging setups, logging statements can be added incrementally later if necessary.

Here is an example logging output for access token request from UserCredentials when DEBUG level is allowed. (TO UPDATE)

```
{"@timestamp":"2024-12-04T21:10:48.596382834-05:00","@version":"1","message":"Sending auth request to refresh access token.","logger_name":"com.google.auth.oauth2.UserCredentials","thread_name":"main","severity":"DEBUG","level_value":10000,"request.method":"POST","request.headers":{"x-goog-api-client":"gl-java/19.0.1 auth/1.32.2-SNAPSHOT cred-type/u","accept-encoding":["gzip"]},"request.url":"https://foo.com/bar","request.payload":{"refresh_token":"bae0258be92ea1d1e14f984507bee05ff4502a29104f4101d22a4a88706b0fc0","grant_type":"refresh_token","client_secret":"2d3c802ef65d75e88b098792e2268cd3f55a08bcbb8c8c4672f1195d1951d4b5","client_id":"ya29.1.AADtN_UtlxN3PuGAxrN2XQnZTVRvDyVWnYq4I6dws"}}
{"@timestamp":"2024-12-04T21:10:48.624914522-05:00","@version":"1","message":"Received auth respond for refresh access token.","logger_name":"com.google.auth.oauth2.UserCredentials","thread_name":"main","severity":"INFO","level_value":20000,"response.status":"200","response.headers":"{}"}
{"@timestamp":"2024-12-04T21:10:48.627483862-05:00","@version":"1","message":"Auth response payload.","logger_name":"com.google.auth.oauth2.UserCredentials","thread_name":"main","severity":"DEBUG","level_value":10000,"access_token":"1/MkS*****KPY2","refresh_token":"1/Tl6*****yGWY","token_type":"Bearer","expires_in":"3600"}
```

Testing setup -  Currently added flavor of tests: 
- unit tests with no extra dependency
- unit tests depending on either binding be present, or logback implementation to capture logging for test
- test for log behaviors for various requests where logs are added. (see LoggingTest)

Remaining issue with test scenarios: 
1. when no binding present: this is tested via regular tests
2.  when env var is T and binding present: tested via test-logging (profile with logback deps)
3. no SLF4J present: same logic as 1, it should be fine with coverage of 1. This is hard to setup this because SLF4J is needed at compile.
4. SLF4J 1x + binding: Hard to setup because SLF4J 2x is needed at compile

One possibility to setup tests for 3 and 4 perhaps is to setup a separate module for testing, but that would require to expose logging classes as public to access outside of package. If no better options, I am inclined to do some manual testing, and do not expose these classes. WDYT?


zhumin8:
@lqiu96  I still have some trouble with the test env, but wanted to put it out there so you can get a preview of the changes. Can you take a quick look and let me know if you have any concerns.

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -92,6 +92,8 @@ public class ComputeEngineCredentials extends GoogleCredentials
   static final Duration COMPUTE_REFRESH_MARGIN = Duration.ofMinutes(3).plusSeconds(45);
 
   private static final Logger LOGGER = Logger.getLogger(ComputeEngineCredentials.class.getName());
+  private static final org.slf4j.Logger SLF4JLOGGER =
+      LoggingConfigs.getLogger(ComputeEngineCredentials.class);
```
lqiu96, line 96: qq, what's the reason why we can't replace LOGGER with the SLF4JLOGGER?

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -361,7 +366,9 @@ private HttpResponse getMetadataResponse(
     request.setThrowExceptionOnExecuteError(false);
     HttpResponse response;
     try {
+      LoggingUtils.logRequest(request, SLF4JLOGGER, "auth get metadata sending request.");
       response = request.execute();
+      LoggingUtils.logResponse(response, SLF4JLOGGER, "auth get metadata received response.");
```
lqiu96, line 371: Is it possible for the request to be more specific for this? Auth sending request for access token, for universe domain, for ...? I think `sending metadata...` is a bit generic and may not be clear as logs

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -461,7 +468,10 @@ private static boolean pingComputeEngineMetadata(
             request,
             MetricsUtils.getGoogleCredentialsMetricsHeader(
                 RequestType.METADATA_SERVER_PING, CredentialTypeForMetrics.DO_NOT_SEND));
+
+        LoggingUtils.logRequest(request, SLF4JLOGGER, "auth pin MDS.");
         HttpResponse response = request.execute();
+        LoggingUtils.logResponse(response, SLF4JLOGGER, "auth received response from MDS.");
```
lqiu96, line 474: nit: Spell out MDS as it may not be super clear what the acronym is

oauth2_http/java/com/google/auth/oauth2/IamUtils.java
```diff
@@ -142,7 +144,9 @@ private static String getSignature(
                     IamUtils.IAM_RETRYABLE_STATUS_CODES.contains(response.getStatusCode())));
     request.setIOExceptionHandler(new HttpBackOffIOExceptionHandler(backoff));
 
+    LoggingUtils.logRequest(request, LOGGER, "auth sending request to get signature.");
     HttpResponse response = request.execute();
+    LoggingUtils.logResponse(response, LOGGER, "auth received response for signature.");
```
lqiu96, line 149: nit: Could the message to be more descriptive? i.e. Signature to sign the blob or something that better describes this

oauth2_http/java/com/google/auth/oauth2/IamUtils.java
```diff
@@ -169,6 +173,7 @@ private static String getSignature(
     }
 
     GenericData responseData = response.parseAs(GenericData.class);
+    LoggingUtils.logGenericData(responseData, LOGGER, "Auth response payload.");
```
lqiu96, line 176: nit: a bit more descriptive like like Sign Blob response payload

oauth2_http/java/com/google/auth/oauth2/LoggingConfigs.java
```diff
@@ -0,0 +1,50 @@
...
+
+  private static EnvironmentProvider environmentProvider = SystemEnvironmentProvider.getInstance();
+  private static final Logger NO_OP_LOGGER = org.slf4j.helpers.NOPLogger.NOP_LOGGER;
+  private static boolean loggingEnabled = isLoggingEnabled();
+  // expose this setter only for testing purposes
+  static void setEnvironmentProvider(EnvironmentProvider provider) {
+    environmentProvider = provider;
+    // Recalculate LOGGING_ENABLED after setting the new provider
+    loggingEnabled = isLoggingEnabled();
+  }
```
lqiu96, line 17: I see. Having these methods as static makes this harder. Not the biggest fan of this and almost would rather us do the maven profile workaround to test this.

It's package-private and used only for testing. We can try and address this last.

oauth2_http/java/com/google/auth/oauth2/LoggingConfigs.java
```diff
@@ -0,0 +1,50 @@
...
+
+  // constructor with LoggerFactoryProvider to make testing easier
+  static Logger getLogger(Class<?> clazz, LoggerFactoryProvider factoryProvider) {
+    if (loggingEnabled) {
+      return factoryProvider.getLoggerFactory().getLogger(clazz.getName());
+    } else {
+      //  use SLF4j's NOP logger regardless of bindings
+      return NO_OP_LOGGER;
+    }
```
lqiu96, line 32: nit:
```suggestion
    if (!loggingEnabled) {
      //  use SLF4j's NOP logger regardless of bindings
      return NO_OP_LOGGER;
    } 
    return factoryProvider.getLoggerFactory().getLogger(clazz.getName());
```

oauth2_http/java/com/google/auth/oauth2/LoggingUtils.java
```diff
@@ -0,0 +1,155 @@
...
+        logger.warn(message);
+        break;
+      case ERROR:
+        logger.error(message);
+        break;
+      default:
+        logger.info(message);
+        // Default to INFO level
+    }
+    MDC.clear();
```
lqiu96, line 42: why does this need to be cleared?

oauth2_http/java/com/google/auth/oauth2/LoggingUtils.java
```diff
@@ -0,0 +1,155 @@
...
+        logWithMDC(logger, org.slf4j.event.Level.INFO, responseLogDataMap, message);
+      } catch (Exception e) {
+
+        logger.error("Error logging response: ", e);
+      }
+    }
+  }
+
+  static void logGenericData(GenericData genericData, Logger logger, String message) {
+    if (logger.isDebugEnabled()) {
```
lqiu96, line 103: nit: `if (!logger.isDebugEnabled()) {` to reduce nesting

oauth2_http/java/com/google/auth/oauth2/LoggingUtils.java
```diff
@@ -0,0 +1,155 @@
...
+
+        Map<String, String> contextMap = parseGenericData(genericData);
+        logWithMDC(logger, org.slf4j.event.Level.DEBUG, contextMap, message);
+      } catch (Exception e) {
+        logger.error("Error logging GenericData: ", e);
+      }
+    }
+  }
+
+  private static Map<String, String> parseGenericData(GenericData genericData) {
```
lqiu96, line 114: I believe genericData can be null when we parse it from the response: https://github.com/googleapis/google-http-java-client/blob/b3f12421026ac3968e5682cf0cdd74a2eadae795/google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java#L457-L459

We may need to guard against it.

oauth2_http/java/com/google/auth/oauth2/LoggingUtils.java
```diff
@@ -0,0 +1,155 @@
...
+    Map<String, String> contextMap = new HashMap<>();
+    genericData.forEach(
+        (key, val) -> {
+          if ("token".equals(key)
+              || "assertion".equals(key)
+              || "access_token".equals(key)
+              || "client_secret".equals(key)
+              || "refresh_token".equals(key)) {
```
lqiu96, line 122: Perhaps we can have these in a list + authorization so that we can do something like `if (list.contains(key))`, then hash the value?

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -489,6 +491,11 @@ boolean isConfiguredForDomainWideDelegation() {
     return serviceAccountUser != null && serviceAccountUser.length() > 0;
   }
 
+  private GenericData parseResponseAs(HttpResponse response) throws IOException {
+    GenericData genericData = response.parseAs(GenericData.class);
+    LoggingUtils.logGenericData(genericData, LOGGER, "Auth response payload.");
+    return genericData;
+  }
```
lqiu96, line 498: Can we move this to LoggingUtils since this seems to be used in the other credentials?

oauth2_http/java/com/google/auth/oauth2/LoggingUtils.java
zhumin8: MDC carries contextual information in log messages. It is tied to thread, and is safer to clear it to avoid it be carried over to other log entries.

oauth2_http/java/com/google/auth/oauth2/LoggingUtils.java
zhumin8: Thanks for the context. Yes, currently usage of parseGenericData() are both guarded with try-catch (e.g. L106 above).

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
zhumin8: They are slightly different, but I will try to extract any useful shared logics into LoggingUtils.

oauth2_http/java/com/google/auth/oauth2/LoggingConfigs.java
zhumin8: Due to time constraints, I am leaning towards keep this logic for now, and make necessary changes in the next iteration/release if possible.

That said, can you share your concerns about this?

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
zhumin8: Yes, I will take another pass on these messages and try to make them more clear.
Will tag to for another round when I update them.

oauth2_http/java/com/google/auth/oauth2/LoggingUtils.java
zhumin8: In case, try-catch block is to ensure logging failures do not cause interruption to user application. This if statement should be inside though.

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
lqiu96: Gotcha, do you know if there is any impact of having these two Loggers working together in the same class? Would they have two different logging formats show up in someone is using ComputeEngineCredentials?

oauth2_http/java/com/google/auth/oauth2/LoggingConfigs.java
lqiu96: The main concern is just having a setter that potentially changes the state of the logging. Calling `setEnvironmentProvider()` ends up re-calculating if logging is enabled.

For now, it's package private and would only be called be us (for now it's only used for tests). I think it's fine for now but ideally I don't think it's something that we should have

oauth2_http/java/com/google/auth/oauth2/LoggingUtils.java
lqiu96: Using the L106 example above: If I'm understanding the flow above correctly, this would fail as a NPE when `genericData.forEach()` runs in L116.

I don't think an exception with NPE being returned is what we want to return to the user. I think we should catch the NPE and return a better error message: `No response body found in the response, etc.`

Additionally, I think this also would happen in the `parseResponseAs` method in ServiceAccountCredentials. An NPE would be caught and thrown back as an exception.

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
lqiu96: Can you highlight the difference? From what I see, the other Credentials also end up calling `GenericData responseData = response.parseAs(GenericData.class);` but this is moved to a function.

Might be missing something, but I don't see what is preventing this from going into LoggingUtils.

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
zhumin8: They will each show up according to the correponding logging config (formatting, sink ...). They do not interfere with each other.
That said, this is a temporary stage , I will migrate the JUL logging in next release.

oauth2_http/java/com/google/auth/oauth2/LoggingUtils.java
zhumin8: This could be future improvements. For now, what I want to ensure is logging should not cause user application to crash.

zhumin8:
sonar has been failing since Dec 2. Filed
https://github.com/googleapis/google-auth-library-java/issues/1591

oauth2_http/java/com/google/auth/oauth2/LoggingUtils.java
```diff
@@ -0,0 +1,207 @@
...
+    if (!contextMap.isEmpty()) {
+      MDC.clear();
+    }
+  }
+
+  static void logRequest(HttpRequest request, Logger logger, String message) {
+    try {
+      if (logger.isInfoEnabled()) {
+        Map<String, String> loggingDataMap = new HashMap<>();
+        loggingDataMap.put("request.method", request.getRequestMethod());
```
blakeli0, line 102: I think it would be good to duplicate information like `request.method` and `request.url` in message instead of always relying on MDC, as we discussed yesterday. It would be great if we can make this change in this release. However, if not, it should not be a blocker either, as adding these info later is not a breaking change.

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
lqiu96: Ok sounds good to me then.

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -361,7 +366,23 @@ private HttpResponse getMetadataResponse(
...
+        responseMessage = "Auth get metadata received response for id token";
+      } else if (requestType.equals(RequestType.ACCESS_TOKEN_REQUEST)) {
+        requestMessage = "Auth get metadata sending request for access token";
+        responseMessage = "Auth get metadata received response for access token";
+      } else {
+        // TODO: this includes get universe domain and get default sa.
+        // refactor for more clear logging message.
+        requestMessage = "Auth get metadata sending request";
+        responseMessage = "Auth get metadata received response";
+      }
```
lqiu96, line 382: nit: How about `Sending request for ...` and `Received response for ...` without the `Auth get metadata` prefix? 

I don't think we need the specify it's Auth (unless that's required) and `get metadata` seems to be a implementation (i.e. private method) that is calling it

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -461,7 +482,11 @@ private static boolean pingComputeEngineMetadata(
             request,
             MetricsUtils.getGoogleCredentialsMetricsHeader(
                 RequestType.METADATA_SERVER_PING, CredentialTypeForMetrics.DO_NOT_SEND));
+
+        LoggingUtils.logRequest(request, SLF4JLOGGER, "Pin auth Metadata Server");
```
lqiu96, line 486:
```suggestion
        LoggingUtils.logRequest(request, SLF4JLOGGER, "Pinging Metadata Server");
```

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -461,7 +482,11 @@ private static boolean pingComputeEngineMetadata(
...
         HttpResponse response = request.execute();
+        LoggingUtils.logResponse(
+            response, SLF4JLOGGER, "Received auth response from Metadata Server");
```
lqiu96, line 489: nit: Thoughts on dropping `auth`? Just `Received response from Metadata Server`?

oauth2_http/java/com/google/auth/oauth2/LoggingConfigs.java
lqiu96: Sounds good. It's package-private and we should be able to change easily in the future if something better comes up.

oauth2_http/java/com/google/auth/oauth2/LoggingConfigs.java
```diff
@@ -0,0 +1,80 @@
...
+  static Logger getLogger(Class<?> clazz, LoggerFactoryProvider factoryProvider) {
+    if (!loggingEnabled) {
+      //  use SLF4j's NOP logger regardless of bindings
+      return NO_OP_LOGGER;
+    }
+    return factoryProvider.getLoggerFactory().getLogger(clazz.getName());
+  }
+
+  static boolean isLoggingEnabled() {
+    String enableLogging = environmentProvider.getEnv("GOOGLE_SDK_JAVA_LOGGING");
```
lqiu96, line 66: Can we make this a env var since it's used by a few test classes as well?

oauth2_http/java/com/google/auth/oauth2/LoggingUtils.java
```diff
@@ -0,0 +1,207 @@
...
+        loggingDataMap.put("request.url", request.getUrl().toString());
+
+        Map<String, Object> headers = new HashMap<>();
+        request
+            .getHeaders()
+            .forEach(
+                (key, val) -> {
+                  if ("authorization".equals(key)) {
```
lqiu96, line 110: Can you confirm if it's `authorization` or `Authorization` (capitalization)? Also, can this be added to the sensitiveKeys list?

oauth2_http/java/com/google/auth/oauth2/LoggingUtils.java
```diff
@@ -0,0 +1,207 @@
...
+                    headers.put(key, val);
+                  }
+                });
+        loggingDataMap.put("request.headers", headers.toString());
+
+        if (request.getContent() != null && logger.isDebugEnabled()) {
+          // are payload always GenericData? If so, can parse and store in json
+          if (request.getContent() instanceof UrlEncodedContent) {
+
+            GenericData data = (GenericData) ((UrlEncodedContent) request.getContent()).getData();
```
lqiu96, line 123: Can you add a quick comment about why UrlEncodedContent is casted to GenericData? But JsonHttpContent (below) doesn't need to be?

oauth2_http/java/com/google/auth/oauth2/LoggingUtils.java
lqiu96: Ok, that's fine for now. Let's fix it in a future PR so we can provide more in-depth and helpful logs.

oauth2_http/java/com/google/auth/oauth2/LoggingUtils.java
```diff
@@ -0,0 +1,207 @@
...
+        responseLogDataMap.put("response.headers", headers.toString());
+        logWithMDC(logger, org.slf4j.event.Level.INFO, responseLogDataMap, message);
+      }
+    } catch (Exception e) {
+
+      logger.error("Error logging response: ", e);
+    }
+  }
+
+  static void logGenericData(GenericData genericData, Logger logger, String message) {
```
lqiu96, line 159: nit: Can we change this to be something like `logResponsePayload` and `parseResponsePayload` so it's clearer as to what it's doing?

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -489,6 +491,11 @@ boolean isConfiguredForDomainWideDelegation() {
     return serviceAccountUser != null && serviceAccountUser.length() > 0;
   }
 
+  private GenericData parseResponseAs(HttpResponse response) throws IOException {
+    GenericData genericData = response.parseAs(GenericData.class);
+    LoggingUtils.logGenericData(genericData, LOGGER, "Auth response payload");
+    return genericData;
+  }
```
lqiu96, line 498: qq, isn't this the same thing as `LoggingUtils.logGenericData()`? I'm missing why this method needs to exist

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
lqiu96: At least I believe the logs will show up as something like `INFO  com.google.auth.ComputeEngineCredentials -- Sending request for ID Token` or something similar

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
zhumin8: This `response.parseAs(GenericData.class);` can only be parsed once. So this is a wrapper to parse the data, log and continue with whatever is to do with the genericData.

oauth2_http/java/com/google/auth/oauth2/LoggingUtils.java
zhumin8: It's already done. (9e20be9390d0343496d32977a6ff9297c3fae615)
It is verified in LoggingTest ([here](https://github.com/googleapis/google-auth-library-java/blob/15d301675ac10bc61d841fc615a70f88b87369b5/oauth2_http/javatests/com/google/auth/oauth2/LoggingTest.java#L126-L128) you can see the message is parsed then verified. The test is not extensive in verifying information. )

oauth2_http/java/com/google/auth/oauth2/LoggingUtils.java
zhumin8: Double-checked, it's `authorization`. 
But also it does not hurt to make this comparison case agnostic. making the change.

oauth2_http/java/com/google/auth/oauth2/LoggingUtils.java
zhumin8: added

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
lqiu96: Apologies, I might be missing something obvious but I'm not see where it's parsed more than once and might as be missing why it's different for ServiceAccountCredentials.

What I see is that it's changed to be:
```
GenericData responseData = parseResponseAs(response);
```

but I think it could just be
```
GenericData responseData = response.parseAs(GenericData.class);
LoggingUtils.logGenericData(responseData, LOGGER, "Auth response payload");
...
```
which is the same as it is for the other Credential types (i.e. https://github.com/googleapis/google-auth-library-java/blob/64855267100a73398c9050670655fb578f4d3b59/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java#L300-L302)

oauth2_http/java/com/google/auth/oauth2/LoggingUtils.java
```diff
@@ -0,0 +1,211 @@
...
+
+          logWithMDC(logger, org.slf4j.event.Level.INFO, loggingDataMap, message);
+        }
+      }
+    } catch (Exception e) {
+      logger.error("Error logging request: ", e);
+    }
+  }
+
+  static void logResponse(HttpResponse response, Logger logger, String message) {
```
blakeli0, line 146: In general, I think we should not try to call these methods if logging is not enabled due to performance concerns, similar to the [comment in sdk-platform-java](https://github.com/googleapis/sdk-platform-java/pull/3403/files#r1881141618).

oauth2_http/java/com/google/auth/oauth2/LoggingUtils.java
```diff
@@ -0,0 +1,211 @@
...
+      if (logger.isInfoEnabled()) {
+        Map<String, String> responseLogDataMap = new HashMap<>();
+        responseLogDataMap.put("response.status", String.valueOf(response.getStatusCode()));
+        responseLogDataMap.put("response.status.message", response.getStatusMessage());
+
+        Map<String, Object> headers = new HashMap<>(response.getHeaders());
+        responseLogDataMap.put("response.headers", headers.toString());
+        logWithMDC(logger, org.slf4j.event.Level.INFO, responseLogDataMap, message);
+      }
+    } catch (Exception e) {
```
blakeli0, line 157: Similar to the [comment in sdk-platform-java](https://github.com/googleapis/sdk-platform-java/pull/3403/files#r1881114202), I think we should try to avoid doing these try catches or fail more gracefully.

zhumin8:
Hi @lqiu96, I am still figuring out some of the CI issues. But in the interests of time, perhaps you can start taking a look at this?
There's a few classes are pretty much mirrored from Gax (except used `EnvironmentProvider` for testing purposes that's already in this repo)

One thing I notice, aside from sonar, we also have jacoco configured. It is giving some trouble with coverage not met, but as far as I see now, it complains about coverage in classes I did not change. I will take another look, but currently lean towards skipping jacoco for most of the ci tests.

.github/workflows/ci.yaml
```diff
@@ -36,6 +36,7 @@ jobs:
     - run: .kokoro/build.sh
       env:
         JOB_TYPE: test
+        GOOGLE_SDK_JAVA_LOGGING: true
```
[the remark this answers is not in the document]
zhumin8: obsolete.

oauth2_http/java/com/google/auth/oauth2/LoggingUtils.java
zhumin8: update: based on offline conversations, we decided to not duplicate information in message.

oauth2_http/java/com/google/auth/oauth2/LoggingUtils.java
zhumin8: as discussed offline, we decided to stick with try catches to let logging fail silently for any logging exceptions.

pom.xml
```diff
@@ -419,6 +421,31 @@
   </reporting>
 
   <profiles>
+    <profile>
+      <id>slf4j2x-test</id>
```
JoeWang1127, line 425: Perhaps this is answered previously but why do we need to add a profile in the parent module if a same profile is added in the child's module?

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
zhumin8: Hope https://github.com/googleapis/google-auth-library-java/pull/1586#discussion_r1956324685 answers this question too.

oauth2_http/javatests/com/google/auth/oauth2/Slf4jUtilsTest.java
```diff
@@ -0,0 +1,101 @@
...
+    Logger logger = Slf4jUtils.getLogger(Slf4jUtilsTest.class);
+
+    assertEquals(NOPLogger.class, logger.getClass());
+    assertFalse(logger.isInfoEnabled());
+    assertFalse(logger.isDebugEnabled());
+  }
+
+  // This test require binding (e.g. logback) be present
+  @Test
+  public void testGetLogger_loggingEnabled_slf4jBindingPresent() {
```
JoeWang1127, line 70: From the test result, I can see the binding is not present.

Could you change the test name to reflect that?

oauth2_http/javatests/com/google/auth/oauth2/Slf4jUtilsLogbackTest.java
```diff
@@ -0,0 +1,218 @@
...
+    TestAppender testAppender = setupTestLogger();
+    Slf4jUtils.logWithMDC(LOGGER, Level.INFO, new HashMap<>(), "test message");
+
+    assertEquals(1, testAppender.events.size());
+    assertEquals(ch.qos.logback.classic.Level.INFO, testAppender.events.get(0).getLevel());
+    testAppender.stop();
+  }
+
+  @Test
+  public void testLogWithMDC_TRACE() {
```
JoeWang1127, line 104: Looks like this test only verifies event size, could you change the test name?

pom.xml
zhumin8: I am still figuring this out, but the `maven-compiler-plugin` testInclude needs to be in the parent pom, otherwise compile would fail. My understanding is test classes from other modules (`cab-token-generator` in this case) tries to compile, and referencing to oauth classes that is not compiled from the module.

oauth2_http/javatests/com/google/auth/oauth2/Slf4jUtilsTest.java
zhumin8: I am not sure I get what you mean?

This test is only included when `slf4j2x-test` is active (logback present), so I expect`assertNotEquals(NOPLogger.class, logger.getClass());`

pom.xml
JoeWang1127: Could we at least remove the `testIncludes` in either parent or child module?

oauth2_http/javatests/com/google/auth/oauth2/Slf4jUtilsTest.java
JoeWang1127: Ah, my bad, I thought it's `assertEquals`.

oauth2_http/pom.xml
```diff
@@ -68,8 +68,113 @@
...
+          <artifactId>gson</artifactId>
+          <version>${project.gson.version}</version>
+        </dependency>
+      </dependencies>
+    </profile>
+    <profile>
+      <id>slf4j2x-test</id>
+      <!--   This profile is for logging test only,
+      use instead of the default slf4j2x.
+      This should merge with parent modules same name profile-->
```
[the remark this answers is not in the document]
zhumin8: done in [5d6eeb0](https://github.com/googleapis/google-auth-library-java/pull/1586/commits/5d6eeb0d6a6685f38501060674fea9b31ad4226c)

oauth2_http/java/com/google/auth/oauth2/LoggingUtils.java
zhumin8: done in [5bcd4ee](https://github.com/googleapis/google-auth-library-java/pull/1586/commits/5bcd4ee83b4a8a2ff7790d1b6382715cbc1d93ec)

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -551,7 +573,11 @@ private static boolean pingComputeEngineMetadata(
             request,
             MetricsUtils.getGoogleCredentialsMetricsHeader(
                 RequestType.METADATA_SERVER_PING, CredentialTypeForMetrics.DO_NOT_SEND));
+
+        LoggingUtils.logRequest(request, LOGGER_PROVIDER, "Pinging Metadata Server");
         HttpResponse response = request.execute();
+        LoggingUtils.logResponse(
+            response, LOGGER_PROVIDER, "Received response from Metadata Server");
```
lqiu96, line 580: IIRC, `pingComputeEngineMetadata` is called as part of one of the ADC checks. Would it be possible to log this only for debug/ trace level?

oauth2_http/java/com/google/auth/oauth2/Slf4jLoggingHelpers.java
```diff
@@ -0,0 +1,180 @@
...
+
+          Slf4jUtils.log(logger, org.slf4j.event.Level.INFO, loggingDataMap, message);
+        }
+      }
+    } catch (Exception e) {
+      // let logging fail silently
+    }
+  }
+
+  static void logResponse(HttpResponse response, LoggerProvider loggerProvider, String message) {
```
lqiu96, line 112: Can you add comments/ javadocs above that explain the difference between `logResponse` vs `logResponsePayload` that that generally both are invoked one after another.

something like: 
LogResponse -> Status Code, Message from Server, Headers
LogResponesPayload -> The parsed json response

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -430,6 +435,7 @@ public IdToken idTokenWithAudience(String targetAudience, List<IdTokenProvider.O
       throw new IOException(METADATA_RESPONSE_EMPTY_CONTENT_ERROR_MESSAGE);
     }
     String rawToken = response.parseAsString();
+    // TODO: log payload here
```
lqiu96, line 438: FYI in case this was forgotten

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -611,9 +621,13 @@ private IdToken getIdTokenOauthEndpoint(String targetAudience) throws IOExceptio
         MetricsUtils.getGoogleCredentialsMetricsHeader(
             RequestType.ID_TOKEN_REQUEST, getMetricsCredentialType()));
 
+    LoggingUtils.logRequest(
+        request, LOGGER_PROVIDER, "Sending request to get ID token via Oauth endpoint");
```
lqiu96, line 625:
```suggestion
        request, LOGGER_PROVIDER, "Sending request to get ID token");
```

I think it would be better to not include the internal implementation details. I think it would be better to keep it as `... retrieving ID Token` or similar.

oauth2_http/java/com/google/auth/oauth2/Slf4jLoggingHelpers.java
```diff
@@ -0,0 +1,180 @@
...
+                  if (SENSITIVE_KEYS.contains(key)) {
+                    String hashedVal = calculateSHA256Hash(String.valueOf(val));
+                    headers.put(key, hashedVal);
+                  } else {
+                    headers.put(key, val);
+                  }
+                });
+        loggingDataMap.put("request.headers", gson.toJson(headers));
+
+        if (request.getContent() != null && logger.isDebugEnabled()) {
```
lqiu96, line 89: qq, if the request doesn't have a payload does it show up as null or as [EmptyContent](https://github.com/googleapis/google-http-java-client/blob/main/google-http-client/src/main/java/com/google/api/client/http/EmptyContent.java)?

Also, to confirm this: we only will log the content if the logger is debug enabled?

oauth2_http/java/com/google/auth/oauth2/Slf4jLoggingHelpers.java
```diff
@@ -0,0 +1,180 @@
...
+    }
+  }
+
+  static void logResponse(HttpResponse response, LoggerProvider loggerProvider, String message) {
+    try {
+      Logger logger = loggerProvider.getLogger();
+      if (logger.isInfoEnabled()) {
+        Map<String, Object> responseLogDataMap = new HashMap<>();
+        responseLogDataMap.put("response.status", String.valueOf(response.getStatusCode()));
+        responseLogDataMap.put("response.status.message", response.getStatusMessage());
```
lqiu96, line 118: qq, the response message is nested as a field of status? Or should this be `request.message`?

oauth2_http/java/com/google/auth/oauth2/Slf4jLoggingHelpers.java
```diff
@@ -0,0 +1,180 @@
...
+      MessageDigest digest = MessageDigest.getInstance("SHA-256");
+      byte[] inputBytes = data.getBytes(StandardCharsets.UTF_8);
+      byte[] hashBytes = digest.digest(inputBytes);
+      return bytesToHex(hashBytes);
+    } catch (NoSuchAlgorithmException e) {
+      return "Error calculating SHA-256 hash."; // do not fail for logging failures
+    }
+  }
+
+  private static String bytesToHex(byte[] hash) {
```
lqiu96, line 169: Could you add a blurb to generalize what this is doing. Converting this to a hex value and XOR's 0xff (255?) with each byte because of reason X

oauth2_http/java/com/google/auth/oauth2/Slf4jUtils.java
```diff
@@ -0,0 +1,164 @@
...
+
+/** Contains util methods to get SLF4J logger and log conditionally based SLF4J major version */
+class Slf4jUtils {
+
+  private static final Logger NO_OP_LOGGER = org.slf4j.helpers.NOPLogger.NOP_LOGGER;
+  private static final Gson gson = new Gson();
+  private static boolean isSLF4J2x;
+
+  static {
+    isSLF4J2x = checkIfClazzAvailable("org.slf4j.event.KeyValuePair");
```
lqiu96, line 51: I'm assuming this class was chosen because it's found in 2.x and 1.x. If so, could you add a quick comment about that here.

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -451,7 +457,23 @@ private HttpResponse getMetadataResponse(
...
+        requestMessage = "Sending request to get ID token";
+        responseMessage = "Received response for ID token request";
+      } else if (requestType.equals(RequestType.ACCESS_TOKEN_REQUEST)) {
+        requestMessage = "Sending request to refresh access token";
+        responseMessage = "Received response for refresh access token";
+      } else {
+        // TODO: this includes get universe domain and get default sa.
+        // refactor for more clear logging message.
+        requestMessage = "Sending request for universe domain/default service account";
+        responseMessage = "Received response for universe domain/default service account";
```
lqiu96, line 472: qq, I see that this takes in a `shouldSendMetricsHeader` header.

Thoughts on perhaps creating our internal RequestType values (one for Universe Domain and one for Service Account). It could differentiate between UD and SA requests, but also not track the metrics header that can be used for logging.

oauth2_http/java/com/google/auth/oauth2/Slf4jLoggingHelpers.java
zhumin8: Good question. We have `request.getContent() == null` for GET requests. EmptyContent seems to be used in PUT or POST to force an empty content ([ref](https://github.com/googleapis/google-http-java-client/blob/3ee8bb6380ee22a8a6eb6e5048c603fbdb9ee320/google-http-client/src/main/java/com/google/api/client/http/EmptyContent.java#L25)). 

> we only will log the content if the logger is debug enabled?

Yes, this is more detailed info. Again, we can modify this if needed.

oauth2_http/java/com/google/auth/oauth2/Slf4jLoggingHelpers.java
zhumin8: This is the status message, not a response message:
https://github.com/googleapis/google-http-java-client/blob/3ee8bb6380ee22a8a6eb6e5048c603fbdb9ee320/google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java#L75-L76

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
zhumin8: hmm, I feel like this could be helpful information for debugging. 
Also, fyi - we currently publish client debug logging with disclaimer. And we will be able to make changes to these log messages without breaking change.

oauth2_http/java/com/google/auth/oauth2/Slf4jLoggingHelpers.java
zhumin8: [e1d9a78](https://github.com/googleapis/google-auth-library-java/pull/1586/commits/e1d9a78d76f037690e87a7fbbe8314f0b4f3cbf8)

oauth2_http/java/com/google/auth/oauth2/Slf4jUtils.java
zhumin8: [e1d9a78](https://github.com/googleapis/google-auth-library-java/pull/1586/commits/e1d9a78d76f037690e87a7fbbe8314f0b4f3cbf8)

zhumin8: [e1d9a78](https://github.com/googleapis/google-auth-library-java/pull/1586/commits/e1d9a78d76f037690e87a7fbbe8314f0b4f3cbf8)

oauth2_http/java/com/google/auth/oauth2/Slf4jLoggingHelpers.java
zhumin8: [e1d9a78](https://github.com/googleapis/google-auth-library-java/pull/1586/commits/e1d9a78d76f037690e87a7fbbe8314f0b4f3cbf8)

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
zhumin8: done in [eaca20b](https://github.com/googleapis/google-auth-library-java/pull/1586/commits/eaca20b1718184eb0f44deed640ef9f71937b977)

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
zhumin8: yea, I think that's a valid improvement. However, given the size of this PR, I'd rather file as a separate issue as clean-up/enhancement.

oauth2_http/java/com/google/auth/oauth2/IamUtils.java
```diff
@@ -154,7 +155,11 @@ private static String getSignature(
                     IamUtils.IAM_RETRYABLE_STATUS_CODES.contains(response.getStatusCode())));
     request.setIOExceptionHandler(new HttpBackOffIOExceptionHandler(backoff));
 
+    LoggingUtils.logRequest(
+        request, LOGGER_PROVIDER, "Sending request to get signature to sign the blob");
     HttpResponse response = request.execute();
+    LoggingUtils.logResponse(
+        response, LOGGER_PROVIDER, "Received response for signature to sign the blob");
```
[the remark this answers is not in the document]
zhumin8: hmm, usually the convention is to logging in the class where it occurs. Although to your point, we can probably add another log entry before entering the `IamUtils` methods, but I am not sure if that's too much details?

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
lqiu96: Users don't get to configure any settings to choose whether to hit the Oauth or IAM endpoint. It's an implementation detail and we decide based on the UD value of the Credentials. As long as users see that it's trying to retrieve the ID token, I think that's should suffice for the log.

The `request.url` value should be populated as part of the structure log IIUC and that could be helpful to determine exactly which endpoint is sent.

oauth2_http/java/com/google/auth/oauth2/Slf4jLoggingHelpers.java
lqiu96: > EmptyContent seems to be used in PUT or POST to force an empty content

I don't see any references in the Auth Library for EmptyContent so I think it's fine now. I don't think we need to add a case for something that we may never use.

oauth2_http/java/com/google/auth/oauth2/IamUtils.java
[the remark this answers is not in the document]
lqiu96: Ok gotcha I see. In the case I think it's fine. Maybe if anything it could be something like `Sending request to get signature to sign the blob for %s, Credentials.getClass())`. That also might be overkill. I think we can keep as is and improve it from user feedback.

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
```diff
@@ -430,6 +436,12 @@ public IdToken idTokenWithAudience(String targetAudience, List<IdTokenProvider.O
       throw new IOException(METADATA_RESPONSE_EMPTY_CONTENT_ERROR_MESSAGE);
     }
     String rawToken = response.parseAsString();
+
+    LoggingUtils.log(
+        LOGGER_PROVIDER,
+        Level.FINE,
+        ImmutableMap.of("id-token", rawToken),
```
lqiu96, line 443: Is there a naming convention for the key here? hyphens to seperate the words, camelCase?

lqiu96 (approved):
LGTM. Added a few comment.

oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java
zhumin8: Good point. I don't think there is not a generic naming convention. However, I agree we should keep consistent as possible. Following most camelCases in [this doc](https://docs.google.com/document/d/1nh1m0o8RADN7wx1w4GFtIa5vpLOIHlNOy9e0ufXItks/edit?resourcekey=0-0dGUiO68s6aMOAHvrMYhYg&tab=t.0#heading=h.swusf33l2yl2), I am going to change this to "idToken"

PR #1589, opened 2024-12-09
fix: JSON parsing of S2A addresses.

rmehta19:
Autoconfig endpoint returns
```
{
  s2a:
  {
    plaintext_address: ""
    mtls_address: "" 
  }
}
```

not 
```
{
  plaintext_address: ""
  mtls_address: "" 
}
```

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/SecureSessionAgent.java
+++ b/oauth2_http/java/com/google/auth/oauth2/SecureSessionAgent.java
@@ -190,7 +190,7 @@ public class SecureSessionAgent {
 
     String plaintextS2AAddress = "";
     String mtlsS2AAddress = "";
-    Object s2aAddressConfig = responseData.get(S2A_JSON_KEY);
+    Map<String, Object> s2aAddressConfig = (Map<String, Object>) responseData.get(S2A_JSON_KEY);
     if (s2aAddressConfig == null) {
       /*
        * Return empty addresses in {@link SecureSessionAgentConfig} if endpoint doesn't return anything.
@@ -200,9 +200,7 @@ public class SecureSessionAgent {
     try {
       plaintextS2AAddress =
           OAuth2Utils.validateString(
-              (Map<String, Object>) s2aAddressConfig,
-              S2A_PLAINTEXT_ADDRESS_JSON_KEY,
-              PARSE_ERROR_S2A);
+              s2aAddressConfig, S2A_PLAINTEXT_ADDRESS_JSON_KEY, PARSE_ERROR_S2A);
     } catch (IOException ignore) {
       /*
        * Do not throw error because of parsing error, just leave the address as empty in {@link SecureSessionAgentConfig}.
@@ -210,8 +208,7 @@ public class SecureSessionAgent {
     }
     try {
       mtlsS2AAddress =
-          OAuth2Utils.validateString(
-              (Map<String, Object>) s2aAddressConfig, S2A_MTLS_ADDRESS_JSON_KEY, PARSE_ERROR_S2A);
+          OAuth2Utils.validateString(s2aAddressConfig, S2A_MTLS_ADDRESS_JSON_KEY, PARSE_ERROR_S2A);
     } catch (IOException ignore) {
       /*
        * Do not throw error because of parsing error, just leave the address as empty in {@link SecureSessionAgentConfig}.
--- a/oauth2_http/java/com/google/auth/oauth2/SecureSessionAgent.java
+++ b/oauth2_http/java/com/google/auth/oauth2/SecureSessionAgent.java
@@ -45,6 +45,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.util.Arrays;
 import java.util.HashSet;
+import java.util.Map;
 import java.util.ServiceLoader;
 import java.util.Set;
 import javax.annotation.concurrent.ThreadSafe;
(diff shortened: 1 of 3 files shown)


rmehta19:
@lqiu96 , @zhumin8, please review this PR, thanks!

oauth2_http/java/com/google/auth/oauth2/SecureSessionAgent.java
```diff
@@ -210,4 +211,23 @@ private SecureSessionAgentConfig getSecureSessionAgentConfigFromMDS() {
...
+      throws IOException {
+    Object value = map.get(S2A_JSON_KEY);
+    if (value == null) {
+      throw new IOException(
+          String.format(OAuth2Utils.VALUE_NOT_FOUND_MESSAGE, errorPrefix, S2A_JSON_KEY));
+    }
+    if (!(value instanceof Map)) {
+      throw new IOException(
+          String.format(OAuth2Utils.VALUE_WRONG_TYPE_MESSAGE, errorPrefix, "Map", S2A_JSON_KEY));
+    }
```
lqiu96, line 225: qq, is this check necessary? Is it possible that it may be returned as an array? I was under the assumption that we had control over this API. Or is this just intended as a defensive check?

oauth2_http/java/com/google/auth/oauth2/SecureSessionAgent.java
```diff
@@ -210,4 +211,23 @@ private SecureSessionAgentConfig getSecureSessionAgentConfigFromMDS() {
...
+    Object address = ((Map<String, Object>) value).get(key);
+    if (!(address instanceof String)) {
+      throw new IOException(
+          String.format(OAuth2Utils.VALUE_WRONG_TYPE_MESSAGE, errorPrefix, "string", key));
+    }
+    return (String) address;
+  }
```
lqiu96, line 232: If we can extract the above out, can we just revert back to using Oauth2Utils.validateString(...) and remove this method? This implementation looks to be the same as https://github.com/googleapis/google-auth-library-java/blob/cc9826e8636c6162996d83f0f2e98762dd395344/oauth2_http/java/com/google/auth/oauth2/OAuth2Utils.java#L151-L158

rmehta19: Yes, it is the same implementation, I extracted out the additional parsing logic into the calling function and reverted back to using `Oauth2Utils.validateString(...)` in the latest commit

oauth2_http/java/com/google/auth/oauth2/SecureSessionAgent.java
rmehta19: This is not really necessary, as you mention we own the MDS autoconfig endpoint. I have gotten rid of this check in the latest commit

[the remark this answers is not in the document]
rmehta19: Yes, we can do this, I've moved it out to the calling function in the latest commit.

rmehta19:
Thanks for the review @lqiu96 !

oauth2_http/javatests/com/google/auth/oauth2/MockMetadataServerTransport.java
```diff
@@ -300,9 +300,7 @@ public LowLevelHttpResponse execute() throws IOException {
         GenericJson content = new GenericJson();
         content.setFactory(OAuth2Utils.JSON_FACTORY);
         if (requestStatusCode == 200) {
-          for (Map.Entry<String, String> entrySet : s2aContentMap.entrySet()) {
-            content.put(entrySet.getKey(), entrySet.getValue());
-          }
+          content.put("s2a", s2aContentMap);
```
lqiu96, line 303: nit: Can you use the Env Var from above?

oauth2_http/java/com/google/auth/oauth2/SecureSessionAgent.java
```diff
@@ -188,17 +190,28 @@ private SecureSessionAgentConfig getSecureSessionAgentConfigFromMDS() {
 
     String plaintextS2AAddress = "";
     String mtlsS2AAddress = "";
+    Object s2aAddressConfig = responseData.get(S2A_JSON_KEY);
```
lqiu96, line 193: nit: Can you case to Map<String, Object> here so it doesn't need to be done twice below?

lqiu96 (approved):
Added a few nits if you could resolve prior to merging

oauth2_http/java/com/google/auth/oauth2/SecureSessionAgent.java
rmehta19: Done in the latest commit

oauth2_http/javatests/com/google/auth/oauth2/MockMetadataServerTransport.java
rmehta19: Done in the latest commit.

rmehta19:
Thanks @lqiu96, I've addressed them, and I think the PR should be good to merge now.

rmehta19:
I see the SonarCloud / Build Github action test failed. From looking at the log, I don't think it's related to this PR. WDYT @lqiu96 ?

lqiu96:
Should be fixed in a separate PR. Feel free to merge

rmehta19:
Thanks @lqiu96 -- I sycned the branch with main. I don't think I have permissions to merge, would you be able to?

PR #1598, opened 2024-12-12
feat: Implement Client-Side CAB token generation. 

huangjiahua:
Implement Client-Side CAB token generation.

* Temporarily added the staging repository for CEL with protobuf 3.25
* Add a new dependency tink for encryption.
* Will wait #1583 to be merged first and resolve conflicts

---

### DIFF
--- a/cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
+++ b/cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
@@ -134,7 +134,8 @@ public class ClientSideCredentialAccessBoundaryFactory {
    * @return The Client-Side CAB token in an {@link AccessToken} object
    * @throws IOException
    */
-  public AccessToken generateToken(CredentialAccessBoundary accessBoundary) throws IOException {
+  public AccessToken generateToken(CredentialAccessBoundary accessBoundary)
+      throws IOException, CelValidationException, GeneralSecurityException {
     this.refreshCredentialsIfRequired();
 
     String intermediateToken, sessionKey;
@@ -471,7 +472,8 @@ public class ClientSideCredentialAccessBoundaryFactory {
    * Serializes a {@link CredentialAccessBoundary} object into Protobuf wire format.
    */
   private byte[] serializeCredentialAccessBoundary(
-      CredentialAccessBoundary credentialAccessBoundary) throws IOException {
+      CredentialAccessBoundary credentialAccessBoundary)
+      throws CelValidationException {
     List<AccessBoundaryRule> rules =
         credentialAccessBoundary.getAccessBoundaryRules();
     ClientSideAccessBoundary.Builder accessBoundaryBuilder =
@@ -483,6 +485,7 @@ public class ClientSideCredentialAccessBoundaryFactory {
               .addAllAvailablePermissions(rule.getAvailablePermissions())
               .setAvailableResource(rule.getAvailableResource());
 
+      // Availability condition is an optional field from the CredentialAccessBoundary
       if (rule.getAvailabilityCondition() != null) {
         String availabilityCondition =
             rule.getAvailabilityCondition().getExpression();
@@ -498,37 +501,28 @@ public class ClientSideCredentialAccessBoundaryFactory {
   /**
    * Compiles CEL expression from String to an {@link Expr} proto object. 
    */
-  private Expr compileCel(String expr) throws IOException {
-    try {
-      CelAbstractSyntaxTree ast = celCompiler.parse(expr).getAst();
+  private Expr compileCel(String expr) throws CelValidationException {
+    CelAbstractSyntaxTree ast = celCompiler.parse(expr).getAst();
 
-      CelProtoAbstractSyntaxTree astProto =
-          CelProtoAbstractSyntaxTree.fromCelAst(ast);
+    CelProtoAbstractSyntaxTree astProto =
+        CelProtoAbstractSyntaxTree.fromCelAst(ast);
 
-      return astProto.getExpr();
-    } catch (CelValidationException exception) {
-      throw new IOException("Failed to parse CEL expression: " +
-                            exception.getMessage());
-    }
+    return astProto.getExpr();
   }
 
   /**
    * Encrypts the given bytes using a sessionKey using Tink Aead.
    */
-  private byte[] encryptRestrictions(byte[] restriction, String sessionKey) throws InternalError {
-    try {
-      byte[] rawKey = Base64.getDecoder().decode(sessionKey);
+  private byte[] encryptRestrictions(byte[] restriction, String sessionKey) throws GeneralSecurityException {
+    byte[] rawKey = Base64.getDecoder().decode(sessionKey);
 
-      KeysetHandle keysetHandle = TinkProtoKeysetFormat.parseKeyset(
-          rawKey, InsecureSecretKeyAccess.get());
+    KeysetHandle keysetHandle = TinkProtoKeysetFormat.parseKeyset(
+        rawKey, InsecureSecretKeyAccess.get());
 
-      Aead aead =
-          keysetHandle.getPrimitive(RegistryConfiguration.get(), Aead.class);
+    Aead aead =
+        keysetHandle.getPrimitive(RegistryConfiguration.get(), Aead.class);
 
-      return aead.encrypt(restriction, /*associatedData=*/new byte[0]);
-    } catch (GeneralSecurityException exception) {
-      throw new InternalError("Failed to parse keyset: " + exception.getMessage());
-    }
+    return aead.encrypt(restriction, /*associatedData=*/new byte[0]);
   }
 
   public static Builder newBuilder() {
--- a/cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
+++ b/cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
@@ -109,7 +109,7 @@ public class ClientSideCredentialAccessBoundaryFactory {
     try {
       AeadConfig.register();
     } catch (GeneralSecurityException e) {
-      throw new IllegalStateException("Error occurred when registering Tink");
+      throw new IllegalStateException("Error occurred when registering Tink", e);
     }
 
     CelOptions options = CelOptions.current().build();
--- a/cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
+++ b/cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
@@ -37,9 +37,12 @@ import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.google.api.client.util.Clock;
 import com.google.auth.Credentials;
+import com.google.auth.credentialaccessboundary.protobuf.ClientSideAccessBoundaryProto.ClientSideAccessBoundary;
+import com.google.auth.credentialaccessboundary.protobuf.ClientSideAccessBoundaryProto.ClientSideAccessBoundaryRule;
 import com.google.auth.http.HttpTransportFactory;
 import com.google.auth.oauth2.AccessToken;
 import com.google.auth.oauth2.CredentialAccessBoundary;
+import com.google.auth.oauth2.CredentialAccessBoundary.AccessBoundaryRule;
 import com.google.auth.oauth2.GoogleCredentials;
 import com.google.auth.oauth2.OAuth2Utils;
 import com.google.auth.oauth2.StsRequestHandler;
@@ -53,12 +56,28 @@ import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.ListenableFutureTask;
 import com.google.common.util.concurrent.MoreExecutors;
+import com.google.crypto.tink.Aead;
+import com.google.crypto.tink.InsecureSecretKeyAccess;
+import com.google.crypto.tink.KeysetHandle;
+import com.google.crypto.tink.RegistryConfiguration;
+import com.google.crypto.tink.TinkProtoKeysetFormat;
+import com.google.crypto.tink.aead.AeadConfig;
 import com.google.errorprone.annotations.CanIgnoreReturnValue;
+import dev.cel.common.CelAbstractSyntaxTree;
+import dev.cel.common.CelOptions;
+import dev.cel.common.CelProtoAbstractSyntaxTree;
+import dev.cel.common.CelValidationException;
+import dev.cel.compiler.CelCompiler;
+import dev.cel.compiler.CelCompilerFactory;
+import dev.cel.expr.Expr;
 import java.io.IOException;
 import java.time.Duration;
 import java.util.Date;
 import java.util.concurrent.ExecutionException;
 import javax.annotation.Nullable;
+import java.util.Base64;
+import java.util.List;
+import java.security.GeneralSecurityException;
 
 public class ClientSideCredentialAccessBoundaryFactory {
   static final Duration DEFAULT_REFRESH_MARGIN = Duration.ofMinutes(30);
@@ -72,6 +91,7 @@ public class ClientSideCredentialAccessBoundaryFactory {
   private final Object refreshLock = new byte[0];
   private volatile IntermediateCredentials intermediateCredentials = null;
   private final Clock clock;
+  private final CelCompiler celCompiler;
 
   enum RefreshType {
     NONE,
@@ -83,6 +103,21 @@ public class ClientSideCredentialAccessBoundaryFactory {
     this.transportFactory = builder.transportFactory;
     this.sourceCredential = builder.sourceCredential;
     this.tokenExchangeEndpoint = builder.tokenExchangeEndpoint;
+
+    // Initializes the Tink AEAD registry for encrypting the client-side
+    // restrictions.
+    try {
+      AeadConfig.register();
+    } catch (GeneralSecurityException e) {
+      throw new IllegalStateException("Error occurred when registering Tink");
+    }
+
+    CelOptions options = CelOptions.current().build();
+    this.celCompiler = CelCompilerFactory
+      .standardCelCompilerBuilder()
+      .setOptions(options)
+      .build();
+
     this.refreshMargin =
         builder.refreshMargin != null ? builder.refreshMargin : DEFAULT_REFRESH_MARGIN;
     this.minimumTokenLifetime =
@@ -92,10 +127,39 @@ public class ClientSideCredentialAccessBoundaryFactory {
     this.clock = builder.clock;
   }
 
-  public AccessToken generateToken(CredentialAccessBoundary accessBoundary) {
-    // TODO(negarb/jiahuah): Implement generateToken
-    // Note: This method will call refreshCredentialsIfRequired().
-    throw new UnsupportedOperationException("generateToken is not yet implemented.");
+  /**
+   * Generates a Client-Side CAB token given the {@link CredentialAccessBoundary}.
+   * 
+   * @param accessBoundary
+   * @return The Client-Side CAB token in an {@link AccessToken} object
+   * @throws IOException
+   */
+  public AccessToken generateToken(CredentialAccessBoundary accessBoundary) throws IOException {
+    this.refreshCredentialsIfRequired();
+
+    String intermediateToken, sessionKey;
+    Date intermediateTokenExpirationTime;
+
+    synchronized (refreshLock) {
+      intermediateToken =
+          this.intermediateCredentials.intermediateAccessToken.getTokenValue();
+      intermediateTokenExpirationTime =
+          this.intermediateCredentials.intermediateAccessToken
+              .getExpirationTime();
+      sessionKey = this.intermediateCredentials.accessBoundarySessionKey;
+    }
+
+    byte[] rawRestrictions =
+        this.serializeCredentialAccessBoundary(accessBoundary);
+
+    byte[] encryptedRestrictions =
+        this.encryptRestrictions(rawRestrictions, sessionKey);
+
+    String tokenValue =
+        intermediateToken + "." +
+        Base64.getUrlEncoder().encodeToString(encryptedRestrictions);
+
+    return new AccessToken(tokenValue, intermediateTokenExpirationTime);
   }
 
   /**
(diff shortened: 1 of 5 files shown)


cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -117,14 +150,91 @@ private void refreshCredentials() throws IOException {
...
-  public AccessToken generateToken(CredentialAccessBoundary accessBoundary) {
-    // TODO(negarb/jiahuah): Implement generateToken
-    throw new UnsupportedOperationException("generateToken is not yet implemented.");
+  public AccessToken generateToken(CredentialAccessBoundary accessBoundary) throws IOException {
+    this.refreshCredentialsIfRequired();
+
+    String intermediaryToken, sessionKey;
+    Date intermediaryTokenExpirationTime;
+
+    synchronized (this) {
```
nbayati, line 164: Why do we need a synchronized block here?

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -83,6 +103,19 @@ private ClientSideCredentialAccessBoundaryFactory(Builder builder) {
     this.transportFactory = builder.transportFactory;
     this.sourceCredential = builder.sourceCredential;
     this.tokenExchangeEndpoint = builder.tokenExchangeEndpoint;
+
+    try {
```
aeitzman, line 107: nit: can you add a comment explaining what this is doing?

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -92,10 +125,39 @@ private ClientSideCredentialAccessBoundaryFactory(Builder builder) {
...
+   * Generates a Client-Side CAB token given the {@link CredentialAccessBoundary}.
+   * 
+   * @param accessBoundary
+   * @return The Client-Side CAB token in an {@link AccessToken} object
+   * @throws IOException
+   */
+  public AccessToken generateToken(CredentialAccessBoundary accessBoundary) throws IOException {
+    this.refreshCredentialsIfRequired();
+
+    String intermediaryToken, sessionKey;
```
aeitzman, line 138: can you call this intermediateToken instead of intermediary to be consistent with Negar's code?

huangjiahua: I think it should be "intermediary" since that's what the API field is called. But I will rename it to "intermediate" in this PR and do a batch update later to make this PR clean.

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
huangjiahua: Done.

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
huangjiahua: Done.

aeitzman (approved):

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -83,6 +103,21 @@ private ClientSideCredentialAccessBoundaryFactory(Builder builder) {
     this.transportFactory = builder.transportFactory;
     this.sourceCredential = builder.sourceCredential;
     this.tokenExchangeEndpoint = builder.tokenExchangeEndpoint;
+
+    // Initializes the Tink AEAD registry for encrypting the client-side
+    // restrictions.
+    try {
+      AeadConfig.register();
+    } catch (GeneralSecurityException e) {
+      throw new IllegalStateException("Error occurred when registering Tink");
```
lqiu96, line 112: Can you throw with the exception as well `IllegalStateException(String, Exception)`?

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -92,10 +127,39 @@ private ClientSideCredentialAccessBoundaryFactory(Builder builder) {
...
+    Date intermediateTokenExpirationTime;
+
+    synchronized (refreshLock) {
+      intermediateToken =
+          this.intermediateCredentials.intermediateAccessToken.getTokenValue();
+      intermediateTokenExpirationTime =
+          this.intermediateCredentials.intermediateAccessToken
+              .getExpirationTime();
+      sessionKey = this.intermediateCredentials.accessBoundarySessionKey;
+    }
```
lqiu96, line 150: qq, why is this in a synchronized block?

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -403,6 +467,70 @@ public void run() {
...
+              .addAllAvailablePermissions(rule.getAvailablePermissions())
+              .setAvailableResource(rule.getAvailableResource());
+
+      if (rule.getAvailabilityCondition() != null) {
+        String availabilityCondition =
+            rule.getAvailabilityCondition().getExpression();
+
+        Expr availabilityConditionExpr = this.compileCel(availabilityCondition);
+        ruleBuilder.setCompiledAvailabilityCondition(availabilityConditionExpr);
+      }
```
lqiu96, line 492: qq, I'm not sure I understand this part. It's creating a ClientSideAccessBoundaryRule.Builder object and adding values to it, but doesn't end up using it again?

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
huangjiahua: Write access to `intermediateCredentials` is in a synchronized block, since the refresh can happen asynchronously: https://github.com/googleapis/google-auth-library-java/blob/0d96dcf508973af23f52602c18723b205b7de03b/cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java#L298-L307

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
huangjiahua: The `ClientSideAccessBoundaryRule.Builder` is created using `accessBoundaryBuilder.addAccessBoundaryRulesBuilder()` so the builder gets created inside the parent builder `accessBoundaryBuilder`, and the whole rule gets built when `accessBoundaryBuilder.build()` is executed. See: https://protobuf.dev/reference/java/java-generated/#sub-builders

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
huangjiahua: Done

nbayati (approved):
LGTM

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
lqiu96: Ah yes. Thanks.

Correct me if I'm wrong, I believe that `refreshCredentialsIfRequired` may trigger a sync refresh and take longer until we get those values, right? And for cases with async refresh (nearing expiration), then it will may potentially run with older, but still valid intermediateAccessToken value (as async may not have been completed yet)?

If the above is correct, perhaps we can call note those behaviors in the javadocs?

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
lqiu96: Gotcha, that's neat. Thanks!

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -403,6 +467,70 @@ public void run() {
...
+    try {
+      CelAbstractSyntaxTree ast = celCompiler.parse(expr).getAst();
+
+      CelProtoAbstractSyntaxTree astProto =
+          CelProtoAbstractSyntaxTree.fromCelAst(ast);
+
+      return astProto.getExpr();
+    } catch (CelValidationException exception) {
+      throw new IOException("Failed to parse CEL expression: " +
+                            exception.getMessage());
```
lqiu96, line 511: Why re-throw as IOException since it doesn't seem like an I/O occurs here? Are we trying to hide the underlying CEL logic from the users?

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -403,6 +467,70 @@ public void run() {
...
+    ClientSideAccessBoundary.Builder accessBoundaryBuilder =
+        ClientSideAccessBoundary.newBuilder();
+
+    for (AccessBoundaryRule rule : rules) {
+      ClientSideAccessBoundaryRule.Builder ruleBuilder =
+          accessBoundaryBuilder.addAccessBoundaryRulesBuilder()
+              .addAllAvailablePermissions(rule.getAvailablePermissions())
+              .setAvailableResource(rule.getAvailableResource());
+
+      if (rule.getAvailabilityCondition() != null) {
```
lqiu96, line 486: Are there specific conditions that make getAvailabilityCondition nullable? If so, can that be documented?

lqiu96: Also, if possible, can you document why availabilityCondition is compiled with CEL and the behavior availability doesn't exist/ isn't compiled?

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -403,6 +467,70 @@ public void run() {
...
+    try {
+      byte[] rawKey = Base64.getDecoder().decode(sessionKey);
+
+      KeysetHandle keysetHandle = TinkProtoKeysetFormat.parseKeyset(
+          rawKey, InsecureSecretKeyAccess.get());
+
+      Aead aead =
+          keysetHandle.getPrimitive(RegistryConfiguration.get(), Aead.class);
+
+      return aead.encrypt(restriction, /*associatedData=*/new byte[0]);
```
lqiu96, line 528: nit: Can you note the significance of `byte[0]`?

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -403,6 +467,70 @@ public void run() {
...
+    } catch (GeneralSecurityException exception) {
+      throw new InternalError("Failed to parse keyset: " + exception.getMessage());
```
lqiu96, line 530: qq, what's the benefit of re-throwing this as InternalError instead of keeping it as GeneralSecurityException?

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
huangjiahua: Do you mean documenting why we need a `byte[0]` here?

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
huangjiahua: Not really. Changed it to throw CelValidationException directly

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
huangjiahua: Done. Availability condition is an optional field from users

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
huangjiahua: Changed it to throw GeneralSecurityException

cab-token-generator/javatests/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactoryTest.java
```diff
@@ -572,4 +586,173 @@ private static void triggerConcurrentRefresh(
...
+    assertEquals(expr.getCallExpr()
+                     .getTarget()
+                     .getSelectExpr()
+                     .getOperand()
+                     .getIdentExpr()
+                     .getName(),
+                 "resource");
+    assertEquals(expr.getCallExpr().getFunction(), "startsWith");
+    assertEquals(expr.getCallExpr().getArgs(0).getConstExpr().getStringValue(),
+                 "projects/_/buckets/example-bucket/objects/customer-a");
```
lqiu96, line 661: If possible, can we break testing two main logic components (serializeCredentialAccessBoundary and encryptRestrictions) into smaller unit tests? I think it may be easier to read, but let me know if you have any concerns with this.

i.e. Perhaps something like
1. serializeCredentialAccessBoundary_withAvailabilityRestrictions_success
2. serializeCredentialAccessBoundary_withoutAvailabilityRestrictions_success
3. serializeCredentialAccessBoundary_failure 

Then also for encryptRestrictions
1. encryptRestrictions_success
2. encryptRestrictions_failure

huangjiahua: I would prefer not exposing private methods but testing the public `generateToken` as a whole, but I see currently we don't have a test to cover encryption failures, so I'll add one with invalid keys from STS. I'll also rename the test cases.

huangjiahua: Done.

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
huangjiahua: Sure. Should I add it to the `generateToken` method or the parameters that control this behavior (`refreshMargin`).

cab-token-generator/javatests/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactoryTest.java
lqiu96: You can make the methods package-private. It will still be accessible via the tests and hidden from the user. We can also annotate them with `@VisibleForTesting`

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
lqiu96: I think to the javadocs here would make sense. Since this is the public method and the logic would be hidden for the callers, they may not know what is going on under the hood.

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
lqiu96: On second thought, do you know what the error message is from the `GeneralSecurityException` when it fails? If it's too vague (i'm assuming it may be because it says General), perhaps we can re-throw with the error message you wrote above. If we do re-throw, I'm not too certain about InternalError. I think a different exception that would highlight the error type would be best.

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
huangjiahua: I already added the comment.

cab-token-generator/javatests/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactoryTest.java
huangjiahua: I considered using `@VisibleForTesting` here, but I'm hesitant to introduce it as I believe it can sometimes be a shortcut for testing things we should ideally test through public APIs.  We can fully test the behavior we need using the generateToken method, even though the assertions are a bit longer.  I don't think the added verbosity warrants using `@VisibleForTesting` in this case.  Let me know if you see other issues or have a different perspective.

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
huangjiahua: The message is "Parse keyset failed". I think it's specific.

cab-token-generator/javatests/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactoryTest.java
huangjiahua: I understand your concerns. I'd like to address them by introducing helper functions in the test code to improve readability and adding comments to clarify the purpose of each test.

I'm still hesitant to add tests for private methods such as `compileCel` or `encryptRestrictions` because I believe it goes against the principle that:

> If you change your software's internal implementation, your tests should not break as long as the change is not observable by users. Therefore, per the black-box testing principle, most of the time you should test your code through its public interfaces.

Testing through the public API allows us to validate the system's behavior as a user would experience it. For instance, we can ensure that an invalid CEL expression results in a CelValidationException for the user. Even if we tested private method failures directly, those tests wouldn't necessarily cover the public API, and we couldn't be certain the same errors would be exposed to the user.

If you think failure test cases for cel compilation and encryption are necessary. I can add tests for those methods but I would still keep the failure cases for the public method.

huangjiahua: Refactored the tests as I mentioned above.

lqiu96: I'll leave this up to you since I believe test unit tests do a thorough enough test and I'm confident about the changes. I don't want to make this a blocker for this PR.

I don't think these unit tests don't seem to follow the ideas above besides just calling from the public method. If you're using the internal logic/ code (the black box) to validate the output, I'm not sure why it needs to be a done via black box testing to begin with. To me, it's opening the black box and making the validation reliant what's in the black box, which seems to ruin the whole idea of having the black box. For example If Cel now requires you to use `byte[100]` instead of `byte[0]` to encrypt and this also returns a different, valid access token, these unit tests would probably fail since it's decrypting with `byte[0]`.

Logic within `refreshCredentialsIfRequired()` also is only run, but that isn't tested within this black-box testing principle. Instead we have dedicated unit tests for `refreshCredentialsIfRequired`. Had this been done via the blackbox testing principle, I would think we'd see a whole matrix of new test cases pop up:
1. generateToken_blocking_availabilityConditionsExist_success()
2. generateToken_blocking_availabilityConditionsExist_celExceptionError()
etc

Perhaps I don't understand how these tests follow black-box testing and I'll leave it as that.

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -92,10 +127,42 @@ private ClientSideCredentialAccessBoundaryFactory(Builder builder) {
...
-  public AccessToken generateToken(CredentialAccessBoundary accessBoundary) {
-    // TODO(negarb/jiahuah): Implement generateToken
-    // Note: This method will call refreshCredentialsIfRequired().
-    throw new UnsupportedOperationException("generateToken is not yet implemented.");
+  /**
+   * Generates a Client-Side CAB token given the {@link CredentialAccessBoundary}.
+   * 
+   * @param accessBoundary
+   * @return The Client-Side CAB token in an {@link AccessToken} object
+   * @throws IOException If an I/O error occurs while refrehsing the source credentials
```
lqiu96, line 135:
```suggestion
   * @throws IOException If an I/O error occurs while refreshing the source credentials
```

cab-token-generator/javatests/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactoryTest.java
```diff
@@ -572,4 +589,334 @@ private static void triggerConcurrentRefresh(
...
+                            .AvailabilityCondition.newBuilder()
+                            .setExpression(
+                                "resource.name.startsWith('projects/_/"
+                                + "buckets/example-bucket/objects/customer-a')")
+                            .build())
+                    .build())
+            .build();
+
+    assertThrows(IllegalArgumentException.class,
+                 () -> { factory.generateToken(accessBoundary); });
```
lqiu96, line 880: for this one, can you document where IllegalArgumentException may stem from. This is a runtimeexception and isn't marked as thrown by `generateToken()` so it may be confusing for future devs to figure out where it's coming from.

lqiu96 (approved):

cab-token-generator/javatests/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactoryTest.java
huangjiahua: This comes from the base64 decoding of the session key. Since the session key is from the server, I changed it to catch it and re-throw it as an IllegalStateExeption.

huangjiahua:
@lqiu96 Looks like CLA check fails because it can't associate your email to Google

PR #1607, opened 2025-01-08
docs: Update README with client-side CAB instructions

nbayati:
This commit updates the README file to include instructions for setting up and using the client-side CAB feature.

Design doc: [go/client-side-cab-client-library-java](http://goto.google.com/client-side-cab-client-library-java)

### DIFF
--- a/README.md
+++ b/README.md
@@ -995,14 +995,19 @@ resources.
 There are two ways to generate downscoped tokens using a
 CredentialAccessBoundary:
 
-* Server-side: Uses the `DownscopedCredentials` class. Each time a
-  downscoped token is needed, the client makes a call to the Security Token Service (STS).
-  This is suitable for applications that require downscoped tokens infrequently, or applications that reuse a single downscoped credential many times.
-* Client-side: Uses the `ClientSideCredentialAccessBoundaryFactory` class. This
-  approach minimizes calls to STS. The client retrieves necessary cryptographic
-  material once and then generates multiple downscoped tokens locally. This is
-  more efficient for applications that need to generate many unique downscoped tokens.
+* **Server-side (using `DownscopedCredentials`):** The client calls the Security
+Token Service (STS) each time a downscoped token is needed. This is suitable for
+applications where the Credential Access Boundary rules change infrequently or 
+when a single downscoped credential is reused many times.  A key consideration 
+is that every rule change requires a new call to the STS.
 
+
+* **Client-side (using `ClientSideCredentialAccessBoundaryFactory`):** The client 
+retrieves cryptographic material once and then generates multiple downscoped 
+tokens locally. This minimizes calls to the STS and is more efficient when 
+Credential Access Boundary rules change frequently, as the client doesn't need 
+to contact the STS for each rule change.  This is also more efficient for 
+applications that need to generate many *unique* downscoped tokens.
 #### Server-side CAB
 
 The `DownscopedCredentials` class can be used to produce a downscoped access
--- a/README.md
+++ b/README.md
@@ -998,7 +998,7 @@ CredentialAccessBoundary:
 * **Server-side (using `DownscopedCredentials`):** The client calls the Security
 Token Service (STS) each time a downscoped token is needed. This is suitable for
 applications where the Credential Access Boundary rules change infrequently or 
-when a single downscoped credential is reused many times.  A key consideration 
+when a single downscoped credential is reused many times. A key consideration 
 is that every rule change requires a new call to the STS.
 
 
@@ -1006,8 +1006,9 @@ is that every rule change requires a new call to the STS.
 retrieves cryptographic material once and then generates multiple downscoped 
 tokens locally. This minimizes calls to the STS and is more efficient when 
 Credential Access Boundary rules change frequently, as the client doesn't need 
-to contact the STS for each rule change.  This is also more efficient for 
+to contact the STS for each rule change. This is also more efficient for 
 applications that need to generate many *unique* downscoped tokens.
+
 #### Server-side CAB
 
 The `DownscopedCredentials` class can be used to produce a downscoped access
--- a/README.md
+++ b/README.md
@@ -953,7 +953,7 @@ enables restricting the Identity and Access Management (IAM) permissions that a
 short-lived credential can use for Cloud Storage. This involves creating a
 `CredentialAccessBoundary` that defines the restrictions applied to the
 downscoped token. Using downscoped credentials ensures tokens in flight always
-have the least privileges (Principle of Least Privilege).
+have the least privileges ([Principle of Least Privilege](https://en.wikipedia.org/wiki/Principle_of_least_privilege)).
 
 #### Creating a CredentialAccessBoundary
 
--- a/README.md
+++ b/README.md
@@ -997,28 +997,41 @@ CredentialAccessBoundary:
 
 * Server-side: Uses the `DownscopedCredentials` class. Each time a
   downscoped token is needed, the client makes a call to the Security Token Service (STS).
-  This is suitable for applications that require downscoped tokens infrequently.
+  This is suitable for applications that require downscoped tokens infrequently, or applications that reuse a single downscoped credential many times.
 * Client-side: Uses the `ClientSideCredentialAccessBoundaryFactory` class. This
   approach minimizes calls to STS. The client retrieves necessary cryptographic
   material once and then generates multiple downscoped tokens locally. This is
-  more efficient for applications that need to generate many downscoped tokens.
+  more efficient for applications that need to generate many unique downscoped tokens.
 
 #### Server-side CAB
 
 The `DownscopedCredentials` class can be used to produce a downscoped access
-token from a source credential and the `CredentialAccessBoundary` created above
-in the Token Broker:
+token from a source credential and the `CredentialAccessBoundary`.
 
 ```java
 // Retrieve the source credentials from ADC.
 GoogleCredentials sourceCredentials = GoogleCredentials.getApplicationDefault()
         .createScoped("https://www.googleapis.com/auth/cloud-platform");
 
+// Create an Access Boundary Rule which will restrict the downscoped token to having readonly
+// access to objects starting with "customer-a" in bucket "bucket-123".
+String availableResource = "//storage.googleapis.com/projects/_/buckets/bucket-123";
+String availablePermission = "inRole:roles/storage.objectViewer";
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -980,12 +983,32 @@ CredentialAccessBoundary credentialAccessBoundary =
...
 
-Using the CredentialAccessBoundary created above in the Token Broker:
+#### Generating Downscoped Tokens
+
+There are two ways to generate downscoped tokens using a
+CredentialAccessBoundary:
+
+* Server-side: Uses the `DownscopedCredentials` class. Each time a
+  downscoped token is needed, the client makes a call to the Security Token Service (STS).
+  This is suitable for applications that require downscoped tokens infrequently.
```
aeitzman, line 1000: Nit: Server side CAB would also be suitable for customers that use a single credential frequently, its more about the number of uniquely downscoped tokens used rather than the frequency of use.
```suggestion
  This is suitable for applications that require downscoped tokens infrequently, or applications that reuse a single downscoped credential many times.
```

README.md
```diff
@@ -980,12 +983,32 @@ CredentialAccessBoundary credentialAccessBoundary =
...
+* Client-side: Uses the `ClientSideCredentialAccessBoundaryFactory` class. This
+  approach minimizes calls to STS. The client retrieves necessary cryptographic
+  material once and then generates multiple downscoped tokens locally. This is
+  more efficient for applications that need to generate many downscoped tokens.
```
aeitzman, line 1004:
```suggestion
  more efficient for applications that need to generate many unique downscoped tokens.
```

README.md
```diff
@@ -1003,45 +1026,40 @@ DownscopedCredentials downscopedCredentials =
...
+objects to create multiple downscoped tokens.
 
-Putting it all together:
 ```java
 // Retrieve the source credentials from ADC.
 GoogleCredentials sourceCredentials = GoogleCredentials.getApplicationDefault()
         .createScoped("https://www.googleapis.com/auth/cloud-platform");
 
-// Create an Access Boundary Rule which will restrict the downscoped token to having readonly
-// access to objects starting with "customer-a" in bucket "bucket-123".
```
aeitzman, line 1020: Can you just stick the portion of this code sample that creates the accessboundary rule into the samples for both the client side and server side sections? I think its good to have a single block of code that includes the whole flow.

README.md
```diff
@@ -949,16 +949,19 @@ googleapis.com domain.
 ### Downscoping with Credential Access Boundaries
 
 [Downscoping with Credential Access Boundaries](https://cloud.google.com/iam/docs/downscoping-short-lived-credentials)
-enables the ability to downscope, or restrict, the Identity and Access Management (IAM) permissions
-that a short-lived credential can use for Cloud Storage.
+enables restricting the Identity and Access Management (IAM) permissions that a
+short-lived credential can use for Cloud Storage. This involves creating a
```
lqiu96, line 953: Is this intended only for Cloud Storage?

README.md
```diff
@@ -980,37 +983,69 @@ CredentialAccessBoundary credentialAccessBoundary =
...
+There are two ways to generate downscoped tokens using a
+CredentialAccessBoundary:
+
+* Server-side: Uses the `DownscopedCredentials` class. Each time a
+  downscoped token is needed, the client makes a call to the Security Token Service (STS).
+  This is suitable for applications that require downscoped tokens infrequently, or applications that reuse a single downscoped credential many times.
+* Client-side: Uses the `ClientSideCredentialAccessBoundaryFactory` class. This
+  approach minimizes calls to STS. The client retrieves necessary cryptographic
+  material once and then generates multiple downscoped tokens locally. This is
+  more efficient for applications that need to generate many unique downscoped tokens.
```
lqiu96, line 1004: Perhaps here or a different section, could highlight when a user may want to use one or the other? 

Any pros/cons between the two? Is one preferred? What general scenarios or guidelines would call for one vs the other?

README.md
nbayati: From @huangjiahua: Yes, at GA this time Client-Side CAB is for GCS only. We have a verification on server side that only allows GCS resource for the available resources. In the future, AION plans to expand Client-Side CAB to other services.

README.md
nbayati: Updated. There are two factors that are mentioned: one is if the user needs many unique tokens or they can re-use a token, and the other is the CAB rule changes frequently. Depending on their use case, the users can choose one or the other.

README.md
lqiu96: nit: Can you specify that this is something that is currently only supported for GCS and may/ will be expanded to other services in the future?

.i.e.
```
short-lived credential (currently only for Cloud Storage and may be expanded to additional services in the future). This involves creating a
```
or something else

README.md
```diff
@@ -980,37 +983,74 @@ CredentialAccessBoundary credentialAccessBoundary =
...
+#### Generating Downscoped Tokens
+
+There are two ways to generate downscoped tokens using a
+CredentialAccessBoundary:
+
+* **Server-side (using `DownscopedCredentials`):** The client calls the Security
+Token Service (STS) each time a downscoped token is needed. This is suitable for
+applications where the Credential Access Boundary rules change infrequently or 
+when a single downscoped credential is reused many times.  A key consideration 
+is that every rule change requires a new call to the STS.
```
lqiu96, line 1002:
```suggestion
when a single downscoped credential is reused many times. A key consideration 
is that every rule change requires a new call to the STS.
```

README.md
```diff
@@ -980,37 +983,74 @@ CredentialAccessBoundary credentialAccessBoundary =
...
+
+
+* **Client-side (using `ClientSideCredentialAccessBoundaryFactory`):** The client 
+retrieves cryptographic material once and then generates multiple downscoped 
+tokens locally. This minimizes calls to the STS and is more efficient when 
+Credential Access Boundary rules change frequently, as the client doesn't need 
+to contact the STS for each rule change.  This is also more efficient for 
+applications that need to generate many *unique* downscoped tokens.
```
lqiu96, line 1010:
```suggestion
to contact the STS for each rule change. This is also more efficient for 
applications that need to generate many *unique* downscoped tokens.
```

README.md
```diff
@@ -980,37 +983,74 @@ CredentialAccessBoundary credentialAccessBoundary =
...
+#### Server-side CAB
```
lqiu96, line 1011: nit: Add a line break
```suggestion

#### Server-side CAB
```

README.md
```diff
@@ -980,37 +983,74 @@ CredentialAccessBoundary credentialAccessBoundary =
...
 The common pattern of usage is to have a token broker with elevated access generate these downscoped
 credentials from higher access source credentials and pass the downscoped short-lived access tokens
 to a token consumer via some secure authenticated channel for limited access to Google Cloud Storage
 resources.
 
-Using the CredentialAccessBoundary created above in the Token Broker:
+#### Generating Downscoped Tokens
+
+There are two ways to generate downscoped tokens using a
+CredentialAccessBoundary:
```
lqiu96, line 996: nit: If possible, could there be like a table or a pro/ cons comparison between the two that explicitly spells out what any considerations/ impacts of one vs the other?

Also, is there a need for something like a migration guide? To help users potentially migrate from server-side to client-side?

lqiu96: Also, is there an general recommendation that that your team would provide between the two. I know it's possible that may not be a recommendation.

i.e. Prefer client-side unless ... X,Y,Z blocker?

lqiu96 (approved):
LGTM. Added a few nits if you could address.

lqiu96:
I also just noticed this here: https://github.com/googleapis/google-auth-library-java/blob/8e59c5940ed75c2fbdf419839960199cb63f0fe5/README.md?plain=1#L10

I think we can update this to reflect this latest module.

README.md
nbayati: Since we're already mentioning that it is used for Cloud Storage, and we don't have a clear timeline of when other services may be supported, I think it's better to keep the language as is.

README.md
nbayati: Checked with the team, we don't have a general recommendation, or a migration guide. Depending on their use case, and whether or not they need many unique downscoped tokens or they can re-use existing ones, they can decide between the two optiosn.

nbayati:
> I also just noticed this here:
> 
> https://github.com/googleapis/google-auth-library-java/blob/8e59c5940ed75c2fbdf419839960199cb63f0fe5/README.md?plain=1#L10
> 
> I think we can update this to reflect this latest module.

Ah good catch! I just realized that all the cab stuff are under `google-auth-library-oauth2-http` section. I'll probably need to add another section for this new module and move some a few things around to make sure it's organized correctly.

README.md
lqiu96: How does your team plan to update the docs when potentially new services are added (say a year or two from now)? I'm just worried that we'll forget to update this when services are added. The wording now makes it seem like this is only supported for GCS. I'd rather it be it sound like GCS (for now) and potentially other services in the future.

README.md
```diff
@@ -980,37 +983,75 @@ CredentialAccessBoundary credentialAccessBoundary =
...
-(token consumers) in the same network will send authenticated requests to that broker for downscoped
-tokens to access or modify specific google cloud storage buckets.
+#### Client-side CAB
 
-The broker will instantiate downscoped credentials instances that can be used to generate short 
-lived downscoped access tokens which will be passed to the token consumer. 
+For client-side CAB, the `ClientSideCredentialAccessBoundaryFactory` is used
+with a source credential. After initializing the factory, the `generateToken()`
+method can be called repeatedly with different `CredentialAccessBoundary`
+objects to create multiple downscoped tokens.
```
lqiu96, line 1053: Before this paragraph, can we add a note about this being in a different package? `cab-token-generator`?

README.md
nbayati: I'll create a doc to track all the places that need to be updated when the time comes. It includes auth libraries, official docs, and code comments and samples. In the current version of the read me file, we have multiple places that emphasizes on this feature only being available for GCS, for example here:

https://github.com/googleapis/google-auth-library-java/blob/8877624ccde62d999b25ab468bd1d474e36dcd45/README.md?plain=1#L1168-L1169

lqiu96: Ok gotcha. Sounds good to me then!

lqiu96 (approved):

lqiu96:
Oh @nbayati I see this is to the client-side-cab branch and not to main. I think we'll need to raise another PR to main

PR #1624, opened 2025-01-22
feat: Adds an X509 certificate provider in the auth libraries.

aeitzman:
This PR creates a new "X509" certificate provider in a new Mtls package to handle workload X509 certificate reading. This is part of a larger feature described in go/x509-workload-auth-library-design

Merging this into a feature branch for now until the entire feature is ready.


oauth2_http/java/com/google/auth/mtls/WorkloadCertificateConfiguration.java
```diff
@@ -0,0 +1,46 @@
...
+
+  public static WorkloadCertificateConfiguration fromCertificateConfigurationStream(
+      InputStream certConfigStream) throws IOException {
+    JsonFactory jsonFactory = GsonFactory.getDefaultInstance();
+    JsonObjectParser parser = new JsonObjectParser(jsonFactory);
+
+    GenericJson fileContents =
+        parser.parseAndClose(certConfigStream, StandardCharsets.UTF_8, GenericJson.class);
+
+    Map<String, Object> certConfigs = (Map<String, Object>) fileContents.get("cert_configs");
```
andyrzhao, line 38: Should we do a null/empty check on every field and throw exception if expected field is not present?

oauth2_http/java/com/google/auth/mtls/X509Provider.java
```diff
@@ -0,0 +1,119 @@
...
+      // Build a key store using the combined stream.
+      return SecurityUtils.createMtlsKeyStore(certAndPrivateKeyStream);
+    } catch (Exception e) {
+      throw new IOException(e);
+    }
+  }
+
+  private WorkloadCertificateConfiguration getWorkloadCertificateConfiguration()
+      throws IOException {
+    String envCredentialsPath = getEnv(CERTIFICATE_CONFIGURATION_ENV_VARIABLE);
```
andyrzhao, line 52: nit: Since this env var is not used if explicit override is set, it's prob better to move it to an else block, ex:
else {
  String env = getEnv(...)
  if (env is set) {
  use env
   } else {
   use wellknown
  }
}

oauth2_http/java/com/google/auth/mtls/X509Provider.java
```diff
@@ -0,0 +1,119 @@
...
+    File certConfig;
+    if (this.certConfigPathOverride != null) {
+      certConfig = new File(certConfigPathOverride);
+    } else if (envCredentialsPath != null && envCredentialsPath.length() > 0) {
```
andyrzhao, line 56: && !envCredentialsPath.isEmpty()

oauth2_http/java/com/google/auth/mtls/X509Provider.java
```diff
@@ -0,0 +1,119 @@
...
+      certConfig = new File(envCredentialsPath);
+    } else {
+      certConfig = getWellKnownCertificateConfigFile();
+    }
+    InputStream certConfigStream = null;
+    try {
+      if (!isFile(certConfig)) {
```
andyrzhao, line 63: nit: fileExists(

oauth2_http/java/com/google/auth/mtls/X509Provider.java
```diff
@@ -0,0 +1,119 @@
...
+          e);
+    } finally {
+      if (certConfigStream != null) {
+        certConfigStream.close();
+      }
+    }
+  }
+
+  boolean isFile(File file) {
+    return file.isFile();
```
andyrzhao, line 86: add null check here as good practice (even if file is not expected to be null)

oauth2_http/javatests/com/google/auth/mtls/X509ProviderTest.java
```diff
@@ -0,0 +1,176 @@
...
+import java.security.cert.Certificate;
+import java.security.cert.CertificateException;
+import java.security.cert.CertificateFactory;
+import java.util.HashMap;
+import java.util.Map;
+import org.junit.Test;
+
+public class X509ProviderTest {
+
+  static String TEST_CERT =
```
andyrzhao, line 24: should "static" be "private static" here and elsewhere in this class?

oauth2_http/java/com/google/auth/mtls/X509Provider.java
```diff
@@ -0,0 +1,119 @@
...
+
+  String getOsName() {
+    return getProperty("os.name", "").toLowerCase(Locale.US);
+  }
+
+  String getProperty(String property, String def) {
+    return System.getProperty(property, def);
+  }
+
+  File getWellKnownCertificateConfigFile() {
```
andyrzhao, line 105: I think all these helper methods should be "private" whenever possible right? and use protected for those that needs to be overridden by the test mock?

andyrzhao (commented):
Thanks for working on this!

oauth2_http/java/com/google/auth/mtls/X509Provider.java
aeitzman: It looks like file.exists() will return true if the file is a directory, which we do not want here. isFile will return true only if it is a file

oauth2_http/java/com/google/auth/mtls/X509Provider.java
aeitzman: The purpose of this function is mainly just to help testing the provider by allowing us to mock the file read. I think we should just let the file.isFile() handle the null the check and throw the error instead of putting in our own error handling here.

I'll add comments to this block to make it more clear what the purpose of these functions are.

oauth2_http/java/com/google/auth/mtls/X509Provider.java
aeitzman: I'll make this function private, for the methods that need to be overridden, I think its better to keep them package private instead of protected, since we wouldn't expect packages that ingest this library to override them. For context, this is the same pattern that we use for the Default credentials provider in this library (https://github.com/googleapis/google-auth-library-java/blob/main/oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java)

oauth2_http/java/com/google/auth/mtls/WorkloadCertificateConfiguration.java
```diff
@@ -0,0 +1,64 @@
+package com.google.auth.mtls;
```
lqiu96, line 1: Missing header here

oauth2_http/java/com/google/auth/mtls/WorkloadCertificateConfiguration.java
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  public String getPrivateKeyPath() {
+    return privateKeyPath;
+  }
+
+  public static WorkloadCertificateConfiguration fromCertificateConfigurationStream(
+      InputStream certConfigStream) throws IOException {
+    JsonFactory jsonFactory = GsonFactory.getDefaultInstance();
+    JsonObjectParser parser = new JsonObjectParser(jsonFactory);
```
lqiu96, line 33: nit: Is it possible for these to be static variables so they aren't re-created on each call?

oauth2_http/java/com/google/auth/mtls/WorkloadCertificateConfiguration.java
```diff
@@ -0,0 +1,64 @@
...
+  public String getCertPath() {
+    return certPath;
+  }
+
+  public String getPrivateKeyPath() {
+    return privateKeyPath;
+  }
+
+  public static WorkloadCertificateConfiguration fromCertificateConfigurationStream(
+      InputStream certConfigStream) throws IOException {
```
lqiu96, line 31: nit: Perhaps we can check for a null `certConfigStream` argument? at the beginning?

oauth2_http/java/com/google/auth/mtls/WorkloadCertificateConfiguration.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    Map<String, Object> workloadConfig = (Map<String, Object>) certConfigs.get("workload");
+    if (workloadConfig == null) {
+      throw new IllegalArgumentException(
+          "A workload certificate configuration must be provided in the cert_configs object.");
+    }
+
+    String certPath = (String) workloadConfig.get("cert_path");
+    if (certPath.isEmpty() || certPath == null) {
```
lqiu96, line 51: nit: Guava's [Strings.isNullOrEmpty](https://guava.dev/releases/20.0/api/docs/com/google/common/base/Strings.html)

oauth2_http/java/com/google/auth/mtls/WorkloadCertificateConfiguration.java
```diff
@@ -0,0 +1,64 @@
...
+import com.google.api.client.json.GenericJson;
+import com.google.api.client.json.JsonFactory;
+import com.google.api.client.json.JsonObjectParser;
+import com.google.api.client.json.gson.GsonFactory;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+import java.util.Map;
+
+public class WorkloadCertificateConfiguration {
```
lqiu96, line 12: This class looks like it's just being used from X509Provider class. Can this be package-private scope?

oauth2_http/java/com/google/auth/mtls/X509Provider.java
```diff
@@ -0,0 +1,127 @@
...
+  static final String CLOUDSDK_CONFIG_DIRECTORY = "gcloud";
+
+  private String certConfigPathOverride;
+
+  public X509Provider(String certConfigPathOverride) {
+    this.certConfigPathOverride = certConfigPathOverride;
+  }
+
+  public X509Provider() {
+    this.certConfigPathOverride = null;
```
lqiu96, line 25:
```suggestion
    super(null);
```

oauth2_http/java/com/google/auth/mtls/X509Provider.java
```diff
@@ -0,0 +1,127 @@
...
+
+      // Merge the two streams into a single stream.
+      SequenceInputStream certAndPrivateKeyStream =
+          new SequenceInputStream(certStream, privateKeyStream);
+
+      // Build a key store using the combined stream.
+      return SecurityUtils.createMtlsKeyStore(certAndPrivateKeyStream);
+    } catch (Exception e) {
+      throw new IOException(e);
+    }
```
lqiu96, line 47: I think we would need to close the streams created above

oauth2_http/java/com/google/auth/mtls/X509Provider.java
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  String getProperty(String property, String def) {
+    return System.getProperty(property, def);
+  }
+  /*
+   * End of methods to allow overriding in the test code to isolate from the environment.
+   */
+
+  private File getWellKnownCertificateConfigFile() {
```
lqiu96, line 113: qq, is the the standard/ normal flow for determining OSes? If not, do we want to catch or guard for any sort of no file found exception in case we determined the wrong OS?

oauth2_http/java/com/google/auth/mtls/X509Provider.java
```diff
@@ -0,0 +1,127 @@
...
+
+  public X509Provider(String certConfigPathOverride) {
+    this.certConfigPathOverride = certConfigPathOverride;
+  }
+
+  public X509Provider() {
+    this.certConfigPathOverride = null;
+  }
+
+  public KeyStore getKeyStore() throws IOException {
```
lqiu96, line 28: If I'm understanding this correctly, users for X.509 (us) will be invoking this method. Can you add some javadocs above for this method? I know the use case is intended to be internal, but may be helpful for future maintainers.

oauth2_http/java/com/google/auth/mtls/X509Provider.java
```diff
@@ -0,0 +1,127 @@
...
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.SequenceInputStream;
+import java.security.KeyStore;
+import java.util.Locale;
+
+public class X509Provider {
```
lqiu96, line 13: We don't have the equivalent for `@InternalApi` here. Can you add some strongly worded comments above regarding this being internalapi and not meant for users (no backwards compatibility guarantee and the like)?

oauth2_http/java/com/google/auth/mtls/WorkloadCertificateConfiguration.java
```diff
@@ -0,0 +1,102 @@
...
+    return privateKeyPath;
+  }
+
+
+
+  static WorkloadCertificateConfiguration fromCertificateConfigurationStream(
+      InputStream certConfigStream) throws IOException {
+    if (certConfigStream == null){
+      throw new IllegalArgumentException("certConfigStream must not be null.");
+    }
```
lqiu96, line 71: nit: `checkNotNull`: https://www.javadoc.io/doc/com.google.guava/guava/33.0.0-jre/com/google/common/base/Preconditions.html

oauth2_http/java/com/google/auth/mtls/X509Provider.java
```diff
@@ -0,0 +1,179 @@
...
+    try {
+      // Read the certificate and private key file paths into separate streams.
+      File certFile = new File(workloadCertConfig.getCertPath());
+      File privateKeyFile = new File(workloadCertConfig.getPrivateKeyPath());
+      certStream = readStream(certFile);
+      privateKeyStream = readStream(privateKeyFile);
+
+      // Merge the two streams into a single stream.
+      SequenceInputStream certAndPrivateKeyStream =
+          new SequenceInputStream(certStream, privateKeyStream);
```
lqiu96, line 86: qq, does this need to be closed after `SecurityUtils.createMtlsKeyStore` in the finally block?

oauth2_http/java/com/google/auth/mtls/X509Provider.java
```diff
@@ -0,0 +1,179 @@
...
+  /*
+   * Start of methods to allow overriding in the test code to isolate from the environment.
+   */
+  boolean isFile(File file) {
+    return file.isFile();
+  }
+
+  InputStream readStream(File file) throws FileNotFoundException {
+    return new FileInputStream(file);
+  }
```
lqiu96, line 148: nit: rename to `createInputStream()` for clarity. I don't think this method is reading the stream at all.

oauth2_http/java/com/google/auth/mtls/X509Provider.java
lqiu96: bumping this comment

oauth2_http/java/com/google/auth/mtls/X509Provider.java
```diff
@@ -0,0 +1,179 @@
...
+
+  private WorkloadCertificateConfiguration getWorkloadCertificateConfiguration()
+      throws IOException {
+    File certConfig;
+    if (this.certConfigPathOverride != null) {
+      certConfig = new File(certConfigPathOverride);
+    } else {
+      String envCredentialsPath = getEnv(CERTIFICATE_CONFIGURATION_ENV_VARIABLE);
+      if (envCredentialsPath != null && !envCredentialsPath.isEmpty()) {
```
lqiu96, line 109: nit: Strings.nullOrEmpty

oauth2_http/javatests/com/google/auth/mtls/WorkloadCertificateConfigurationTest.java
```diff
@@ -0,0 +1,146 @@
...
+
+    try {
+      WorkloadCertificateConfiguration config =
+          WorkloadCertificateConfiguration.fromCertificateConfigurationStream(configStream);
+      fail("Expected an error due to missing the certificate path field.");
+    } catch (IllegalArgumentException e) {
+      assertEquals(
+          "The cert_path field must be provided in the workload certificate configuration.",
+          e.getMessage());
+    }
```
lqiu96, line 71: nit: for the failure test cases, can we use [assertThrows](https://junit.org/junit4/javadoc/latest/)

lqiu96 (approved):
Changes generally LGTM. Added a few nits and things about comments to warn uses about these classes (intended to be internal).

oauth2_http/java/com/google/auth/mtls/X509Provider.java
aeitzman: https://github.com/googleapis/google-auth-library-java/blob/c78065c351bc65a05eff67e568acfe143109d7cb/oauth2_http/java/com/google/auth/oauth2/GoogleAuthUtils.java#L67 yeah, I just copies the logic from our existing method for finding the Application Default file location

oauth2_http/java/com/google/auth/mtls/X509Provider.java
aeitzman: Added a comment explaining that it is internal and warning that we will not support its backward compat.

oauth2_http/java/com/google/auth/mtls/X509Provider.java
```diff
@@ -120,13 +146,13 @@ private WorkloadCertificateConfiguration getWorkloadCertificateConfiguration()
...
       }
-      certConfigStream = readStream(certConfig);
+      certConfigStream = createInputStream(certConfig);
       return WorkloadCertificateConfiguration.fromCertificateConfigurationStream(certConfigStream);
     } catch (Exception e) {
       // Although it is also the cause, the message of the caught exception can have very
       // important information for diagnosing errors, so include its message in the
       // outer exception message also.
-      throw new IOException(
+      throw new CertificateSourceUnavailableException(
```
andyrzhao, line 155: So "fromCertificateConfigurationStream" now throws two types of exceptions, illegal argument and the new sentinel error - this part seems right. However, the catch-all-exception clause on line 151 appears to swallow all the errors and wrap it as the new sentinel error, which doesn't seem right - an illegal argument exception could indicate a typo (vs intentionally leaving out the workload section.) Would it be better to remove the catch on line 151 altogether so the caller can handle the raw exceptions?

andyrzhao (approved):
Thanks for working on this PR! LGTM!

oauth2_http/java/com/google/auth/mtls/WorkloadCertificateConfiguration.java
```diff
@@ -0,0 +1,99 @@
...
+    if (certConfigs == null) {
+      throw new IllegalArgumentException(
+          "The cert_configs object must be provided in the certificate configuration file.");
+    }
+
+    Map<String, Object> workloadConfig = (Map<String, Object>) certConfigs.get("workload");
+    if (workloadConfig == null) {
+      throw new CertificateSourceUnavailableException(
+          "A workload certificate configuration must be provided in the cert_configs object.");
+    }
```
lqiu96, line 83: can you add a small note why this one exception is different? IIRC, it is because this case can still be used to indicate fallback, right?

oauth2_http/javatests/com/google/auth/mtls/X509ProviderTest.java
```diff
@@ -0,0 +1,245 @@
...
+    testProvider.addFile(keyPath, new ByteArrayInputStream(TEST_PRIVATE_KEY.getBytes()));
+
+    CertificateFactory cf = CertificateFactory.getInstance("X.509");
+    Certificate expectedCert =
+        cf.generateCertificate(new ByteArrayInputStream(TEST_CERT.getBytes()));
+
+    // Assert that the store has the expected certificate and only the expected certificate.
+    KeyStore store = testProvider.getKeyStore();
+    assertTrue(store.size() == 1);
+    assertTrue(store.getCertificateAlias(expectedCert) != null);
```
lqiu96, line 137: nit: Can this be
assertEquals(1, store.size()) and assertNotNull(store.getCertificateAlias(expectedCert)).

I think if this line is wrong, the logs would be something like `expected true (got false)`, which may not be super intuitive as to what went wrong. I think this would show up as something like `expected 1 (got 2)`.

oauth2_http/javatests/com/google/auth/mtls/X509ProviderTest.java
```diff
@@ -0,0 +1,245 @@
...
```
lqiu96, line 163: nit: same for here

oauth2_http/java/com/google/auth/mtls/WorkloadCertificateConfiguration.java
aeitzman: Yeah, this exception means that the calling method (the mtls code in the transport layer) should move on to the next certificate source rather than throwing an exception to the user and breaking the flow.

PR #1630, opened 2025-01-24
fix: respect token_uri from json in UserCredentials creation.

zhumin8:
context: b/382442857

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/UserCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/UserCredentials.java
@@ -124,6 +124,7 @@ public class UserCredentials extends GoogleCredentials implements IdTokenProvide
           "Error reading user credential from JSON, "
               + " expecting 'client_id', 'client_secret' and 'refresh_token'.");
     }
+    // currently "token_uri" is not a default field and needs to be added to json file manually
     String tokenUrl = (String) json.get("token_uri");
     URI tokenUri = (tokenUrl == null || tokenUrl.isEmpty()) ? null : URI.create(tokenUrl);
     return UserCredentials.newBuilder()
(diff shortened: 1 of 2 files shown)


sai-sunder-s (approved):
Thank you for the quick fix!

oauth2_http/java/com/google/auth/oauth2/UserCredentials.java
```diff
@@ -124,13 +124,15 @@ static UserCredentials fromJson(Map<String, Object> json, HttpTransportFactory t
           "Error reading user credential from JSON, "
               + " expecting 'client_id', 'client_secret' and 'refresh_token'.");
     }
+    String tokenUrl = (String) json.get("token_uri");
+    URI tokenUri = (tokenUrl == null || tokenUrl.isEmpty()) ? null : URI.create(tokenUrl);
```
lqiu96, line 128: qq, is this a value that exists? I was looking at my credentials.json file in the well known ADC path and I don't think I see an entry for tokenUri. Perhaps I needed additional flags for `gcloud application-default login`?

zhumin8: I don't think this exists in the standard credentials.json file. Rather iiuc, we are providing a workaround that if user edit their json file and add this field, we use value from it. (ref: http://b/382442857#comment73)

lqiu96: Can we add that as a quick comment above? I'm not entirely sure that this is going to be permanent/ how long we intend to support this. 

And given the prior stuff about malicious URIs, perhaps a warning about modifying the token_uri/ using a JSON with a malicious token_uri?

zhumin8: Good point. Although in this case, user has to explicitly provide this. @sai-sunder-s do we need to add similar warning here?

sai-sunder-s: I do not think we need a warning here because even if url is malicious, we will be sending it a token that is built purely using the data in the json.

lqiu96 (approved):

PR #1642, opened 2025-01-31
chore: add Go team to members for 2P reviews

meltsufin:


suztomo:
yoshi-automation for code generation?

PR #1668, opened 2025-02-07
ci: Add required checks for main branch

lqiu96:


meltsufin (commented):
sync-repo-settings app is suspended. So, this won't work.

lqiu96:
> sync-repo-settings app is suspended. So, this won't work.

Oh gotcha. I don't have access to the Settings tab. If you have access, could you add the checks as required? Mostly wanted to add the GraalVM ones

meltsufin:
Configured.

PR #1706, opened 2025-03-10
fix: Handle optional fields with null JSON value in ExternalAccountCredentials gracefully

### LINKED ISSUE #1705: ExternalAccountCredentials.fromStream() throws exception on optional fields with null value
#### Steps to reproduce

  1. Try to load the following JSON configuration using ExternalAccountCredentials.fromStream()

```json
{
  "service_account_impersonation_url": null, //optional field, would be similar with any other
  "audience": "audience",
  "subject_token_type": "subjectTokenType",
  "credential_source": {"file":"file"}
}
```

#### Code example

```java
  ExternalAccountCredentials credentials = ExternalAccountCredentials.fromStream(
          new ByteArrayInputStream(("""
                  {\
                  "service_account_impersonation_url": null,\
                  "audience": "audience",\
                  "subject_token_type": "subjectTokenType",\
                  "credential_source": {"file":"file"}\
                  }""").getBytes()));
```

#### Stack trace
```
com.google.auth.oauth2.CredentialFormatException: An invalid input stream was provided.

	at com.google.auth.oauth2.ExternalAccountCredentials.fromStream(ExternalAccountCredentials.java:398)
	at com.google.auth.oauth2.ExternalAccountCredentials.fromStream(ExternalAccountCredentials.java:366)
	at com.google.auth.oauth2.ExternalAccountCredentialsTest.fromStream_nullOptionalField(ExternalAccountCredentialsTest.java:153)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:63)
	at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
	at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:413)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:69)
	at com.intellij.rt.junit.IdeaTestRunner$Repeater$1.execute(IdeaTestRunner.java:38)
	at com.intellij.rt.execution.junit.TestsRepeater.repeat(TestsRepeater.java:11)
	at com.intellij.rt.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:35)
	at com.intellij.rt.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:232)
	at com.intellij.rt.junit.JUnitStarter.main(JUnitStarter.java:55)
Caused by: java.lang.ClassCastException: class java.lang.Object cannot be cast to class java.lang.String (java.lang.Object and java.lang.String are in module java.base of loader 'bootstrap')
	at com.google.auth.oauth2.ExternalAccountCredentials.fromJson(ExternalAccountCredentials.java:422)
	at com.google.auth.oauth2.ExternalAccountCredentials.fromStream(ExternalAccountCredentials.java:396)
	... 28 more```

#### Any additional information below

This is caused by com.google.api.client.json.JsonParser used to parse the JSON payload.
This implementation does not deserializes null JSON value to null reference in Java, instead using magic null objects from com.google.api.client.util.Data => this causes this ClassCastException when trying to deserialize, as the magic object can't be cast to the expected field type.

The current behavior does not follow the principle of least astonishment, that be that a null value for a specific optional field in JSON configuration is just considered as missing.

This issue is likely to happen for other types of JSON credential configuration that have optional values.

### PR DESCRIPTION
**Pull request by clementdenis**
This PR handles correctly JSON null values for optional fields in ExternalAccountCredentials configuration.

This is likely a partial fix, other credential classes have the same issue (just search for `(String) json.get(` in the codebase).

The approach might applied to all json fields, including required ones, to avoid ClassCastException and have better errors.

Fixes #1705 ☕️

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
+++ b/oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
@@ -36,6 +36,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import com.google.api.client.http.HttpHeaders;
 import com.google.api.client.json.GenericJson;
 import com.google.api.client.json.JsonObjectParser;
+import com.google.api.client.util.Data;
 import com.google.auth.RequestMetadataCallback;
 import com.google.auth.http.HttpTransportFactory;
 import com.google.common.base.MoreObjects;
@@ -418,15 +419,16 @@ public abstract class ExternalAccountCredentials extends GoogleCredentials {
     Map<String, Object> credentialSourceMap = (Map<String, Object>) json.get("credential_source");
 
     // Optional params.
-    String serviceAccountImpersonationUrl = (String) json.get("service_account_impersonation_url");
-    String tokenInfoUrl = (String) json.get("token_info_url");
-    String clientId = (String) json.get("client_id");
-    String clientSecret = (String) json.get("client_secret");
-    String quotaProjectId = (String) json.get("quota_project_id");
-    String userProject = (String) json.get("workforce_pool_user_project");
-    String universeDomain = (String) json.get("universe_domain");
+    String serviceAccountImpersonationUrl =
+        getOptional(json, "service_account_impersonation_url", String.class);
+    String tokenInfoUrl = getOptional(json, "token_info_url", String.class);
+    String clientId = getOptional(json, "client_id", String.class);
+    String clientSecret = getOptional(json, "client_secret", String.class);
+    String quotaProjectId = getOptional(json, "quota_project_id", String.class);
+    String userProject = getOptional(json, "workforce_pool_user_project", String.class);
+    String universeDomain = getOptional(json, "universe_domain", String.class);
     Map<String, Object> impersonationOptionsMap =
-        (Map<String, Object>) json.get("service_account_impersonation");
+        getOptional(json, "service_account_impersonation", Map.class);
 
     if (impersonationOptionsMap == null) {
       impersonationOptionsMap = new HashMap<String, Object>();
@@ -479,6 +481,14 @@ public abstract class ExternalAccountCredentials extends GoogleCredentials {
         .setServiceAccountImpersonationOptions(impersonationOptionsMap)
         .setUniverseDomain(universeDomain)
         .build();
+  }
+
+  private static <T> T getOptional(Map<String, Object> json, String fieldName, Class<T> clazz) {
+    Object value = json.get(fieldName);
+    if (value == null || Data.isNull(value)) {
+      return null;
+    }
+    return clazz.cast(value);
   }
 
   private static boolean isPluggableAuthCredential(Map<String, Object> credentialSource) {
--- a/oauth2_http/javatests/com/google/auth/oauth2/ExternalAccountCredentialsTest.java
+++ b/oauth2_http/javatests/com/google/auth/oauth2/ExternalAccountCredentialsTest.java
@@ -149,6 +149,23 @@ public class ExternalAccountCredentialsTest extends BaseSerializationTest {
   }
 
   @Test
+  public void fromStream_nullOptionalField() throws IOException {
+    ExternalAccountCredentials credentials =
+        ExternalAccountCredentials.fromStream(
+            new ByteArrayInputStream(
+                ("{"
+                        + "\"service_account_impersonation_url\": null,"
+                        +
+                        // required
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java [hunk 1]
```diff
@@ -481,6 +482,14 @@ static ExternalAccountCredentials fromJson(
         .build();
   }
 
+  private static <T> T getOptional(Map<String, Object> json, String fieldName, Class<T> clazz) {
+    Object value = json.get(fieldName);
+    if (value == null || Data.isNull(value)) {
+      return null;
+    }
```
lqiu96, line 489: qq, why the need for an additional `Data.isNull(value)` check?

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
```diff
@@ -481,6 +482,14 @@ static ExternalAccountCredentials fromJson(
...
+    return clazz.cast(value);
```
lqiu96, line 490: Can we just cast this as a `(String)` when we return? Or is there a specific reason to use `Class.cast()`?

lqiu96:
For the PR, can you update title to use [convention commits](https://www.conventionalcommits.org/en/v1.0.0/)? I think can just add a `fix: ...` prefix to the title.

Also, I think the linter CI is failing. Can you run `mvn fmt:format` on this change?

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
lqiu96: Oh nvm. I see after pulling this down locally.

lqiu96: Not entirely sure why the value comes back as an `Object` instead of null and I'm not too familiar with the parser inside the http-client. I don't think this check adds any harm since null values from the file wouldn't have worked and this allows the client to gracefully continue for optional fields.

lqiu96:
Adding @aeitzman for review on 3pi code.

oauth2_http/java/com/google/auth/oauth2/ExternalAccountCredentials.java
clementdenis: It's also handling impersonationOptionsMap, which is a Map, not a String.

lqiu96 (approved):
LGTM. @aeitzman to review as a 3PI reviewer in case I'm missing any context about ExternalAccountCredentials and null values.

lqiu96:
Adding a different reviewer as there have been some changes.

PR #1717, opened 2025-03-26
Adds an X509 Certificate Provider to support Workload Identity Federation

andyrzhao:
This X509 Certificate Provider will be used to support both a new cert-based Auth flow for Workload Identity Federation as will as enable X509 bound mTLS transport in the Java GAX library. This is the first step for bringing Java SDK to feature-parity with Golang and Python SDKs in terms of WIF support.


lqiu96:
I think this may be failing the CLA check as it has my non-google email as part of the commit. Could you raise this PR from the X509 branch inside this repo instead of Alex's fork?

PR #1719, opened 2025-03-27
feat: Adds an X509 Certificate Provider to support Workload Identity Federation. (#1624)

andyrzhao:
This X509 Certificate Provider will be used to support both a new cert-based Auth flow for Workload Identity Federation as will as enable X509 bound mTLS transport in the Java GAX library. This is the first step for bringing Java SDK to feature-parity with Golang and Python SDKs in terms of WIF support.


lsirac:
@nbayati is working on this and will have a PR up shortly. Thanks Andy.

PR #1728, opened 2025-04-01
fix: Do not add padding in Client-Side CAB tokens.

huangjiahua:
There are early adopters reporting that the padding at the end of the Client-Side CAB tokens can break some use cases including using URL-encoded token in URL. While we have other solutions to resolve this issue, we think it's better to remove the padding in the client library to stay consistent with other types of tokens.

### DIFF
--- a/cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
+++ b/cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
@@ -207,9 +207,12 @@ public class ClientSideCredentialAccessBoundaryFactory {
     byte[] encryptedRestrictions = this.encryptRestrictions(rawRestrictions, sessionKey);
 
     // withoutPadding() is used to stay consistent with server-side CAB
-    // withoutPadding() avoids additional URL encoded token issues (i.e. extra equal signs `=` in the path)
+    // withoutPadding() avoids additional URL encoded token issues (i.e. extra equal signs `=` in
+    // the path)
     String tokenValue =
-        intermediateToken + "." + Base64.getUrlEncoder().withoutPadding().encodeToString(encryptedRestrictions);
+        intermediateToken
+            + "."
+            + Base64.getUrlEncoder().withoutPadding().encodeToString(encryptedRestrictions);
 
     return new AccessToken(tokenValue, intermediateTokenExpirationTime);
   }
--- a/cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
+++ b/cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
@@ -206,6 +206,8 @@ public class ClientSideCredentialAccessBoundaryFactory {
 
     byte[] encryptedRestrictions = this.encryptRestrictions(rawRestrictions, sessionKey);
 
+    // withoutPadding() is used to stay consistent with server-side CAB
+    // withoutPadding() avoids additional URL encoded token issues (i.e. extra equal signs `=` in the path)
     String tokenValue =
         intermediateToken + "." + Base64.getUrlEncoder().withoutPadding().encodeToString(encryptedRestrictions);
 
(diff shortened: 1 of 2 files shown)


cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
```diff
@@ -207,7 +207,7 @@ public AccessToken generateToken(CredentialAccessBoundary accessBoundary)
     byte[] encryptedRestrictions = this.encryptRestrictions(rawRestrictions, sessionKey);
 
     String tokenValue =
-        intermediateToken + "." + Base64.getUrlEncoder().encodeToString(encryptedRestrictions);
+        intermediateToken + "." + Base64.getUrlEncoder().withoutPadding().encodeToString(encryptedRestrictions);
```
nbayati, line 210: Can you please update the relevant unit tests to reflect this change, and add a comment here explaining why we're using `withoutPadding()`.

huangjiahua: Added tests.

nbayati: Thanks! Can you also add a comment here documenting why there shouldn't be any padding?

huangjiahua: Added.

lqiu96: nit: Can you add a bit more to explain the `why`? Is there a specific behavior/requirement from server-side CAB that needs `withoutPadding` or just something that was always done? If not, can this comment be added like `withoutPadding() avoids URL encoded token issues (i.e. `=` in the path)`, etc.

cab-token-generator/javatests/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactoryTest.java
```diff
@@ -745,6 +745,9 @@ public void generateToken_withAvailablityCondition_success() throws Exception {
     CabToken cabToken = parseCabToken(token);
     assertEquals("accessToken", cabToken.intermediateToken);
 
+    // Verifies the encrypted restriction has no padding
+    assertFalse(cabToken.encryptedRestriction.contains(String.valueOf("=")));
```
lqiu96, line 749: qq, does the Base64.Encoder keep the `=` by default? I was assuming that it was keeping something like additional spaces at the end. Would the test case make more sense to ensure that the token value doesn't end with spaces? If so, do you have an example/ mock token that runs into these padding issues?

huangjiahua: `=` is used to [pad the base64 encoding output](https://en.wikipedia.org/wiki/Base64#Output_padding) if the input length in bytes is not a multiple of 3. This is the default behavior for Base64.Encoder if `withoutPadding` is not used. The base64 encoded restriction of this test case includes the padding at the end.

cab-token-generator/java/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactory.java
huangjiahua: Done.

cab-token-generator/javatests/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactoryTest.java
lqiu96: Gotcha, thanks! Can you just add a small comment above in the test to explain that padding adds `=` for values not in multiple of 3 (just for future maintainers who don't know this/ aren't familiar with padding)

lqiu96 (approved):
LGTM, thanks!

cab-token-generator/javatests/com/google/auth/credentialaccessboundary/ClientSideCredentialAccessBoundaryFactoryTest.java
huangjiahua: Added.

lqiu96:
@zhumin8 Could you also take a look and approve?

lqiu96:
@huangjiahua FYI, I think we have a lint issue. Can you run with mvn format?

zhumin8 (approved):

nbayati (approved):

PR #1730, opened 2025-04-07
feat(mtls): Introduce DefaultMtlsProviderFactory and SecureConnectProvider

andyrzhao:
This PR allows the Java auth lib to provide either an X509 mTLS provider or SecureConnect mTLS provider based on availability and under a common "MtlsProvider" interface. The Java GAX lib will be refactored to use this factory and interface in a subsequent PR.


oauth2_http/java/com/google/auth/mtls/DefaultMtlsProviderFactory.java
```diff
@@ -0,0 +1,66 @@
...
+    MtlsProvider mtlsProvider;
+    try {
+      mtlsProvider = new X509Provider();
+      mtlsProvider.getKeyStore();
+      return mtlsProvider;
+    } catch (CertificateSourceUnavailableException e) {
+      try {
+        mtlsProvider = new SecureConnectProvider();
+        mtlsProvider.getKeyStore();
+        return mtlsProvider;
```
lqiu96, line 59: Can `SecureConnectProvider` be added in a future PR? Or is this required for this PR? From my last discussion with Alex, the requirements was the inclusion of X.509 and having the Auth library STS functionality use the X.509 MtlsProvider.

I'm a bit concerned that `SecureConnectProvider` will exist in both Gax and Auth at the same time. I know that in the future that will commit to this existing in this repo.

oauth2_http/java/com/google/auth/mtls/ContextAwareMetadataJson.java
```diff
@@ -0,0 +1,48 @@
...
+
+package com.google.auth.mtls;
+
+import com.google.api.client.json.GenericJson;
+import com.google.api.client.util.Key;
+import com.google.common.collect.ImmutableList;
+import java.util.List;
+
+/** Data class representing context_aware_metadata.json file. */
+public class ContextAwareMetadataJson extends GenericJson {
```
lqiu96, line 39: Are the users expected to parse the json and create this class themselves? Will we have default functionality to help them create `ContextAwareMetadataJson` classes?

oauth2_http/java/com/google/auth/mtls/DefaultMtlsProviderFactory.java
andyrzhao: Thanks for taking a look Lawrence! So this PR is complementary to Alex's work to support X.509-based ADC auth. The goal of this PR is to bring Java SDK to feature parity with Golang SDK and Python SDK, both of which supports X.509-based ADC auth as well as X.509-backed mTLS transports (alongside SecureConnect backed mTLS transports). This is basically the final checkbox that allows the CBA team to declare client-side GA w.r.t X.509 work. 

I think the fact that the SecureConnectProvider will exist in both Gax and Auth at the same time (in the near term) is a small inconvenience that we can solve with documentation updates. Btw, you will note that the SecureConnectProvider implementation in this CL does not have any of the "CBA environment variable look up logic" present in the gax version - this is intentional: I've already started a separate refactoring effort on the GAX side to create a new helper class "CertificateBasedAccess" that will handle all env var checks and endpoint calculations. This is necessary to ensure that both X.509 mTLS and SecureConnect mTLS are gated behind the same env var variables, regardless of which mTLS provider is used.

oauth2_http/java/com/google/auth/mtls/ContextAwareMetadataJson.java
andyrzhao: The ContextAwareMetadataJson was introduced along side SecureConnectNative helper 4+ years ago, and is bootstrapped on all corp machines - it is entirely internal to Google at this point, and no users are expected to modify the json at all. It contains a command that will print out a self-signed EndpointVerification cert to use for SecureConnect based mTLS. This will eventually be replaced by ECP down the road.

oauth2_http/java/com/google/auth/mtls/DefaultMtlsProviderFactory.java
```diff
@@ -0,0 +1,66 @@
+/*
+ * Copyright 2025, Google Inc. All rights reserved.
```
zhumin8, line 2: nit: [go/copyright](https://goto.google.com/copyright), same below
```suggestion
 * Copyright 2025 Google LLC
```

oauth2_http/java/com/google/auth/mtls/MtlsProvider.java
```diff
@@ -0,0 +1,40 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package com.google.auth.mtls;
+
+import java.io.IOException;
+import java.security.KeyStore;
+
+public interface MtlsProvider {
```
zhumin8, line 37: Can we have some comment about this and the MtlsProvider in Gax exists together in the short term, while we figure out longer term plans. Just to remind us the context when coming back to this code.
similarly, for SecureConnectProvider.

oauth2_http/java/com/google/auth/mtls/DefaultMtlsProviderFactory.java
andyrzhao: Fixed! (Used the format used by gax lib)

oauth2_http/java/com/google/auth/mtls/MtlsProvider.java
andyrzhao: Thanks for the suggestion! I added comments here and elsewhere.

oauth2_http/java/com/google/auth/mtls/DefaultMtlsProviderFactory.java
lqiu96: Ok, I think the SecureConnectProvider duplication is OK for now, given that the implementation isn't intended to be used by customers and we don't expose this on the public surface for customers to set anyways. There shouldn't be any confusion from their POV. 

I'll need to double check this behavior in our downstream libraries, but i'm 90% confident that it can be fine for now.

oauth2_http/java/com/google/auth/mtls/DefaultMtlsProviderFactory.java
```diff
@@ -0,0 +1,65 @@
...
+      mtlsProvider = new X509Provider();
+      mtlsProvider.getKeyStore();
+      return mtlsProvider;
+    } catch (CertificateSourceUnavailableException e) {
+      try {
+        mtlsProvider = new SecureConnectProvider();
+        mtlsProvider.getKeyStore();
+        return mtlsProvider;
+      } catch (CertificateSourceUnavailableException ex) {
+        throw new CertificateSourceUnavailableException(
```
zhumin8, line 60: Is it useful here to wrap the the original exception? It's usually good practice for debugging to wrap the original exception (the one from X509Provider or SecureConnectProvider) as the cause.

oauth2_http/java/com/google/auth/mtls/X509Provider.java
```diff
@@ -89,6 +88,7 @@ public X509Provider() {
    * @return a KeyStore containing the X.509 certificate specified by the certificate configuration.
    * @throws IOException if there is an error retrieving the certificate configuration.
    */
+  @Override
```
zhumin8, line 91: Not part of this pr, but consider these are recent added code, can you help refactor the "finally" block with "try-with-resources"? go/java-practices/closing#try

oauth2_http/java/com/google/auth/mtls/SecureConnectProvider.java
```diff
@@ -0,0 +1,161 @@
...
+      throws IOException, InterruptedException {
+    long startTime = System.currentTimeMillis();
+    long remainTime = timeoutMilliseconds;
+
+    // In the while loop, keep checking if the process is terminated every 100 milliseconds
+    // until timeout is reached or process is terminated. In getKeyStore we set timeout to
+    // 1000 milliseconds, so 100 millisecond is a good number for the sleep.
+    while (remainTime > 0) {
+      Thread.sleep(Math.min(remainTime + 1, 100));
+      remainTime -= System.currentTimeMillis() - startTime;
```
zhumin8, line 145: Instead of manual timeout logic implementation, for Java 8+, `Process.waitFor(long timeout, TimeUnit unit)` is the standard and much cleaner way to wait for a process with a timeout. 
If check "every 100 milliseconds" itself is not a requirement, this would simplify logic here, to something along the lines of
```
static int runCertificateProviderCommand(Process commandProcess, long timeoutMilliseconds)
    throws IOException, InterruptedException {
  boolean terminated = commandProcess.waitFor(timeoutMilliseconds, TimeUnit.MILLISECONDS);
  if (!terminated) {
    commandProcess.destroy();
    throw new IOException("Cert provider command timed out");
  }
  return commandProcess.exitValue();
}
```

oauth2_http/java/com/google/auth/mtls/SecureConnectProvider.java
```diff
@@ -0,0 +1,161 @@
...
+    }
+  }
+
+  @VisibleForTesting
+  static KeyStore getKeyStore(InputStream metadata, ProcessProvider processProvider)
+      throws IOException, InterruptedException, GeneralSecurityException {
+    Process process = processProvider.createProcess(metadata);
+
+    // Run the command and timeout after 1000 milliseconds.
+    int exitCode = runCertificateProviderCommand(process, 1000);
```
zhumin8, line 117: I know this is from existing code in Gax, but do you know any reasoning to this 1000 number that we can add to comment?

oauth2_http/java/com/google/auth/mtls/SecureConnectProvider.java
```diff
@@ -0,0 +1,161 @@
...
+  interface ProcessProvider {
+    public Process createProcess(InputStream metadata) throws IOException;
+  }
+
+  static class DefaultProcessProvider implements ProcessProvider {
+    @Override
+    public Process createProcess(InputStream metadata) throws IOException {
+      if (metadata == null) {
+        return null;
+      }
```
zhumin8, line 72: if `metadata == null` a valid scenario? If so, when later calling methods on process can lead to NullPointerException?
https://github.com/googleapis/google-auth-library-java/blob/7a2b2ce62a9acfb41992ca6268bb208186833498/oauth2_http/java/com/google/auth/mtls/SecureConnectProvider.java#L114-L123

oauth2_http/java/com/google/auth/mtls/DefaultMtlsProviderFactory.java
zhumin8: Looks like although we do not expose the public surface for customers to set, downstream handwritten libraries have handles to set  MtlsProvider from [EndpointContext](https://github.com/googleapis/sdk-platform-java/blob/a5d9b5857f5c7c0e787bac0e1ae25198430a7622/gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java#L241) or [InstantiatingGrpcChannelProvider](https://github.com/googleapis/sdk-platform-java/blob/a5d9b5857f5c7c0e787bac0e1ae25198430a7622/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java#L992)? (I don't know if that's something touched in your related GAX changes?) It is unlikely a handwritten library used it, but can you double check this?

oauth2_http/java/com/google/auth/mtls/DefaultMtlsProviderFactory.java
```diff
@@ -0,0 +1,65 @@
...
+   * @throws IOException if an I/O error occurs during provider creation.
+   */
+  public static MtlsProvider create() throws IOException {
+    MtlsProvider mtlsProvider;
+    try {
+      mtlsProvider = new X509Provider();
+      mtlsProvider.getKeyStore();
+      return mtlsProvider;
+    } catch (CertificateSourceUnavailableException e) {
+      try {
```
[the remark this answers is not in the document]
andyrzhao: Done. Added isCertificateSourceAvailable() to the MtlsProvider interface.

oauth2_http/java/com/google/auth/mtls/DefaultMtlsProviderFactory.java
andyrzhao: So with the update to use "isCertificateSourceAvailable()", we no longer have the raw exceptions here. I don't have a strong preference here either-way - this is a non-blocking error message for the general use-case.

oauth2_http/java/com/google/auth/mtls/X509Provider.java
andyrzhao: done.

oauth2_http/java/com/google/auth/mtls/SecureConnectProvider.java
andyrzhao: Done~

oauth2_http/java/com/google/auth/mtls/SecureConnectProvider.java
andyrzhao: Unfortunately not, it seems like an arbitrary interval based on the fact that the command usually finishes instantly or hangs. I guess I can mention this in the comment.

oauth2_http/java/com/google/auth/mtls/SecureConnectProvider.java
andyrzhao: Good point! "metadata == null" in createProcess is not a case we expect to reach within the current code logic, since "new FileInputStream" would result in "FileNotFoundException" first. To safeguard against null pointer possibility, I updated "return null" to "throw new IOException("Error creating Process: metadata is null");"

oauth2_http/java/com/google/auth/mtls/DefaultMtlsProviderFactory.java
andyrzhao: mTLSProvder references in EndpointContext and InstantiatingGrpcChannelProviders are internal logic that was authored around the same time the original mTLS support was added, and I'm 99% sure there are no one else is using the set MtlsProvider functionality (it's mainly used by unit tests from the same suite as far as I can tell.) We can check it again when working on the Gax refactoring. This PR alone should not introduce any breaking changes.

oauth2_http/java/com/google/auth/mtls/SecureConnectProvider.java
```diff
@@ -0,0 +1,156 @@
...
+      throws IOException {
+    JsonParser parser = new GsonFactory().createJsonParser(contextAwareMetadata);
+    ContextAwareMetadataJson json = parser.parse(ContextAwareMetadataJson.class);
+    return json.getCommands();
+  }
+
+  @VisibleForTesting
+  static int runCertificateProviderCommand(Process commandProcess, long timeoutMilliseconds)
+      throws IOException, InterruptedException {
+    boolean terminated = commandProcess.waitFor(timeoutMilliseconds, TimeUnit.MILLISECONDS);
```
zhumin8, line 149: TimeUnit is missed in imports.

oauth2_http/javatests/com/google/auth/mtls/SecureConnectProviderTest.java
```diff
@@ -0,0 +1,159 @@
...
+import static org.junit.Assert.assertTrue;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.security.GeneralSecurityException;
+import java.util.List;
+import org.junit.Test;
+
+public class SecureConnectProviderTest {
```
zhumin8, line 45: Does it makes sense to create similar tests for the new X509Provider?

oauth2_http/java/com/google/auth/mtls/SecureConnectProvider.java
andyrzhao: Fixed!

oauth2_http/javatests/com/google/auth/mtls/SecureConnectProviderTest.java
andyrzhao: So X509ProviderTest already exists (and was written by AION team), which should suffice for now?

zhumin8 (dismissed):
LGTM

oauth2_http/java/com/google/auth/mtls/DefaultMtlsProviderFactory.java
```diff
@@ -0,0 +1,60 @@
...
+   */
+  public static MtlsProvider create() throws IOException {
+    MtlsProvider mtlsProvider = new X509Provider();
+    if (mtlsProvider.isCertificateSourceAvailable()) {
+      return mtlsProvider;
+    }
+    mtlsProvider = new SecureConnectProvider();
+    if (mtlsProvider.isCertificateSourceAvailable()) {
+      return mtlsProvider;
+    }
```
lqiu96, line 56: Is there another way to determine which MtlsProvider to resolve to without essentially calling `getKeyStore()` (seeing that `isCertificateSourceAvailable` is essentially a wrapper around it)? I'm not the biggest fan of this and ideally there could a simpler way to do this instead of trying to create the entire keystore object and testing if an exception throws. 

What's the expected behavior if the auth library does `DefaultMtlsProviderFactory.create()` and the customer also does `MtlsProvider.getKeyStore()`? Should two different KeyStores be created or should it be same?

oauth2_http/java/com/google/auth/mtls/MtlsProvider.java
lqiu96: Can you create this as a GH issue and linked it above in the comments? I think it would be easier for future maintainers to tracker and follow.

oauth2_http/java/com/google/auth/mtls/DefaultMtlsProviderFactory.java
```diff
@@ -0,0 +1,60 @@
...
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package com.google.auth.mtls;
+
+import java.io.IOException;
+
+public class DefaultMtlsProviderFactory {
```
zhumin8, line 35: Can we also add a note similar to the one in SecureConnectProvider to state it only intend to be used internally?

oauth2_http/java/com/google/auth/mtls/X509Provider.java
```diff
@@ -135,19 +130,19 @@ public KeyStore getKeyStore() throws IOException {
...
     }
   }
 
+  @Override
+  public boolean isCertificateSourceAvailable() throws IOException {
+    try {
+      this.getKeyStore();
+    } catch (CertificateSourceUnavailableException e) {
+      return false;
+    }
```
lqiu96, line 142: qq, should this also catch IO exception as well so that it falls back to SecureConnect? Or does it only fallback on CertificateSourceUnavailableException

oauth2_http/java/com/google/auth/mtls/X509Provider.java
```diff
@@ -135,19 +130,19 @@ public KeyStore getKeyStore() throws IOException {
...
-        privateKeyStream.close();
-      }
-      if (certAndPrivateKeyStream != null) {
-        certAndPrivateKeyStream.close();
-      }
     }
   }
 
+  @Override
+  public boolean isCertificateSourceAvailable() throws IOException {
```
lqiu96, line 137: can this public method have javadocs. Same for the impl in SecureConnect

oauth2_http/java/com/google/auth/mtls/ContextAwareMetadataJson.java
lqiu96, line 39: Can this class be package-private? Or does this need to be public?

oauth2_http/java/com/google/auth/mtls/X509Provider.java
andyrzhao: Yup, will do.

zhumin8 (approved):

oauth2_http/java/com/google/auth/mtls/X509Provider.java
lqiu96: If the design decision is made to have IOExceptions be throw back, can we do two additional things:
1. Make the IOExceptions clearer as to where they stem from (SecureConnect vs X.509) to make this slightly more clear to users? i.e. the IOException message states something like `SecureConnect: Unable/ Error/ etc`. I'm worried that the slight fallback may not be obvious to users (javadocs certainly do help!)
2. Can you add this explanation as a small comment for not catching IOExceptions in `DefaultMtlsProviderFactory.create()` for future maintainers?

lqiu96 (approved):
Changes generally LGTM. Added a few small comments if you could take a look at.

andyrzhao:
> Changes generally LGTM. Added a few small comments if you could take a look at.

Thanks for the thorough review Lawrence and Min!

lqiu96 (approved):

PR #1736, opened 2025-04-22
feat: Add support for mTLS authentication via X.509 certificates

nbayati:
This commit introduces a new credential source type, 'certificate', enabling the use of mTLS for authentication with X.509 certificates. It includes the necessary logic to load certificate configurations (both explicit paths and default locations) and establish an mTLS-enabled transport.


oauth2_http/java/com/google/auth/mtls/X509Provider.java
```diff
@@ -75,6 +76,34 @@ public X509Provider() {
...
+   * @throws IOException if the certificate configuration cannot be found or loaded, or if the
+   *     configuration file does not specify a certificate path.
+   * @throws CertificateSourceUnavailableException if the configuration file is not found.
+   */
+  public String getCertificatePath() throws IOException {
+    if (loadedConfig == null) {
+      // Attempt to load the configuration. This call might throw IOException or
+      // CertificateSourceUnavailableException if loading fails.
+      loadedConfig = getWorkloadCertificateConfiguration();
+    }
```
lqiu96, line 97: > If the configuration has not been loaded yet (e.g., if {@link #getKeyStore()} has not been

qq, should `getKeyStore()` also store the result of `getWorkloadCertificateConfiguration()`? IIUC, the result should be cached between the two methods. 

Additionally, does it matter if either `getCertificatePath()` or `getKeyStore()` is called first? Would it be possible to call `getWorkloadCertificateConfiguration()` in the constructor so that whatever method gets called first has the result?

oauth2_http/java/com/google/auth/mtls/X509Provider.java
```diff
@@ -75,6 +76,34 @@ public X509Provider() {
...
+    String certPath = loadedConfig.getCertPath();
+    if (Strings.isNullOrEmpty(certPath)) {
+      // Ensure the loaded configuration actually contains the required path.
+      throw new IOException(
```
lqiu96, line 101: I think this should be `CertificateSourceUnavailableException`

oauth2_http/java/com/google/auth/mtls/X509Provider.java
lqiu96: Oh I moved a bit too fast on this one. I see below `getKeyStore` is updated below. Gonna cross that part out.

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentialSource.java
```diff
@@ -48,6 +50,157 @@ public class IdentityPoolCredentialSource extends ExternalAccountCredentials.Cre
...
+  }
+
+  /**
+   * Retrieves an optional string value from a map.
+   *
+   * @param map The map to retrieve from.
+   * @param key The key of the string value.
+   * @return The string value if present and of the correct type, otherwise null.
+   * @throws IllegalArgumentException if the value is present but not a string.
+   */
```
lqiu96, line 110: thanks for writing the javadocs! I think it isn't required for private methods, unless they're moreso on the complex end (lots of moving parts and logic)

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentialSource.java
```diff
@@ -48,6 +50,157 @@ public class IdentityPoolCredentialSource extends ExternalAccountCredentials.Cre
...
+              key, value.getClass().getSimpleName()));
+    }
+    return (String) value;
+  }
+  /**
+   * Represents the configuration options for X.509-based workload credentials (mTLS). It specifies
+   * how to locate and use the client certificate, private key, and optional trust chain for mutual
+   * TLS authentication.
+   */
+  public static class CertificateConfig implements java.io.Serializable {
```
lqiu96, line 129: qq, why does this class need to be Serializable?

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentialSource.java
```diff
@@ -48,6 +50,157 @@ public class IdentityPoolCredentialSource extends ExternalAccountCredentials.Cre
...
+        @Nullable Boolean useDefaultCertificateConfig,
+        @Nullable String certificateConfigLocation,
+        @Nullable String trustChainPath) {
+
+      boolean useDefault = useDefaultCertificateConfig != null && useDefaultCertificateConfig;
+      boolean locationIsPresent =
+          certificateConfigLocation != null && !certificateConfigLocation.isEmpty();
+
+      checkArgument(
+          !(!useDefault && !locationIsPresent),
```
lqiu96, line 175: nit: Can this just be `(useDefault || locationIsPresent)`?

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentialSource.java
```diff
@@ -48,6 +50,157 @@ public class IdentityPoolCredentialSource extends ExternalAccountCredentials.Cre
...
+          "credentials: \"certificate\" object must either specify a certificate_config_location or use_default_certificate_config should be true");
```
lqiu96, line 176: nit: I think this error message may be a bit confusing since users may not know where the `certificate object` comes from.

Would it be accurate to say something like `The certificate source must specify a certificate_config_location value or set the use_default_certificate_config value to be true`?

oauth2_http/java/com/google/auth/oauth2/CertificateIdentityPoolSubjectTokenSupplier.java
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  private static X509Certificate loadLeafCertificate(String path)
+      throws IOException, CertificateException {
+    byte[] leafCertBytes;
+    try {
+      // IdentityPoolCredentials should have already validated the path exists via X509Provider.
+      leafCertBytes = Files.readAllBytes(Paths.get(path));
+    } catch (InvalidPathException e) {
+      throw new IOException("Invalid certificate file path provided: " + path, e);
```
lqiu96, line 80: qq, why is this re-thrown as an IOException?

oauth2_http/java/com/google/auth/oauth2/CertificateIdentityPoolSubjectTokenSupplier.java
```diff
@@ -0,0 +1,122 @@
...
+    if (certData == null || certData.length == 0) {
+      throw new IllegalArgumentException("Invalid certificate data: empty or null input");
+    }
+
+    try {
+      CertificateFactory certificateFactory = CertificateFactory.getInstance("X.509");
+      InputStream certificateStream = new ByteArrayInputStream(certData);
+      return (X509Certificate) certificateFactory.generateCertificate(certificateStream);
+    } catch (CertificateException e) {
+      throw new CertificateException("Failed to parse X.509 certificate data.", e);
```
lqiu96, line 96: qq, why is this re-thrown? Is this for better error messages?

oauth2_http/java/com/google/auth/oauth2/CertificateIdentityPoolSubjectTokenSupplier.java
```diff
@@ -0,0 +1,122 @@
...
+      String encodedCert = encodeCert(leafCert);
+
+      JsonArray certChain = new JsonArray();
+      certChain.add(new JsonPrimitive(encodedCert));
+
+      return GSON.toJson(certChain);
+    } catch (CertificateException e) {
+      throw new IOException(
+          "Failed to parse certificate from: " + credentialSource.credentialLocation, e);
+    }
```
lqiu96, line 120: same here, is there any specific reasons some exceptions are re-thrown as IOExceptions?

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentialSource.java
nbayati: No worries! Would you recommend removing the javadocs?

oauth2_http/java/com/google/auth/mtls/X509Provider.java
nbayati: Yes, the order of getCertificatePath() and getKeyStore() doesn't matter with the current lazy loading. The first call will trigger getWorkloadCertificateConfiguration() if loadedConfig is null, and subsequent calls will reuse the cached result.

While loading in the constructor, as you suggested, would ensure the config is loaded upfront, I'm leaning towards keeping the lazy loading approach. Moving the getWorkloadCertificateConfiguration() call to the constructor would mean the X509Provider constructor could throw IOExceptions (e.g., file not found), which feels less intuitive for object creation.

However, I'm open to changing this if you strongly feel loading in the constructor is a better pattern here.

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentialSource.java
nbayati: Good question! CertificateConfig is marked as Serializable because it's used as a field within IdentityPoolCredentialSource. Since IdentityPoolCredentialSource itself needs to be Serializable (as it extends ExternalAccountCredentials.CredentialSource), my understanding is that any non-transient fields it holds, like the certificateConfig instance, must also implement Serializable to allow the containing object to be serialized correctly.

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentialSource.java
nbayati: The error uses the exact wording as our Go implementation, but I can change it to certificate "source" if you think it'd be more clear for the users.

oauth2_http/java/com/google/auth/oauth2/CertificateIdentityPoolSubjectTokenSupplier.java
nbayati: Yes, the main reason for re-throwing the CertificateException here is to provide a more specific error message ("Failed to **parse** X.509 certificate data.") that adds context about what operation was being attempted when the failure occurred.

oauth2_http/java/com/google/auth/oauth2/CertificateIdentityPoolSubjectTokenSupplier.java
nbayati: The main reason we catch CertificateException and re-throw it as an IOException here is to comply with the `IdentityPoolSubjectTokenSupplier` interface, which declares that `getSubjectToken()` throws IOException.

oauth2_http/java/com/google/auth/oauth2/CertificateIdentityPoolSubjectTokenSupplier.java
nbayati: We catch InvalidPathException and re-throw it as an IOException primarily because if `loadLeafCertificate()` throws InvalidPathException directly, the `getSubjectToken()` method (which calls it) is required by the `IdentityPoolSubjectTokenSupplier` interface to only throw IOException for checked exceptions.
So, we'd have to catch the InvalidPathException in getSubjectToken anyway and wrap it in an IOException there. Catching and wrapping it inside `loadLeafCertificate()` keeps the error handling closer to the source of the potential path issue and simplifies `getSubjectToken`.

oauth2_http/java/com/google/auth/mtls/X509Provider.java
lqiu96: I don't have any *strong* opinions on this and just wanted to see if there is any requirements for lazy loading (performance or business logic).

I also see that `getWorkloadCertificateConfiguration()` also throws an IOException so the constructor would need to catch for that. I think this is fine for now.

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentialSource.java
lqiu96: I think they're fine since they're already here!

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentialSource.java
lqiu96: I see, I will leave it up to you if consistency among languages is a concern.

My preference would be for the error message be more actionable. I think the wording above may be a bit unclear for a first time user as they may not know where the `certificate object` is coming from. If possible, could we specify that the certificate object is from the `certificate source` (or even something more specific)?

oauth2_http/java/com/google/auth/oauth2/CertificateIdentityPoolSubjectTokenSupplier.java
lqiu96: > So, we'd have to catch the InvalidPathException in getSubjectToken anyway and wrap it in an IOException there.

I believe `InvalidPathException` is a a runtime exception and you won't need to add the exception to the method signature. Since IOExceptions are not being explicitly caught and handled, I think `InvalidPathException` should can also just be thrown at runtime.

oauth2_http/java/com/google/auth/oauth2/CertificateIdentityPoolSubjectTokenSupplier.java
lqiu96: Ah understood, can you add a small comment with the above info?

oauth2_http/java/com/google/auth/oauth2/CertificateIdentityPoolSubjectTokenSupplier.java
lqiu96: Sg, thanks! Could you add a small comment regarding adding more info into the exception?

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentialSource.java
nbayati: Updated the error to: "Invalid 'certificate' configuration in credential source: Must specify either 'certificate_config_location' or set 'use_default_certificate_config' to true."
Let me know what you think!

oauth2_http/javatests/com/google/auth/oauth2/IdentityPoolCredentialsTest.java
```diff
@@ -35,74 +35,46 @@
...
-      new HashMap<String, Object>() {
-        {
-          put("file", "file");
-        }
-      };
-
-  private static final IdentityPoolCredentialSource FILE_CREDENTIAL_SOURCE =
-      new IdentityPoolCredentialSource(FILE_CREDENTIAL_SOURCE_MAP);
-
-  private static final IdentityPoolCredentials FILE_SOURCED_CREDENTIAL =
```
nbayati, line 74: I moved these setups to helper functions because they are not needed for every tests.

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentialSource.java
```diff
@@ -80,9 +240,12 @@ public IdentityPoolCredentialSource(Map<String, Object> credentialSourceMap) {
     } else if (credentialSourceMap.containsKey("url")) {
       credentialLocation = (String) credentialSourceMap.get("url");
       credentialSourceType = IdentityPoolCredentialSourceType.URL;
+    } else if (credentialSourceMap.containsKey("certificate")) {
```
lqiu96, line 243: nit: Can the if/else blocks use the boolean variables above? `filePresent` and the others?

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
```diff
@@ -48,6 +53,8 @@ public class IdentityPoolCredentials extends ExternalAccountCredentials {
 
   static final String FILE_METRICS_HEADER_VALUE = "file";
   static final String URL_METRICS_HEADER_VALUE = "url";
+  static final String CERTIFICATE_METRICS_HEADER_VALUE = "certificate"; // Added constant
```
lqiu96, line 56: nit: remove comment

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
```diff
@@ -72,17 +80,60 @@ public class IdentityPoolCredentials extends ExternalAccountCredentials {
...
       this.subjectTokenSupplier = new FileIdentityPoolSubjectTokenSupplier(credentialSource);
       this.metricsHeaderValue = FILE_METRICS_HEADER_VALUE;
-    } else {
+    } else if (credentialSource.credentialSourceType == IdentityPoolCredentialSourceType.URL) {
       this.subjectTokenSupplier =
           new UrlIdentityPoolSubjectTokenSupplier(credentialSource, this.transportFactory);
       this.metricsHeaderValue = URL_METRICS_HEADER_VALUE;
+    } else if (credentialSource.credentialSourceType
+        == IdentityPoolCredentialSourceType.CERTIFICATE) {
+      assert credentialSource.certificateConfig != null;
```
lqiu96, line 97: qq, is this assert needed?

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
```diff
@@ -72,17 +80,60 @@ public class IdentityPoolCredentials extends ExternalAccountCredentials {
...
+        }
+
+        // Update the transport factory to use a mTLS transport with the provided certificate.
+        KeyStore mtlsKeyStore = x509Provider.getKeyStore();
+        final NetHttpTransport mtlsTransport =
+            new NetHttpTransport.Builder().trustCertificates(null, mtlsKeyStore, "").build();
+        this.transportFactory = () -> mtlsTransport;
+
+        // Initialize the subject token supplier with the certificate path
+        credentialSource.credentialLocation = x509Provider.getCertificatePath();
```
lqiu96, line 123: nit: Can we extract this to a helper function so the flow of the constructor generally does:
1. set `subjectTokenSupplier`
2. set `metricsHeaderValue`

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
```diff
@@ -144,6 +195,21 @@ public static class Builder extends ExternalAccountCredentials.Builder {
...
+     * Sets a custom {@link X509Provider} to manage the client certificate and private key for mTLS.
+     * If set, this provider will be used instead of the default behavior which initializes an
+     * {@code X509Provider} based on the {@code certificateConfigLocation} or default paths found in
+     * the {@code credentialSource}. This is primarily used for testing.
+     *
+     * @param x509Provider the custom X509 provider to use.
+     * @return this {@code Builder} object
+     */
+    @CanIgnoreReturnValue
+    public Builder setX509Provider(X509Provider x509Provider) {
```
lqiu96, line 208: Can this be package-private?

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
```diff
@@ -72,17 +80,60 @@ public class IdentityPoolCredentials extends ExternalAccountCredentials {
...
+          // Initialize X509Provider with the explicit path (if provided).
+          x509Provider = new X509Provider(explicitCertConfigPath);
+        }
+
+        // Update the transport factory to use a mTLS transport with the provided certificate.
+        KeyStore mtlsKeyStore = x509Provider.getKeyStore();
+        final NetHttpTransport mtlsTransport =
+            new NetHttpTransport.Builder().trustCertificates(null, mtlsKeyStore, "").build();
+        this.transportFactory = () -> mtlsTransport;
+
```
lqiu96, line 121: (not sure if this is possible) Just a thought: Would it be possible to create like a default mtls transport factory that is similar to https://github.com/googleapis/google-auth-library-java/blob/9f4cda9acad26ff9c6855cb729c4fde4c11bc6ea/oauth2_http/java/com/google/auth/oauth2/OAuth2Utils.java#L122-L128.

It could take in a mtlsKeyStore and be used for mTLS?

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
nbayati: That's a good question. I don't think it's really needed at this point. I removed it!

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
nbayati: Created a new class `MtlsHttpTransportFactory`

oauth2_http/java/com/google/auth/mtls/X509Provider.java
```diff
@@ -90,8 +119,10 @@ public X509Provider() {
    * @throws IOException if there is an error retrieving the certificate configuration.
    */
   public KeyStore getKeyStore() throws IOException {
-
-    WorkloadCertificateConfiguration workloadCertConfig = getWorkloadCertificateConfiguration();
+    if (loadedConfig == null) {
+      loadedConfig = getWorkloadCertificateConfiguration();
+    }
+    WorkloadCertificateConfiguration workloadCertConfig = loadedConfig;
```
lqiu96, line 125: qq, do we need this local workloadCertConfig object? Can we just read from `loadedConfig`?

oauth2_http/java/com/google/auth/oauth2/CertificateIdentityPoolSubjectTokenSupplier.java
```diff
@@ -0,0 +1,122 @@
...
+      throws IOException, CertificateException {
+    byte[] leafCertBytes;
+    leafCertBytes = Files.readAllBytes(Paths.get(path));
+    return parseCertificate(leafCertBytes);
+  }
+
+  @VisibleForTesting
+  static X509Certificate parseCertificate(byte[] certData) throws CertificateException {
+    if (certData == null || certData.length == 0) {
+      throw new IllegalArgumentException("Invalid certificate data: empty or null input");
```
lqiu96, line 83: nit: Can this be `"Invalid certificate data: Certificate file is empty or null"`

oauth2_http/java/com/google/auth/oauth2/CertificateIdentityPoolSubjectTokenSupplier.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+  }
+
+  private static String encodeCert(X509Certificate certificate)
+      throws CertificateEncodingException {
+    return Base64.getEncoder().encodeToString(certificate.getEncoded());
+  }
+
+  @Override
+  public String getSubjectToken(ExternalAccountSupplierContext context) throws IOException {
```
lqiu96, line 103: Can you add public javadocs for this method? Just a bit notes details any `CertificateIdentityPoolSubjectTokenSupplier` specific logic not specified in `IdentityPoolSubjectTokenSupplier`

oauth2_http/javatests/com/google/auth/oauth2/IdentityPoolCredentialsTest.java
```diff
@@ -1109,4 +1213,75 @@ private static IdentityPoolCredentialSource buildUrlBasedCredentialSource(
...
+    TestX509Provider(KeyStore keyStore, String certificatePath) {
+      super();
+      this.keyStore = keyStore;
+      this.certificatePath = certificatePath;
+    }
+
+    @Override
+    public KeyStore getKeyStore() {
+      if (shouldThrowOnGetKeyStore) {
+        throw new RuntimeException("Exception on get keystore");
```
lqiu96, line 1266: Could these methods throw a more specific exception type to mimic to actual behavior (i.e. IOException or the other type)?

oauth2_http/javatests/com/google/auth/oauth2/IdentityPoolCredentialsTest.java
```diff
@@ -1060,7 +1036,135 @@ public void serialize() throws IOException, ClassNotFoundException {
...
+    RuntimeException exception = assertThrows(RuntimeException.class, builder::build);
+
+    assertEquals(
+        "Failed to initialize IdentityPoolCredentials from certificate source due to an I/O error.",
+        exception.getMessage());
+  }
+
+  @Test
+  public void build_withCustomProvider_throwsOnGetKeyStore()
+      throws IOException, KeyStoreException, CertificateException, NoSuchAlgorithmException {
```
lqiu96, line 1114: Same here. I'm not 100% sure why it's supposed to fail. Is this to mimic the an IO error in the keystore?

oauth2_http/java/com/google/auth/mtls/MtlsHttpTransportFactory.java
```diff
@@ -0,0 +1,70 @@
...
+import com.google.api.client.http.javanet.NetHttpTransport;
+import com.google.auth.http.HttpTransportFactory;
+import java.security.GeneralSecurityException;
+import java.security.KeyStore;
+import java.util.Objects;
+
+/**
+ * An HttpTransportFactory that creates {@link NetHttpTransport} instances configured for mTLS
+ * (mutual TLS) using a specific {@link KeyStore} containing the client's certificate and private
+ * key.
```
lqiu96, line 43: Thanks! Can you also add a small warning about this being an internal class and not intended to be directly used by the user?

lqiu96 (approved):
I think things LGTM. Added a few additional comments.

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentialSource.java
```diff
@@ -45,9 +47,193 @@ public class IdentityPoolCredentialSource extends ExternalAccountCredentials.Cre
   private static final long serialVersionUID = -745855247050085694L;
   IdentityPoolCredentialSourceType credentialSourceType;
   CredentialFormatType credentialFormatType;
-  String credentialLocation;
+  private String credentialLocation;
```
nbayati, line 50: Made `credentialLocation` and `certificateConfig` fields private and added public getters to improve encapsulation and provide the ability to properly mock the object.

oauth2_http/javatests/com/google/auth/oauth2/IdentityPoolCredentialsTest.java
```diff
@@ -1060,7 +1036,135 @@ public void serialize() throws IOException, ClassNotFoundException {
...
+    RuntimeException exception =
+        assertThrows(
+            RuntimeException.class,
+            () -> createCredentialsWithCertificate(x509Provider, certificateMap));
+
+    assertEquals("Exception on get keystore", exception.getMessage());
+  }
+
+  @Test
+  public void build_withCustomProvider_throwsOnGetCertificatePath()
```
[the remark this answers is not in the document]
nbayati: Done!

oauth2_http/java/com/google/auth/mtls/X509Provider.java
```diff
@@ -75,6 +76,34 @@ public X509Provider() {
...
+   * <p>If the configuration has not been loaded yet (e.g., if {@link #getKeyStore()} has not been
+   * called), this method will attempt to load it first by searching the override path, environment
+   * variable, and well-known locations.
+   *
+   * @return The path to the certificate file.
+   * @throws IOException if the certificate configuration cannot be found or loaded, or if the
+   *     configuration file does not specify a certificate path.
+   * @throws CertificateSourceUnavailableException if the configuration file is not found.
+   */
+  public String getCertificatePath() throws IOException {
```
andyrzhao, line 92: So we previously envisioned X509Provider will implement an "MtlsProvider" interface with a single "getKeyStore()" public method exposed. (SecureConnectProvider will also implement the same interface) As such, it doesn't feel like the getCertificatePath() helper should live inside the X509Provider. I don't see a lot of value with caching the cert config loading (this pattern is not common in client libraries IMO given the low anticipated QPS for auth or setting up an mTLS transport) Furthermore, having 2 separate public methods sharing this internal state could make code execution unpredictable - what if the underlying cert config changed between the call between getCertificatePath and getKeyStore()? And getKeyStore() may be stuck with an outdated loadedConfig in memory as long is the object instance is alive with no way to refresh the internal state. We have to handle cert rotation in the future as well. IMO, it's much safer to call getWorkloadCertificateConfiguration on demand instead of attempting to cache the result. Rest of your PR looks good!

andyrzhao (changes requested):
LGTM overall. Small comment about X509Provider caching.

oauth2_http/java/com/google/auth/mtls/X509Provider.java
nbayati: Makes sense. I'll refactor move the getCertificatePath() method outside of the X509Provider class and remove caching.

oauth2_http/java/com/google/auth/oauth2/CertificateIdentityPoolSubjectTokenSupplier.java
```diff
@@ -0,0 +1,133 @@
...
+   *     this implementation.
+   * @return The JSON string representation of the base64-encoded leaf certificate in a JSON array.
+   * @throws IOException If an I/O error occurs while reading the certificate file.
+   */
+  @Override
+  public String getSubjectToken(ExternalAccountSupplierContext context) throws IOException {
+    try {
+      // credentialSource.credentialLocation is expected to be non-null here,
+      // set during IdentityPoolCredentials construction for certificate type.
+      X509Certificate leafCert = loadLeafCertificate(credentialSource.getCredentialLocation());
```
lsirac, line 118: Hmm, can you explain why we sending just the leaf? I guess you will add the entire trust chain in another PR?

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
```diff
@@ -131,9 +144,49 @@ public static Builder newBuilder(IdentityPoolCredentials identityPoolCredentials
...
+      // Initialize the subject token supplier with the certificate path.
+      credentialSource.setCredentialLocation(x509Provider.getCertificatePath());
+      return new CertificateIdentityPoolSubjectTokenSupplier(credentialSource);
+
+    } catch (IOException e) {
+      throw new RuntimeException(
+          // Wrap IOException in RuntimeException because constructors cannot throw checked
+          // exceptions.
+          "Failed to initialize IdentityPoolCredentials from certificate source due to an I/O error.",
```
lsirac, line 163: nit: It's a bit strange to have this method do this given this is a separate method, even if it is only called by the constructor

lsirac (approved):
Looks good!

oauth2_http/java/com/google/auth/oauth2/CertificateIdentityPoolSubjectTokenSupplier.java
nbayati: Yes, supporting the trust chain is coming in the next PR.

oauth2_http/java/com/google/auth/oauth2/IdentityPoolCredentials.java
nbayati: Moved the try/catch block to the constructor.

andyrzhao (approved):
lgtm. Thanks for working on this PR!

PR #1746, opened 2025-05-07
feat: Return X509 certificate chain as the subject token.

nbayati:
design: [go/x509-auth-lib-redesign](http://goto.google.com/x509-auth-lib-redesign)

The CertificateIdentityPoolSubjectTokenSupplier's subjectToken function now returns the full X509 certificate chain, including the leaf certificate and any provided trust chain certificates, as a JSON array of base64-encoded strings. This chain is used as the subject token for mTLS authentication.

Similar work was done in the [python](https://github.com/googleapis/google-auth-library-python/pull/1695/) and [Go](https://github.com/googleapis/google-cloud-go/pull/11948) libraries.


oauth2_http/java/com/google/auth/oauth2/CertificateIdentityPoolSubjectTokenSupplier.java
```diff
@@ -97,33 +103,153 @@ private static String encodeCert(X509Certificate certificate)
...
+
     try {
-      // credentialSource.credentialLocation is expected to be non-null here,
-      // set during IdentityPoolCredentials construction for certificate type.
-      X509Certificate leafCert = loadLeafCertificate(credentialSource.getCredentialLocation());
+      // Load the leaf certificate.
+      X509Certificate leafCert = loadLeafCertificate(leafCertPath);
       String encodedLeafCert = encodeCert(leafCert);
 
+      // Add the leaf certificate first.
```
lqiu96, line 131: qq, is there any specific reason why the leaf needs to be added first? I'm assuming the logic relies on the order

oauth2_http/java/com/google/auth/oauth2/CertificateIdentityPoolSubjectTokenSupplier.java
```diff
@@ -97,33 +103,153 @@ private static String encodeCert(X509Certificate certificate)
...
+      // Process the trust chain certificates read from the file.
+      if (!trustChainCerts.isEmpty()) {
+        // Check the first certificate in the trust chain file.
+        X509Certificate firstTrustCert = trustChainCerts.get(0);
+        String encodedFirstTrustCert = encodeCert(firstTrustCert);
+
+        // Add the first certificate only if it is not the same as the leaf certificate.
+        if (!encodedFirstTrustCert.equals(encodedLeafCert)) {
+          certChain.add(encodedFirstTrustCert);
+        }
```
lqiu96, line 147: Can you add the reason why the first certificate check doesn't throw an exception if it's the same as the leaf certificate to the comment above?

oauth2_http/java/com/google/auth/oauth2/CertificateIdentityPoolSubjectTokenSupplier.java
```diff
@@ -97,33 +103,153 @@ private static String encodeCert(X509Certificate certificate)
...
+    } catch (NoSuchFileException e) {
+      throw new IOException("Trust chain file not found: " + trustChainPath, e);
+    } catch (IOException e) {
+      throw new IOException("Failed to read trust chain file: " + trustChainPath, e);
+    }
+
+    // Split the file content into PEM certificate blocks.
+    String content = new String(trustChainData);
+    Pattern pemCertPattern =
+        Pattern.compile("-----BEGIN CERTIFICATE-----.*?-----END CERTIFICATE-----", Pattern.DOTALL);
```
lqiu96, line 221: Can the `Pattern` object be a static final constant? This would be compiling the regex every time  `readTrustChain()` is called.

lqiu96: Should the inner regex just be `.*`? Just to try an capture where there is the cert details between the `---CERTIFICATE---` blocks? I think it should fail during the parser below if it doesn't match.

oauth2_http/java/com/google/auth/oauth2/CertificateIdentityPoolSubjectTokenSupplier.java
```diff
@@ -97,33 +103,153 @@ private static String encodeCert(X509Certificate certificate)
...
+    Matcher matcher = pemCertPattern.matcher(content);
+
+    while (matcher.find()) {
+      String pemCertBlock = matcher.group(0);
```
lqiu96, line 225: nit: No param should match with the entire block
```suggestion
      String pemCertBlock = matcher.group();
```

oauth2_http/java/com/google/auth/oauth2/CertificateIdentityPoolSubjectTokenSupplier.java
lqiu96: Hmm, after looking at this again I think I'm a bit confused by this. I think if the leafCertificate isn't added in then an exception is thrown from `parseCertificate`. If that's the case, then this block would never be hit as we would be in the catch block below.

Do we need a separate try-catch block for loading the leaf certificate?

oauth2_http/java/com/google/auth/oauth2/CertificateIdentityPoolSubjectTokenSupplier.java
nbayati: Yes, the STS expects to see the certificates in a specific order. Here's the flow. I'll update the comment to clarify it.
Screenshot 2025-05-08 at 1 52 33 PM

oauth2_http/java/com/google/auth/oauth2/CertificateIdentityPoolSubjectTokenSupplier.java
nbayati: I think`.*` can lead to problems when we have multiple certificates concatenated in the trust chain file. 

If I'm understanding correctly, Using `.*` (greedy) would match from the first `-----BEGIN CERTIFICATE-----` all the way to the very last `-----END CERTIFICATE-----` in the file, effectively treating all certificates as one large chunk. This would prevent us from processing each certificate individually.

The ? in `.*?` makes the quantifier non-greedy, ensuring it matches the shortest possible string that still allows the `-----END CERTIFICATE-----` to match. This way we can extract each individual certificate block.

oauth2_http/java/com/google/auth/oauth2/CertificateIdentityPoolSubjectTokenSupplier.java
nbayati: I separated loading the leaf certificate and trust chain into their own try/catch blocks, and added inline comments to explainw why the first certificate check doesn't throw an exception if it's the same as the leaf certificate.

lsirac (approved):

oauth2_http/java/com/google/auth/oauth2/CertificateIdentityPoolSubjectTokenSupplier.java
lqiu96: I see, I think you're right. Thanks for the explanation!

oauth2_http/java/com/google/auth/oauth2/CertificateIdentityPoolSubjectTokenSupplier.java
```diff
@@ -97,33 +118,157 @@ private static String encodeCert(X509Certificate certificate)
...
 
-      java.util.List<String> certChain = new java.util.ArrayList<>();
-      certChain.add(encodedLeafCert);
+          // Throw an error if the current certificate (from the user-provided trust chain file,
+          // at an index beyond the first) is the same as the leaf certificate.
+          // This enforces that if the leaf certificate is included in the trust chain file by the
+          // user, it must be the very first certificate in that file. It should not appear
+          // elsewhere in the chain.
+          if (encodedCurrentCert.equals(encodedLeafCert)) {
+            throw new IllegalArgumentException(
```
lqiu96, line 183: Logic looks fine to me. If possible, would we be able to try and reduce the nesting a bit to make this a bit easier to follow.

Adding some of the if checks to return early (if possible)
```
if (trustChainCerts.isEmpty()) {
   return OAuth2Utils.JSON_FACTORY.toString(certChain)
}
```
potentially:
```
if (trustChainCerts.subList(1, end).stream().anyMatch(leafCert)) {
   throw new IllegalArgumentException()
}
```

oauth2_http/java/com/google/auth/oauth2/CertificateIdentityPoolSubjectTokenSupplier.java
```diff
@@ -97,33 +118,157 @@ private static String encodeCert(X509Certificate certificate)
...
+   * @throws IOException If an error occurs while reading the file.
+   * @throws CertificateException If an error occurs while parsing a certificate.
+   */
+  @VisibleForTesting
+  static List<X509Certificate> readTrustChain(String trustChainPath)
+      throws IOException, CertificateException {
+    List<X509Certificate> certificateTrustChain = new ArrayList<>();
+
+    // If no trust chain path is provided, return an empty list.
+    if (trustChainPath == null || trustChainPath.isEmpty()) {
```
lqiu96, line 227: nit: Strings.isNullOrEmpty()

oauth2_http/java/com/google/auth/oauth2/CertificateIdentityPoolSubjectTokenSupplier.java
```diff
@@ -97,33 +118,157 @@ private static String encodeCert(X509Certificate certificate)
...
+        // starts with an intermediate certificate. If the first certificate in the trust chain file
+        // *is* the leaf certificate, this means the user has explicitly included the leaf in their
+        // trust chain file. In this case, we skip adding it again to prevent duplication, as the
+        // leaf is already at the beginning of `certChain`.
+        if (!encodedFirstTrustCert.equals(encodedLeafCert)) {
+          certChain.add(encodedFirstTrustCert);
+        }
+
+        // Iterate over the remaining certificates in the trust chain.
+        for (int i = 1; i < trustChainCerts.size(); i++) {
```
andyrzhao, line 173: nit: iterating from 1 looks a bit funny to me especially given the code duplication for 0, I think the more idiomatic way to write this for-loop is to iterate from 0, then check for i==0 inside the for-loop for the special case handling.

oauth2_http/java/com/google/auth/oauth2/CertificateIdentityPoolSubjectTokenSupplier.java
```diff
@@ -97,33 +118,157 @@ private static String encodeCert(X509Certificate certificate)
...
    *     this implementation.
-   * @return The JSON string representation of the base64-encoded leaf certificate in a JSON array.
-   * @throws IOException If an I/O error occurs while reading the certificate file.
+   * @return The JSON string representation of the base64-encoded certificate chain (leaf
+   *     certificate followed by the trust chain, if present).
+   * @throws IOException If an I/O error occurs while reading the certificate file(s).
    */
   @Override
   public String getSubjectToken(ExternalAccountSupplierContext context) throws IOException {
+    String leafCertPath = credentialSource.getCredentialLocation();
```
andyrzhao, line 135: nit: No need to address in this PR, but it feels like the getCredentialLocation method should be renamed to getLeafCertPath() for clarity (unless it's a shared abstraction for non-X509 use cases as well?). The fact that the leaf cert is used as "Credentials" w.r.t to X.509 WIF is more of a higher-level business logic that's better addressed elsewhere.

andyrzhao (approved):
LGTM with some minor comments.

oauth2_http/java/com/google/auth/oauth2/CertificateIdentityPoolSubjectTokenSupplier.java
nbayati: Thanks for the suggestion! I see your point about the loop starting at 1 and the duplication of get()/encodeCert().

My reasoning for the current structure was to keep the logic for handling the first certificate in the trust chain (which has the rule: 'add if not leaf') distinctly separate from the logic for subsequent certificates (which have the rule: 'throw if leaf, else add'). I think this separation, aided by the comments, makes the different rules very explicit.

Combining them into a single loop with an if (i == 0) check would make the loop body a bit more branched. Given that the core conditional logic for adding/throwing is quite different for the first element versus the rest, I found the current separation slightly clearer to follow for these distinct cases.

oauth2_http/java/com/google/auth/oauth2/CertificateIdentityPoolSubjectTokenSupplier.java
nbayati: Thanks for the feedback on the trust chain processing and the ideas for reducing nesting!

Regarding the idea of an early return if `trustChainCerts.isEmpty()`: My main concern with returning `OAuth2Utils.JSON_FACTORY.toString(certChain)` directly from within that `if` block (which would be inside the `try-catch` for trust chain operations) is how it might interact with exception handling. If `toString()` itself threw an `IOException`, it could potentially be caught by the `catch (IOException e)` block that's intended for trust chain file reading errors (e.g., "Failed to read trust chain file..."). This might lead to a misleading error message for an issue that's actually related to JSON conversion. The current structure keeps the final toString() call outside that specific try-catch block.

For the suggestion of using a stream like `trustChainCerts.subList(1, end).stream().anyMatch`: A couple of things that gave me pause were:
1. `encodeCert()` throws a checked `CertificateEncodingException`. Handling and wrapping the checked exceptions within stream lambdas could be a bit verbose.
2.We'd still need to iterate through the certificates to encode them and add them to the certChain. The current for loop (from index 1) handles the encoding, validation (checking against encodedLeafCert), and adding in a single pass for the remaining certificates.

As an alternative to address the nesting, what do you think about extracting the `if (!trustChainCerts.isEmpty()) { ... }` block into a new private helper method, say `populateCertChainFromTrust()`? This would simplify getSubjectToken by encapsulating the trust chain processing details, and the existing try-catch in getSubjectToken would still handle exceptions from this helper. 

Let me know your thoughts. If that sounds good, I can refactor it.
Thanks!

oauth2_http/java/com/google/auth/oauth2/CertificateIdentityPoolSubjectTokenSupplier.java
```diff
@@ -50,6 +50,7 @@
 import java.util.List;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import jdk.internal.joptsimple.internal.Strings;
```
zhumin8, line 53: You probably meant Guava's Strings from com.google.common.base

nbayati: Thanks for catching this!

oauth2_http/java/com/google/auth/oauth2/CertificateIdentityPoolSubjectTokenSupplier.java
lqiu96: Gotcha, thanks for explaining. I think these are valid concerns and given our short timeframe for this, I don't think this is a blocker for the PR. I think this LGTM from my end and if we want to clean this up then we can follow up with this in a future PR.

oauth2_http/java/com/google/auth/oauth2/CertificateIdentityPoolSubjectTokenSupplier.java
```diff
@@ -97,33 +119,157 @@ private static String encodeCert(X509Certificate certificate)
...
+
+    // initialize certificate factory to retrieve x509 certificates.
+    CertificateFactory cf = CertificateFactory.getInstance("X.509");
+
+    // Read the trust chain file.
+    byte[] trustChainData;
+    trustChainData = Files.readAllBytes(Paths.get(trustChainPath));
+
+    // Split the file content into PEM certificate blocks.
+    String content = new String(trustChainData);
```
zhumin8, line 240: I wonder if it makes sense to be a bit safer with charset specified?
```suggestion
    String content = new String(trustChainData, StandardCharsets.UTF_8);
```

oauth2_http/java/com/google/auth/oauth2/CertificateIdentityPoolSubjectTokenSupplier.java
```diff
@@ -97,33 +119,157 @@ private static String encodeCert(X509Certificate certificate)
...
+      trustChainPath = credentialSource.getCertificateConfig().getTrustChainPath();
+    }
+
+    // Load and encode the leaf certificate.
+    String encodedLeafCert = loadAndEncodeLeafCertificate(leafCertPath);
+
+    // Initialize the certificate chain for the subject token. The Security Token Service (STS)
+    // requires that the leaf certificate (the one used for authenticating this workload) must be
+    // the first certificate in this chain.
+    java.util.List<String> certChain = new java.util.ArrayList<>();
```
zhumin8, line 148: nit:
```suggestion
   List<String> certChain = new ArrayList<>();
```

oauth2_http/java/com/google/auth/oauth2/CertificateIdentityPoolSubjectTokenSupplier.java
```diff
@@ -97,33 +119,157 @@ private static String encodeCert(X509Certificate certificate)
...
+    } catch (IllegalArgumentException e) {
+      // This catches the specific error for misconfigured trust chain (e.g., leaf in wrong place).
+      throw new IOException("Trust chain misconfiguration: " + e.getMessage(), e);
+    } catch (NoSuchFileException e) {
+      throw new IOException(String.format("Trust chain file not found: %s", trustChainPath), e);
     } catch (CertificateException e) {
-      // Catch CertificateException to provide a more specific error message including
-      // the path of the file that failed to parse, and re-throw as IOException
-      // as expected by the getSubjectToken method signature for I/O related issues.
+      // This catches parsing errors if certificates in the trust chain file are invalid.
```
zhumin8, line 198: nit: this comment seems a bit redundant, the exception message seems self-explainatory.

oauth2_http/java/com/google/auth/oauth2/CertificateIdentityPoolSubjectTokenSupplier.java
zhumin8: Perhaps extract a private method to improve readability of the nested logic?

oauth2_http/javatests/com/google/auth/oauth2/CertificateIdentityPoolSubjectTokenSupplierTest.java
```diff
@@ -143,4 +157,274 @@ public void getSubjectToken_success() throws Exception {
...
+    assertEquals(expectedRootErrorMessage, cause.getMessage());
+  }
+
+  @Test
+  public void getSubjectToken_trustChainOnlyLeaf_success() throws Exception {
+    // Configure mock to use the leaf certificate file itself as the trust chain file,
+    // simulating a scenario where the trust chain file contains only the leaf.
+    ClassLoader classLoader = getClass().getClassLoader();
+    URL trustChainUrl = classLoader.getResource("x509_leaf_certificate.pem");
+    assertNotNull("Leaf certificate file not found!", trustChainUrl);
```
lqiu96, line 263: This should be trustchain file not found, right? In this case, the trustchain only contains the leaf cert.

oauth2_http/javatests/com/google/auth/oauth2/CertificateIdentityPoolSubjectTokenSupplierTest.java
```diff
@@ -143,4 +157,274 @@ public void getSubjectToken_success() throws Exception {
     // Verify
     assertEquals(expectedSubjectToken, actualSubjectToken);
   }
+
+  @Test
+  public void getSubjectToken_trustChainWithLeaf_success() throws Exception {
```
lqiu96, line 162: nit: Can this be `getSubjectToken_trustChainWithLeafFirst_success` or something else to clarify that this is only successful because the leaf is first in the trustChain. IIUC, leaf in trustchain not first is an error.

oauth2_http/javatests/com/google/auth/oauth2/CertificateIdentityPoolSubjectTokenSupplierTest.java
```diff
@@ -143,4 +157,274 @@ public void getSubjectToken_success() throws Exception {
...
+
+    // Execute
+    String actualSubjectToken = supplier.getSubjectToken(mockContext);
+
+    // Verify
+    assertEquals(expectedSubjectToken, actualSubjectToken);
+  }
+
+  @Test
+  public void getSubjectToken_trustChainWrongOrder_throwsIllegalArgumentException() {
```
lqiu96, line 231: Can you add a comment above this test to clarify what wrong order means? I believe this case would be that the leaf is not first

oauth2_http/testresources/trust_chain_with_leaf.pem
```diff
@@ -0,0 +1,52 @@
...
+ZpsL7WqgsZS7Q+0VRK8gKfqkxg5OYQIDAQABo3YwdDAdBgNVHQ4EFgQU2RQ8yO+O
+gN8oVW2SW7RLrfYd9jEwRQYDVR0jBD4wPIAU2RQ8yO+OgN8oVW2SW7RLrfYd9jGh
+GaQXMBUxEzARBgNVBAMTCnVuaXQtdGVzdHOCCQDHyErgUOZvuTAMBgNVHRMEBTAD
+AQH/MA0GCSqGSIb3DQEBBQUAA4IBAQBRv+M/6+FiVu7KXNjFI5pSN17OcW5QUtPr
+odJMlWrJBtynn/TA1oJlYu3yV5clc/71Vr/AxuX5xGP+IXL32YDF9lTUJXG/uUGk
++JETpKmQviPbRsvzYhz4pf6ZIOZMc3/GIcNq92ECbseGO+yAgyWUVKMmZM0HqXC9
+ovNslqe0M8C1sLm1zAR5z/h/litE7/8O2ietija3Q/qtl2TOXJdCA6sgjJX2WUql
+ybrC55ct18NKf3qhpcEkGQvFU40rVYApJpi98DiZPYFdx1oBDp/f4uZ3ojpxRVFT
+cDwcJLfNRCPUhormsY7fDS9xSyThiHsW9mjJYdcaKQkwYZ0F11yB
+-----END CERTIFICATE-----
```
lqiu96, line 19: (not in scope for this PR/ can you create a follow up issue for this): Can you document how you create these dummy Certs for tests for future maintainers in a DEVELOPMENT.md file? 

and/ or how I can decode/ parse/ debug these certs to troubleshoot any flaky or failing tests cases?

lqiu96 (approved):
Changes LGTM. Added a few clarifying nits in the tests, but everything else is fine.

Leaving this up to @zhumin8 for final approval. She can help coordinate if this going into the next release cycle or the one after it.

oauth2_http/javatests/com/google/auth/oauth2/CertificateIdentityPoolSubjectTokenSupplierTest.java
nbayati: The certificate file is being used for both leaf and trust chain here. I've updated the message to `Test resource 'x509_leaf_certificate.pem' (used as trust chain) not found!` to be a bit more clear.

oauth2_http/java/com/google/auth/oauth2/CertificateIdentityPoolSubjectTokenSupplier.java
nbayati: Done!

oauth2_http/testresources/trust_chain_with_leaf.pem
nbayati: Created https://github.com/googleapis/google-auth-library-java/issues/1748.

oauth2_http/javatests/com/google/auth/oauth2/CertificateIdentityPoolSubjectTokenSupplierTest.java
```diff
@@ -42,11 +42,14 @@
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
+import java.nio.file.NoSuchFileException;
```
zhumin8, line 45: Cannot comment on code not part of this PR, thus leaving a note here.
Can you extend L34 and import specific classes used in `org.junit.Assert`? go/java-style#s3.3.1-wildcard-imports

zhumin8 (approved):
LGTM

PR #1754, opened 2025-05-22
feat: support ability to set universe domain in ServiceAccountJwtAccessCredentials

mpeddada1:
* Adds the ability to read the universe_domain field from a json file and set it to ServiceAccountJwtAccessCredentials.
* Adds the universeDomain field to the ServiceAccountJwtAccessCredentials#Builder. For java-bigtable's use case, when GOOGLE_APPLICATION_CREDENTIALS is set to a service account's JSON path, these lines get exercised to create ServiceAccountJwtAccessCredentials using the builder:
https://github.com/googleapis/sdk-platform-java/blob/49a7ae50071e75fe0d161a4eb9360a4fe4e6147b/gax-java/gax/src/main/java/com/google/api/gax/core/GoogleCredentialsProvider.java#L82-L92

Once this PR is merged, we need to update auth-library-java in sdk-platform-java and merge https://github.com/googleapis/sdk-platform-java/pull/3806

See https://github.com/mpeddada1/sa-universe-domain for full test setup.


oauth2_http/java/com/google/auth/oauth2/ServiceAccountJwtAccessCredentials.java
```diff
@@ -115,21 +123,29 @@ private ServiceAccountJwtAccessCredentials(
...
+      String quotaProjectId,
+      String universeDomain) {
     this.clientId = clientId;
     this.clientEmail = Preconditions.checkNotNull(clientEmail);
     this.privateKey = Preconditions.checkNotNull(privateKey);
     this.privateKeyId = privateKeyId;
     this.defaultAudience = defaultAudience;
     this.credentialsCache = createCache();
     this.quotaProjectId = quotaProjectId;
+    if (universeDomain == null || universeDomain.trim().isEmpty()) {
```
lqiu96, line 144: Can this validation be done in `setUniverseDomain()`? If user passes null or empty universe_domain, then we throw an IllegalArgumentException

oauth2_http/java/com/google/auth/oauth2/ServiceAccountJwtAccessCredentials.java
```diff
@@ -160,6 +176,10 @@ static ServiceAccountJwtAccessCredentials fromJson(Map<String, Object> json, URI
     String privateKeyPkcs8 = (String) json.get("private_key");
     String privateKeyId = (String) json.get("private_key_id");
     String quoataProjectId = (String) json.get("quota_project_id");
+    String rawUniverseDomain = (String) json.get("universe_domain");
+    String resolvedUniverseDomain =
+        (rawUniverseDomain == null) ? Credentials.GOOGLE_DEFAULT_UNIVERSE : rawUniverseDomain;
```
lqiu96, line 181: qq, is this for backwards compatibility? If the `universe_domain` is null, then we set it to GDU?

Should we also check if it's null or empty, we throw an exception?

oauth2_http/java/com/google/auth/oauth2/ServiceAccountJwtAccessCredentials.java
```diff
@@ -399,6 +437,11 @@ public final String getPrivateKeyId() {
     return privateKeyId;
   }
 
+  @Override
+  public final String getUniverseDomain() {
+    return universeDomain;
+  }
```
lqiu96, line 443: Can you add javadocs for this public method

oauth2_http/java/com/google/auth/oauth2/ServiceAccountJwtAccessCredentials.java
```diff
@@ -522,6 +567,12 @@ public Builder setQuotaProjectId(String quotaProjectId) {
       return this;
     }
 
+    @CanIgnoreReturnValue
+    public Builder setUniverseDomain(String universeDomain) {
+      this.universeDomain = universeDomain;
+      return this;
+    }
```
lqiu96, line 574: nit: javadocs here as well

oauth2_http/java/com/google/auth/oauth2/ServiceAccountJwtAccessCredentials.java
mpeddada1: That's right, this defaults to GDU if it's empty. But good point, it might be better to throw an exception.

mpeddada1: After offline discussion, when reading from credential files, we want to default to GDU if a universe_domain is not provided. This is to support legacy credential file patterns.

oauth2_http/java/com/google/auth/oauth2/ServiceAccountJwtAccessCredentials.java
mpeddada1: Done, thank you.

oauth2_http/java/com/google/auth/oauth2/ServiceAccountJwtAccessCredentials.java
```diff
@@ -546,9 +597,19 @@ public String getQuotaProjectId() {
       return quotaProjectId;
     }
 
+    public String getUniverseDomain() {
+      return universeDomain;
+    }
```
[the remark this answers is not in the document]
mpeddada1: Done, thanks!

oauth2_http/java/com/google/auth/oauth2/ServiceAccountJwtAccessCredentials.java
mpeddada1: From offline discussion: We've decided to keep this logic in the constructor. This follows the convention set by ServiceAccountCredentials which sets the default value to GDU if empty. See https://github.com/googleapis/google-auth-library-java/blob/b347603db4deabb4eb34ed14b96178c95d3e7b45/oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java#L127

where super is GoogleCredentials and sets the universe_domain in https://github.com/googleapis/google-auth-library-java/blob/b347603db4deabb4eb34ed14b96178c95d3e7b45/oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java#L355-L358

This logic is slightly hidden since ServiceAccountCredentials takes in a Builder as a parameter to the constructor whereas ServiceAccountJwtCredentials takes in individual fields

lqiu96 (approved):

PR #1798, opened 2025-09-03
fix: indicate non-validated external credentials in generic methods

diegomarquezp:
### Changes summary
 * `GoogleCredentials`' generic methods for loading from stream have been marked with `@ObsoleteApi` due to lack of validation.
 * `ImpersonatedCredentials` and `ExternalAccountCredentials` `fromJson()` methods now have a javadoc indicating the lack of validation.
 
### Observations
 * `GoogleCredentials` already had a warning on the lack of validation in their `fromStream()` methods.
 * `ImpersonatedCredentials` `fromJson()` method delegates to the concrete `fromJson()` methods, meaning an obsolete api annotation is not necessary.
 * `ExternalAccountCredentials` `fromJson()` method delegates to the AutoValue builders, meaning an obsolete api annotation is not necessary.

BEGIN_COMMIT_OVERRIDE

BEGIN_NESTED_COMMIT
fix: indicate non-validated external credentials in generic methods
END_NESTED_COMMIT

BEGIN_NESTED_COMMIT
deps: add `com.google.api:api-commons` dependency
END_NESTED_COMMIT

BEGIN_NESTED_COMMIT
deps: update `com.google.errorprone:error_prone_annotations` dependency to 2.38.0
END_NESTED_COMMIT

END_COMMIT_OVERRIDE


pom.xml
```diff
@@ -195,6 +196,21 @@
...
+        <exclusions>
+          <exclusion>
+            <groupId>com.google.errorprone</groupId>
+            <artifactId>error_prone_annotations</artifactId>
+          </exclusion>
+          <exclusion>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+          </exclusion>
+        </exclusions>
```
lqiu96, line 212: What was the reason we need to exclude these deps? Is this failing enforcer errors?

lqiu96:
> This will be out of phase by one release of sdk-platform-java due to a circular dependency.
However, this library is stable and we don't expect changes to happen more than a few times per year.

Can you remove this from the commit message/ release notes page? I think this would be moreso implementation detail for us. If anything, I think we can manually add this to the github release page when the auth library is released.

Also, can you add a similar description to our README here with what we have in gax for these annotations? https://github.com/googleapis/sdk-platform-java/blob/f18318660c05d0d8466e3ead7127f0747fac2e2e/gax-java/README.md?plain=1#L89-L113

pom.xml
diegomarquezp: Yes, indeed. I added this to the PR description (and release notes).

diegomarquezp:
@lqiu96 Thanks for your review. I have added a small section in the README ("Versioning") and modified the description to avoid exposing too much detail on the api-commons dependency.

lqiu96:
Changes LGTM. Just a few things to check out and smaller notes to add. I would love to add this as part of the v1.39.0 release.

README.md
```diff
@@ -1521,6 +1521,35 @@ To run the tests you will need:
...
+1. Components marked with `@InternalExtensionOnly` are stable for usage, but
+   not for extension. Thus, methods will not be removed from interfaces marked
+   with this annotation, but methods can be added, thus breaking any
+   code implementing the interface. See the javadocs for more details on other
+   consequences of this annotation.
+   
+1. Components marked with `@ObsoleteApi` are stable for usage in the current major version,
+   but will be marked with `@Deprecated` in a future major version.
+   **NOTE**: We reserve the right to mark anything as `@Deprecated` and introduce breaking
+   changes in a minor version for anythign that we deem to be ***critical bugs and
```
lqiu96, line 1550:
```suggestion
   changes in a minor version to fix any ***critical bugs and
```

lqiu96: Can you also move the section for InternalApi and ObsoleteApi/ Deprecated to the top of the section? I think those two are a far more important for readers

pom.xml
diegomarquezp: I missed this comment. 
Done, and the exclusions are not necessary anymore, but guava was not bumped up or down, it simply went from `x.x.x-android` to `x.x.x-jre`.
Do you know if this has any implications for android compatibility, or whether we are meant to support android? I see no claim of this in this repository.

lqiu96: Is there a reason it must go from `-android` to `-jre`? Can we keep it as `-android` and only bump the version?

I don't know if we promise anything for android compatibility in this repo. But I would prefer to err on the side of caution and keep it the way it is until we do a deeper investigation to swap from `-android` to `-jre`.

pom.xml
```diff
@@ -195,6 +196,17 @@
...
+      <dependency>
+        <groupId>com.google.api</groupId>
+        <artifactId>api-common</artifactId>
+        <version>${project.api-common.version}</version>
+        <exclusions>
+          <exclusion>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+          </exclusion>
+        </exclusions>
```
lqiu96, line 208: @diegomarquezp I meant if you can add a comment about the exclusion into the pom.xml and not in the release notes. 

I think us deciding the exclude guava is more of a implementation detail for us and not relevant for users.

PR #1808, opened 2025-09-30
feat: Add TrustBoundaries support for ServiceAccounts

vverman:
Trust Boundary for service accounts in google-auth-java library.

### DIFF
--- a/oauth2_http/javatests/com/google/auth/oauth2/MockTokenServerTransport.java
+++ b/oauth2_http/javatests/com/google/auth/oauth2/MockTokenServerTransport.java
@@ -327,8 +327,8 @@ public class MockTokenServerTransport extends MockHttpTransport {
           };
       return request;
     } else if (urlWithoutQuery.endsWith("/allowedLocations")) {
-        // This is for mocking the call to IAM lookup when we attempt to refresh trust boundaries
-        // after refreshing the access token.
+      // This is for mocking the call to IAM lookup when we attempt to refresh trust boundaries
+      // after refreshing the access token.
       request =
           new MockLowLevelHttpRequest(url) {
             @Override


oauth2_http/java/com/google/auth/oauth2/TrustBoundary.java
```diff
@@ -0,0 +1,192 @@
+/*
+ * Copyright 2024, Google LLC
```
lqiu96, line 2:
```suggestion
 * Copyright 2025, Google LLC
```

oauth2_http/java/com/google/auth/oauth2/TrustBoundaryProvider.java
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright 2024, Google LLC
```
lqiu96, line 2:
```suggestion
 * Copyright 2025, Google LLC
```

oauth2_http/java/com/google/auth/oauth2/TrustBoundaryProvider.java
```diff
@@ -0,0 +1,45 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package com.google.auth.oauth2;
+
+import com.google.auth.http.HttpTransportFactory;
+import java.io.IOException;
+
+/** An interface for credentials that support trust boundaries. This is an experimental feature. */
+public interface TrustBoundaryProvider {
```
lqiu96, line 38: Can you try and include the `@BetaApi` annotation? I think this should be available from api-common-java deps

oauth2_http/java/com/google/auth/oauth2/TrustBoundary.java
```diff
@@ -0,0 +1,192 @@
...
+    public String toString() {
+      return MoreObjects.toStringHelper(this)
+          .add("encodedLocations", encodedLocations)
+          .add("locations", locations)
+          .toString();
+    }
+  }
+
+  // expose this setter only for testing purposes
+  public static void setEnvironmentProviderForTest(@Nullable EnvironmentProvider provider) {
```
lqiu96, line 115: For testing helpers, can you make this package-private scope. And optionally mark this with `@VisibleForTesting`?

oauth2_http/java/com/google/auth/oauth2/TrustBoundary.java
```diff
@@ -0,0 +1,192 @@
...
+    environmentProvider = provider == null ? SystemEnvironmentProvider.getInstance() : provider;
+  }
+
+  static boolean isTrustBoundaryEnabled() throws IOException {
+    String tbEnabled = environmentProvider.getEnv(GOOGLE_AUTH_TRUST_BOUNDARY_ENABLED_ENV_VAR);
```
lqiu96, line 120: nit: Can you spell out the variables: `trustBoundaryEnabled`?

oauth2_http/java/com/google/auth/oauth2/TrustBoundary.java
```diff
@@ -0,0 +1,192 @@
...
+    if ("true".equals(lowercasedTbEnabled) || "1".equals(tbEnabled)) {
+      return true;
+    }
+    if ("false".equals(lowercasedTbEnabled) || "0".equals(tbEnabled)) {
+      return false;
+    }
+    throw new IOException(
+        String.format(
+            "Invalid value for %s environment variable: \"%s\". Supported values are 'true', '1', 'false', or '0'.",
+            GOOGLE_AUTH_TRUST_BOUNDARY_ENABLED_ENV_VAR, tbEnabled));
```
lqiu96, line 134: IIUC, I think this should just be `return false`? The feature gate would be any value other than "true" or "1"?

oauth2_http/java/com/google/auth/oauth2/TrustBoundary.java
```diff
@@ -0,0 +1,192 @@
...
+      String url,
+      AccessToken accessToken,
+      @Nullable TrustBoundary cachedTrustBoundary)
+      throws IOException {
+    if (accessToken == null) {
+      throw new IOException("The provided access token is null.");
+    }
+    if (accessToken.getExpirationTime() != null
+        && accessToken.getExpirationTime().before(new Date())) {
+      throw new IOException("The provided access token is expired.");
```
lqiu96, line 148: Same here, perhaps better suited as `IllegalArgumentException` or maybe `IllegalStateException`?

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -382,14 +391,51 @@ static Map<String, List<String>> addQuotaProjectIdToRequestMetadata(
     return Collections.unmodifiableMap(newRequestMetadata);
   }
 
+  protected void refreshTrustBoundaries(AccessToken newAccessToken) throws IOException {
+    if (!(this instanceof TrustBoundaryProvider)) {
+      return;
+    }
```
lqiu96, line 397: qq, why is this check needed?

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -382,14 +391,51 @@ static Map<String, List<String>> addQuotaProjectIdToRequestMetadata(
...
+    if (!TrustBoundary.isTrustBoundaryEnabled() || !isDefaultUniverseDomain()) {
+      return;
+    }
+
+    TrustBoundaryProvider provider = (TrustBoundaryProvider) this;
+    synchronized (lock) {
```
lqiu96, line 403: I think we should avoid using a lock around a network call. Can we isolate the part that need sync/ locking for?

oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java
```diff
@@ -819,6 +823,18 @@ public boolean getUseJwtAccessWithScope() {
...
 
+  @Override
+  public HttpTransportFactory getTransportFactory() {
+    return transportFactory;
+  }
+
+  @Override
+  public String getTrustBoundaryUrl() throws IOException {
+    return String.format(
+        "https://iamcredentials.%s/v1/projects/-/serviceAccounts/%s/allowedLocations",
```
lqiu96, line 834: Can this string format be a constant in the Oath2Utils or some shared place?

oauth2_http/java/com/google/auth/oauth2/TrustBoundary.java
```diff
@@ -0,0 +1,192 @@
...
+
+  private static EnvironmentProvider environmentProvider = SystemEnvironmentProvider.getInstance();
+
+  public String getEncodedLocations() {
+    return encodedLocations;
+  }
+
+  public List<String> getLocations() {
+    return locations;
+  }
```
lqiu96, line 83: For all other public methods, can you add public javadocs?

Not required for package-private, but definitely would help with future maintaince! Thanks!

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
vverman: 1. The instanceof TrustBoundaryProvider check is needed to determine if the specific credential instance actually supports trust boundaries. If it doesn't implement the interface, we simply exit.

   2. The cast to TrustBoundaryProvider is then necessary so we can call the methods defined in that interface (like getTrustBoundaryUrl() and getTransportFactory()) on the credential to proceed with the refresh logic.

  This pattern allows us to avoid duplicating the refresh logic in every credential subclass while ensuring it only runs on the credentials that explicitly support the feature.

lqiu96 (dismissed):

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -331,6 +334,74 @@ public GoogleCredentials createWithQuotaProject(String quotaProject) {
...
+   * to enforce security policies by ensuring that the credentials used to access Google Cloud APIs
+   * are not used outside a trusted environment.
+   *
+   * @param newAccessToken The new access token to be used for the refresh.
+   * @param trustBoundaryUrl The URL of the trust boundary service.
+   * @param transportFactory The HTTP transport factory to be used for the refresh.
+   * @throws IOException If the refresh fails and no cached value is available.
+   */
+  @InternalApi
+  void refreshTrustBoundary(
```
lsirac: Multiple threads might read trustBoundary as null inside the first synchronized block, and then all proceed to call TrustBoundary.refresh() simultaneously, resulting in a 'thundering herd' of redundant requests to the TB endpoint.

Do we need to implement request coalescing here (similar to how OAuth2Credentials uses refreshTask for access tokens) so that only one network request is in flight at a time? 


cc @lqiu96 since I think you guys might've already discussed this

oauth2_http/java/com/google/auth/oauth2/TrustBoundary.java
```diff
@@ -0,0 +1,224 @@
...
+   * Checks if this TrustBoundary represents a "no-op" (no restrictions).
+   *
+   * @return True if the encoded locations indicate no restrictions, false otherwise.
+   */
+  public boolean isNoOp() {
+    return NO_OP_VALUE.equals(encodedLocations);
+  }
+
+  /** Represents the JSON response from the trust boundary endpoint. */
+  public static class TrustBoundaryResponse extends GenericJson {
```
lsirac: Does this need to be public?

lsirac (commented):
Mostly LG 👍 , would like us to go back to using TrustBoundaryProvider interface

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
vverman: per discussion: We're now implementing `TrustBoundaryProvider` interface which contains `getTrustBoundaryUrl()`. The classes will override the method and mark them as @internalapi.

oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
```diff
@@ -331,6 +334,74 @@ public GoogleCredentials createWithQuotaProject(String quotaProject) {
...
+   * @return {@code true} if the credentials support trust boundary, {@code false} otherwise.
+   */
+  boolean supportsTrustBoundary() {
+    return false;
+  }
+
+  /**
+   * Refreshes the trust boundary by making a call to the trust boundary URL.
+   *
+   * <p>This method is for internal use only and should not be called by users directly. It is used
```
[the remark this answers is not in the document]
vverman: Removed the comment

oauth2_http/java/com/google/auth/oauth2/TrustBoundary.java
```diff
@@ -0,0 +1,227 @@
...
+      throws IOException {
+    Preconditions.checkNotNull(accessToken, "The provided access token is null.");
+    if (accessToken.getExpirationTime() != null
+        && accessToken.getExpirationTime().before(new Date())) {
+      throw new IllegalArgumentException("The provided access token is expired.");
+    }
+
+    HttpRequestFactory requestFactory = transportFactory.create().createRequestFactory();
+    HttpRequest request = requestFactory.buildGetRequest(new GenericUrl(url));
+    //    request.getHeaders().setAuthorization("Bearer " + accessToken.getTokenValue());
```
lqiu96: I think these comments were left here

oauth2_http/javatests/com/google/auth/oauth2/LoggingTest.java
```diff
@@ -441,7 +441,7 @@ public void getRequestMetadata_hasAccessToken() throws IOException {
 
     TestUtils.assertContainsBearerToken(metadata, ACCESS_TOKEN);
 
-    assertEquals(6, testAppender.events.size());
```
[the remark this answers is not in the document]
lqiu96: > The number of logs increased because we make additional calls to fetch the default service account from the MDS.

Wait sorry, I'm confused about this. I don't think anything changed regarding accessing the default SA? Why are we fetching the default SA now?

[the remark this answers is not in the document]
vverman: No, there isn't any change. It is the way the code is structured.

1. We pass in the `trustBoundaryUrl` into the `refreshTrustBoundaries(...)` after each token refresh.

2. The way we construct the `trustboundaryUrl` is by making a call to the MDS server, getting the default SA and generating the lookup url for that.

3. This call to the metadata server adds to the number of logs we see in the test.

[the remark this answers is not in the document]
lqiu96: > The way we construct the trustboundaryUrl is by making a call to the MDS server, getting the default SA and generating the lookup url for that.

OH. Wait, this is adding a network call for folks that don't even enable TB then for ComputeEngineCredentials. Since a call to `getTrustBoundaryUrl()` invokes a call to MDS. 

Can we change the logic so that the url is constructed inside `refreshTrustBoundary()` call?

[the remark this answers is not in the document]
lqiu96: Ah, yeah that's fine. Since we're using interfaces, I think we're limited to having to to instanceof checks.

I just saw Leo's comment about this functionality being enabled by default once the experiment phase is over. I think long term doesn't matter since this network call will have to happen at some point. I think this change still makes sense so we don't surprise users with increased latency even if they don't explicitly opt-in to the experiment.

oauth2_http/java/com/google/auth/oauth2/TrustBoundary.java
```diff
@@ -0,0 +1,227 @@
...
+    HttpRequestFactory requestFactory = transportFactory.create().createRequestFactory();
+    HttpRequest request = requestFactory.buildGetRequest(new GenericUrl(url));
+    request.getHeaders().setAuthorization("Bearer " + accessToken.getTokenValue());
+
+    // Add the cached trust boundary header, if available.
+    if (cachedTrustBoundary != null) {
+      String headerValue =
+          cachedTrustBoundary.isNoOp() ? "" : cachedTrustBoundary.getEncodedLocations();
+      request.getHeaders().set(TRUST_BOUNDARY_KEY, headerValue);
+    }
```
lqiu96: I think I missed this earlier. Why does the call to `/allowedLocations` also need the header? I understand that the header needs to be added to client library calls. I might have missed this part in the doc

SHANATHAF1:
Hello,
Can you please tell me what these are and what they mean?  Are you fixing
my site or?  Thank you!

Best regards,

Shanatha Fazenbaker

*Fazenbaker Agency*
*Call or Text (607)317-7753 SHARE & SAVE ALL OF INFORMATION TO YOUR
CONTACTS AT THE LINK BELOW*
*https://dot.cards/sfazenbakeragency <https://dot.cards/sfazenbakeragency>*
NEED AN APPOINTMENT SCHEDULE AT YOUR CONVIENIENCE AT
calendly.com/sfazenbakeragency
*A PRODUCER & PROTECTION YOU CAN COUNT ON *
[image: Picture]

oauth2_http/javatests/com/google/auth/oauth2/MockTokenServerTransport.java
```diff
@@ -321,6 +326,27 @@ public LowLevelHttpResponse execute() throws IOException {
             }
           };
       return request;
+    } else if (urlWithoutQuery.endsWith("/allowedLocations")) {
+      // This is for mocking the call to IAM lookup when we attempt to refresh trust boundaries
+      // after refreshing the access token.
```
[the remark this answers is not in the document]
vverman: Will respond to the comments in my upcoming PR for TB for external accounts.

PR #1842, opened 2025-11-07
feat: implement agentic identities authentication via cert bound tokens.

vverman:
This feature implements the ability for agentic identities to authenticate themselves via X509 cert bound tokens. We are limiting the scope here to only cloud run based agentic workloads.

### DIFF
--- a/oauth2_http/java/com/google/auth/oauth2/AgentIdentityUtils.java
+++ b/oauth2_http/java/com/google/auth/oauth2/AgentIdentityUtils.java
@@ -40,6 +40,7 @@ import com.google.common.io.BaseEncoding;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Paths;
@@ -337,14 +338,23 @@ final class AgentIdentityUtils {
     return false;
   }
 
-  /** Calculates the SHA-256 fingerprint of the certificate, Base64Url encoded without padding. */
+  /**
+   * Calculates the URL-encoded, unpadded, base64-encoded SHA256 fingerprint of a
+   * DER-encoded certificate.
+   */
   static String calculateCertificateFingerprint(X509Certificate cert) throws IOException {
     try {
       MessageDigest md = MessageDigest.getInstance("SHA-256");
       byte[] der = cert.getEncoded();
       md.update(der);
       byte[] digest = md.digest();
-      return BaseEncoding.base64Url().omitPadding().encode(digest);
+      // The certificate fingerprint is generated in two steps to align with GFE's
+      // expectations and ensure proper URL transmission:
+      // 1. Standard base64 encoding is applied, and padding ('=') is removed.
+      // 2. The resulting string is then URL-encoded to handle special characters
+      //    ('+', '/') that would otherwise be misinterpreted in URL parameters.
+      String base64Fingerprint = BaseEncoding.base64().omitPadding().encode(digest);
+      return URLEncoder.encode(base64Fingerprint, "UTF-8");
     } catch (GeneralSecurityException e) {
       throw new IOException("Failed to calculate fingerprint for Agent Identity certificate.", e);
     }
--- a/oauth2_http/javatests/com/google/auth/oauth2/AgentIdentityUtilsTest.java
+++ b/oauth2_http/javatests/com/google/auth/oauth2/AgentIdentityUtilsTest.java
@@ -117,9 +117,9 @@ public class AgentIdentityUtilsTest {
 
     // SHA-256 of {0x01, 0x02, 0x03, 0x04, 0xFF} is:
     // fc402e5e4d71483c6d537984a30c2b4c8b065539a4bd1b026c6112926ba52793
-    // Base64Url (no padding) of that hash is:
-    // _EAuXk1xSDxtU3mEowwrTIsGVTmkvRsCbGESkmulJ5M
-    String expectedFingerprint = "_EAuXk1xSDxtU3mEowwrTIsGVTmkvRsCbGESkmulJ5M";
+    // URL-encoded unpadded standard Base64 of that hash is:
+    // %2FEAuXk1xSDxtU3mEowwrTIsGVTmkvRsCbGESkmulJ5M
+    String expectedFingerprint = "%2FEAuXk1xSDxtU3mEowwrTIsGVTmkvRsCbGESkmulJ5M";
 
     String actualFingerprint = AgentIdentityUtils.calculateCertificateFingerprint(mockCert);
     assertEquals(expectedFingerprint, actualFingerprint);


oauth2_http/java/com/google/auth/oauth2/AgentIdentityUtils.java
```diff
@@ -0,0 +1,287 @@
...
+        if (Files.exists(Paths.get(certConfigPath))) {
+          String certPath = extractCertPathFromConfig(certConfigPath);
+          if (!Strings.isNullOrEmpty(certPath) && Files.exists(Paths.get(certPath))) {
+            return certPath;
+          }
+        }
+      } catch (Exception e) {
+        // Ignore exceptions during polling and retry
+        LOGGER.log(Level.FINE, "Error while polling for certificate files");
+      }
```
nbayati: Catching any exception and logging it at every loop iteration can make the logs noisy, instead you can use the `warned` flag to surface the user friendly warning here once. 

We can also limit the catch to only the type of exceptions we are interested in, which would be the IOException thrown from extractCertPathFromConfig().

oauth2_http/java/com/google/auth/oauth2/AgentIdentityUtils.java
```diff
@@ -0,0 +1,287 @@
...
+    String optOut = envReader.getEnv(GOOGLE_API_PREVENT_AGENT_TOKEN_SHARING_FOR_GCP_SERVICES);
+    return optOut != null && "false".equalsIgnoreCase(optOut);
+  }
+
+  /** Polls for the certificate config file and the certificate file it references. */
+  private static String getCertificatePathWithRetry(String certConfigPath) throws IOException {
+    long startTime = timeService.currentTimeMillis();
+    boolean warned = false;
+
+    while (true) {
```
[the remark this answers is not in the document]
vverman: Thanks for the suggestion, it makes sense. I made the changes, do give it a look!

oauth2_http/java/com/google/auth/oauth2/AgentIdentityUtils.java
vverman: Made the changes! Thanks for the catch!

oauth2_http/java/com/google/auth/oauth2/AgentIdentityUtils.java
```diff
@@ -0,0 +1,335 @@
...
+   * @param certConfigPath The path to the certificate configuration JSON file.
+   * @return The path to the certificate file extracted from the config.
+   * @throws IOException If the files cannot be found after the timeout, or if the thread is
+   *     interrupted while waiting.
+   */
+  private static String getCertificatePathWithRetry(String certConfigPath) throws IOException {
+    boolean warned = false;
+
+    // Deterministic polling loop based on pre-calculated intervals.
+    for (long sleepInterval : POLLING_INTERVALS) {
```
lqiu96: I'm concerned about this given that this is to be opt-out and not opt-in. By default, customers may now experience up to 30s of busy waiting just because the filesystem on certain environments can't be loaded fast enough. This applies for customers who don't touch Agent Identities.

IIUC the code from above, the only way to to know if Agent Identities shouldn't be used:
1. Load the cert file (up to 30s of busy waiting)
2. Then load and parse the cert file to check the SAN

OR 
Require users to explicitly add the new env var to their workloads + envs

Can you let me know if my understanding is correct? If this is the case that the cert has to be loaded, then I feel that opt-in would be better.

oauth2_http/java/com/google/auth/oauth2/AgentIdentityUtils.java
```diff
@@ -0,0 +1,335 @@
...
+      try {
+        if (Files.exists(Paths.get(certConfigPath))) {
+          String certPath = extractCertPathFromConfig(certConfigPath);
```
lqiu96: IIUC, null is used to represent that the `cert_path` doesn't exist inside the `certConfigPath`. I think this would be a case where we wouldn't want to retry anymore since the `certConfigPath` is found.

oauth2_http/java/com/google/auth/oauth2/AgentIdentityUtils.java
```diff
@@ -0,0 +1,335 @@
...
+
+  static final String GOOGLE_API_CERTIFICATE_CONFIG = "GOOGLE_API_CERTIFICATE_CONFIG";
+  static final String GOOGLE_API_PREVENT_AGENT_TOKEN_SHARING_FOR_GCP_SERVICES =
+      "GOOGLE_API_PREVENT_AGENT_TOKEN_SHARING_FOR_GCP_SERVICES";
+
+  private static final List<Pattern> AGENT_IDENTITY_SPIFFE_PATTERNS =
+      ImmutableList.of(
+          Pattern.compile("^agents\\.global\\.org-\\d+\\.system\\.id\\.goog$"),
+          Pattern.compile("^agents\\.global\\.proj-\\d+\\.system\\.id\\.goog$"));
+  private static final int SAN_URI_TYPE = 6;
```
lqiu96: nit: Same here, perhaps we can spell out SAN? I don't think this is as well known of an acronym

oauth2_http/java/com/google/auth/oauth2/AgentIdentityUtils.java
```diff
@@ -0,0 +1,335 @@
...
+import java.util.*;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import java.util.regex.Pattern;
+
+/** Internal utility class for handling Agent Identity certificate-bound access tokens. */
+final class AgentIdentityUtils {
+  private static final Logger LOGGER = Logger.getLogger(AgentIdentityUtils.class.getName());
+
+  static final String GOOGLE_API_CERTIFICATE_CONFIG = "GOOGLE_API_CERTIFICATE_CONFIG";
```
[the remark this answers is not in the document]
vverman: The way I see it, it is present in a different module and I don't think it is worth it to add a cross module dependency for one constant import.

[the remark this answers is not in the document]
vverman: I believe that is in the auth/mtls module and we would have to add a dependency for one constant.

oauth2_http/java/com/google/auth/oauth2/AgentIdentityUtils.java
vverman: Interesting thought, in that case the cert-config file itself is malformed which I didn't account for. 

Your thoughts @nbayati, would this even be possible?

As a corollary, what if the cert file is malformed?

oauth2_http/java/com/google/auth/oauth2/AgentIdentityUtils.java
```diff
@@ -0,0 +1,335 @@
...
+  static final String GOOGLE_API_PREVENT_AGENT_TOKEN_SHARING_FOR_GCP_SERVICES =
+      "GOOGLE_API_PREVENT_AGENT_TOKEN_SHARING_FOR_GCP_SERVICES";
+
+  private static final List<Pattern> AGENT_IDENTITY_SPIFFE_PATTERNS =
```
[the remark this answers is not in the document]
vverman: Done!

oauth2_http/java/com/google/auth/oauth2/AgentIdentityUtils.java
vverman: Done!

nbayati (changes requested):
For our future reference: this PR can't merge to main as an standalone without the accompanying changes needed to ensure client libraries have mtls enabled by default, cert rotation is handled, etc (everything that was handled for python in [go/sdk-agent-identity](http://goto.google.com/sdk-agent-identity) 

Also, in python we had to hardcode a well known path due to a bug in CR. We don't have to add it to java yet, but calling this out if by the time we're ready to release this in java, we decide to keep the hardcoded path as a fallback. 
Reference: https://github.com/googleapis/google-auth-library-python/pull/1821/

oauth2_http/java/com/google/auth/oauth2/AgentIdentityUtils.java
```diff
@@ -0,0 +1,378 @@
...
+  }
+
+  /** Calculates the SHA-256 fingerprint of the certificate, Base64Url encoded without padding. */
+  static String calculateCertificateFingerprint(X509Certificate cert) throws IOException {
+    try {
+      MessageDigest md = MessageDigest.getInstance("SHA-256");
+      byte[] der = cert.getEncoded();
+      md.update(der);
+      byte[] digest = md.digest();
+      return BaseEncoding.base64Url().omitPadding().encode(digest);
```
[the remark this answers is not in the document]
vverman: Just made the change, thanks again for catching this!

PR #1886, opened 2026-02-06
doc: add doc for manual re-generation of cab-token-generator protobuf classes

zhumin8:
These classes are generated and in case they need to be re-generated, maintainers can follow these steps to manually generate code.  Text are copied from https://github.com/googleapis/google-auth-library-java/pull/1875 description.


JoeWang1127:
I didn't mention that the package path needs updates, do you think we should add this step?

PR #1896, opened 2026-03-04
fix(gdch): support EC private keys

diegomarquezp:
Context: b/488439640

Implementation originally proposed in b/431924643#comment9

The primary objective is to enable support for Elliptic Curve (EC) keys and non-URI audience formats, aligning the Java SDK with the behavior of the [Python](https://github.com/googleapis/google-auth-library-python/blob/main/google/oauth2/gdch_credentials.py) and Go implementations. Additionally, the GDCH key creation tool creates EC keys only, meaning the GDCH implementation was not following the convention.

### Key Changes

*   **Algorithm Support**: Updated the credential signing process to use the **ES256** (ECDSA) algorithm, which is the required standard for GDCH service accounts.
*   **Audience Flexibility**: Changed the `apiAudience` field from a `URI` to a `String` to accommodate "magic" non-URI strings (e.g., specific administrative audiences) required by certain GDCH services.
*   **OAuth2Utils Refactoring**: Enhanced `privateKeyFromPkcs8` to accept an algorithm parameter, allowing the library to parse EC keys instead of defaulting exclusively to RSA.
*   **Test fixes**: Fixes in a couple of test files as per Sonarqube analysis.

## Testing

*   Updated `GdchCredentialsTest` to include test cases for EC key parsing and token signing.


oauth2_http/java/com/google/auth/oauth2/GdchCredentials.java
```diff
@@ -353,7 +366,7 @@ public final URI getTokenServerUri() {
     return tokenServerUri;
   }
 
-  public final URI getApiAudience() {
+  public final String getApiAudience() {
```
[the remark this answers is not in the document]
diegomarquezp: Something like this: 

```java
  public final String getApiAudienceString() {
    return apiAudience;
  }
  
  public final URI getApiAudience() {
    try {
      return new URI(apiAudience);
    } catch (URISyntaxException e) {
      return null;
    }
  }
```

oauth2_http/java/com/google/auth/oauth2/GdchCredentials.java
```diff
@@ -212,26 +228,42 @@ static GdchCredentials fromJson(Map<String, Object> json, HttpTransportFactory t
...
+
+  /**
+   * Create a copy of GDCH credentials with the specified audience.
+   *
+   * @param apiAudience The intended audience for GDCH credentials.
+   */
+  public GdchCredentials createWithGdchAudience(String apiAudience) {
+    if (Strings.isNullOrEmpty(apiAudience)) {
+      throw new IllegalArgumentException(
+          "Audience are not configured for GDCH service account credentials.");
```
lqiu96:
```suggestion
          "Audience cannot be null or empty for GDCH service account credentials.");
```

oauth2_http/java/com/google/auth/oauth2/GdchCredentials.java
```diff
@@ -497,11 +551,18 @@ public Builder setCaCertPath(String caCertPath) {
     }
 
     @CanIgnoreReturnValue
-    public Builder setGdchAudience(URI apiAudience) {
+    public Builder setGdchAudience(String apiAudience) {
```
lqiu96: for this new setter, can we have validation to check for null or empty audience as well?

oauth2_http/java/com/google/auth/oauth2/GdchCredentials.java
```diff
@@ -37,44 +37,60 @@
...
-  private static final String PARSE_ERROR_PREFIX = "Error parsing token refresh response. ";
+  private static final String BAD_VALUE_ERROR_MESSAGE_FORMAT = "%s Expected %s %s %s.";
+  private static final String PARSE_ERROR_PREFIX = "Error parsing token refresh response.";
+  @VisibleForTesting static final String SUPPORTED_FORMAT_VERSION = "1";
+
+  private static final String ACCESS_TOKEN_TYPE = "urn:ietf:params:oauth:token-type:access_token";
+  private static final String SERVICE_ACCOUNT_TOKEN_TYPE =
+      "urn:k8s:params:oauth:token-type:serviceaccount";
+  private static final String TOKEN_TYPE_TOKEN_EXCHANGE =
+      "urn:ietf:params:oauth:token-type:token-exchange";
```
lqiu96: Actually, I just noticed that a lot of these constants already exist in OAuth2Utils (https://github.com/googleapis/google-auth-library-java/blob/90b61a95112999443e622e31d58e1bda1c1c0bd9/oauth2_http/java/com/google/auth/oauth2/OAuth2Utils.java#L85-L96). 

Can we re-use those values and add any that are missing to that file?

oauth2_http/java/com/google/auth/oauth2/GdchCredentials.java
```diff
@@ -594,4 +658,162 @@ private void setTransport(String caCertPath) throws IOException {
...
+      BigDecimal bigDecimalValue = (BigDecimal) value;
+      return bigDecimalValue.intValueExact();
+    }
+    if (!(value instanceof Integer)) {
+      throw new IOException(
+          String.format(
+              BAD_VALUE_ERROR_MESSAGE_FORMAT, errorPrefix, "integer value", key, "of wrong type"));
+    }
+    return (Integer) value;
+  }
```
lqiu96: I missed this earlier. I think Oauth2Utils has variants for this: https://github.com/googleapis/google-auth-library-java/blob/90b61a95112999443e622e31d58e1bda1c1c0bd9/oauth2_http/java/com/google/auth/oauth2/OAuth2Utils.java#L214

oauth2_http/javatests/com/google/auth/oauth2/DefaultCredentialsProviderTest.java
```diff
@@ -180,48 +181,36 @@ void getDefaultCredentials_noCredentials_singleGceTestRequest() {
...
   @Test
-  void getDefaultCredentials_static_unsupportedPlatform_notGce() throws IOException {
+  void getDefaultCredentials_static_unsupportedPlatform_notGce() {
+    checkStaticGceDetection("macos", "Googlekdjsfhg", false);
+  }
+
+  private void checkStaticGceDetection(String osName, String productContent, boolean expected) {
+    assumeTrue(
+        System.getProperty("os.name").toLowerCase().startsWith("linux"),
+        "This test only runs on Linux.");
```
lqiu96: Correct me if I'm wrong, but I don't think this is the behavior we want. IIUC, this looks like it will only run the test on Linux environments (e.g. our if our CI or local was linux). The original tests was mocking the environment and testing it so that it would mock the os.name value regardless of the CI or local OS.

oauth2_http/javatests/com/google/auth/oauth2/GdchCredentialsTest.java
```diff
@@ -81,7 +85,7 @@ class GdchCredentialsTest extends BaseSerializationTest {
   private static final String CA_CERT_FILE_NAME = "cert.pem";
   private static final String CA_CERT_PATH =
       GdchCredentialsTest.class.getClassLoader().getResource(CA_CERT_FILE_NAME).getPath();
-  private static final URI API_AUDIENCE = URI.create("https://gdch-api-audience");
+  private static final String API_AUDIENCE = URI.create("https://gdch-api-audience").toString();
```
lqiu96: nit: Can we just hard code the string `https://gdch-api-audience` instead of creating a URI object?

alsabinsabi809-coder (approved):

oauth2_http/javatests/com/google/auth/oauth2/GdchCredentialsTest.java
diegomarquezp: Done

lqiu96 (approved):
Changes LGTM. added a few comments.

oauth2_http/java/com/google/auth/oauth2/OAuth2Utils.java
```diff
@@ -269,23 +279,143 @@ static Map<String, Object> validateMap(Map<String, Object> map, String key, Stri
...
     Reader reader = new StringReader(privateKeyPkcs8);
     Section section = PemReader.readFirstSectionAndClose(reader, "PRIVATE KEY");
     if (section == null) {
+      if (algorithm == Pkcs8Algorithm.EC) {
+        reader = new StringReader(privateKeyPkcs8);
+        section = PemReader.readFirstSectionAndClose(reader, "EC PRIVATE KEY");
+        if (section != null) {
+          return privateKeyFromSec1(section.getBase64DecodedBytes());
+        }
+      }
```
lqiu96: Is PKCS#8 or SEC1 two different algorithms/ flows? Also Is this logic only needed for GDCH?

If so, can this logic exists inside GDCH itself?
e.g. 
1. Check the first section and determine if we should call pckc8 or sec1 method
2. call privateKeyFromPkcs8 or privateKeyFromSec1

I think it would better to separate these concerns since it looks like privateKeyFromPkcs8 is now also making a call to privateKeyFromSec1. I think we have added proper checks, but this resides in OAuth2Utils and this may be used in other classes.

oauth2_http/java/com/google/auth/oauth2/GdchCredentials.java
```diff
@@ -814,4 +848,91 @@ static byte[] transcodeDerToConcat(byte[] derSignature, int outputLength)
...
+   *
+   * <p>Assumes the prefix has already been verified.
+   *
+   * @param bytes The raw bytes of the key.
+   * @return The BigInteger value of 's'.
+   */
+  private static BigInteger extractPrivateKeyValue(byte[] bytes) {
+    byte[] sBytes = new byte[32];
+    System.arraycopy(bytes, 7, sBytes, 0, 32);
+    return new BigInteger(1, sBytes);
```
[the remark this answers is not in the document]
diegomarquezp: Done.

PR #1914, opened 2026-04-07
ci: GraalVM CI uses existing Truststore if configured

lqiu96:
Temporary fix for the new Truststore that is being passed in and breaking the GraalVM checks.


diegomarquezp (approved):
Thank you.

